PR #164, opened 2015-08-04
Xml editor demo

cemartins:
Hi Tomas,

I closed the previous pull request because I found a bug in my demo. This one is working ok, now.

Here is my pull request with an XMLEditor demo. I thought the poor man's regex parsing method was still the best (I tried with stax, but found it difficult to get the cursor positions on attributes).

I think it might not be very performant with large xml files, since it calls for a full text parsing on every keystroke. Any ideas how I can make this better?

Kind regards
Carlos

### DIFF
--- a/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/XMLEditor.java
+++ b/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/XMLEditor.java
@@ -17,17 +17,16 @@ import javafx.stage.Stage;
 
 public class XMLEditor extends Application {
 	
-    private static final Pattern XML_TAG = Pattern.compile("(?<SELEMENT>(<\\h*\\w+)([^<>]*)(/?>))"
-    		+"|(?<EELEMENT><\\h*/\\h*\\w+>)"
+    private static final Pattern XML_TAG = Pattern.compile("(?<ELEMENT>(</?)(\\w+)([^<>]*)(/?>))"
     		+"|(?<COMMENT><!--[^<>]+-->)");
     
-    private static final Pattern ATTRIBUTES = Pattern.compile("(\\w+)(=\".+\")");
+    private static final Pattern ATTRIBUTES = Pattern.compile("(\\w+\\h*)(=)(\\h*\"[^\"]+\")");
 
     private static final String sampleCode = String.join("\n", new String[] {
     		"<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\"?>",
     		"<!-- Sample XML -->",
     		"<orders>",
-    		"	<Order number=\"1\">",
+    		"	<Order number=\"1\" table=\"center\">",
     		"		<items>",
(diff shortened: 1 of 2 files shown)


TomasMikula:
Thanks for the contribution, looks good to me.

As to your question, you can

 - a) use a really fast parser and don't mind re-parsing the whole file;
 - b) trigger parsing only when the user paused typing, plus run the parser in the background thread so that it does not slow down the UI. This is implemented in the [JavaKeywordsAsync demo](https://github.com/TomasMikula/RichTextFX/blob/b51c205bcdd096a03e826307aaf3b6cd2ae0a4aa/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/JavaKeywordsAsync.java#L94-L100) and you can read about the mechanism in [this blog post](http://tomasmikula.github.io/blog/2014/04/25/combining-reactfx-and-asynchronous-processing.html);
 - c) use an incremental parser that does not have to re-parse the whole file on every edit. I know about [PapaCarlo](https://github.com/Eliah-Lakhin/papa-carlo), but I don't know how good/mature it is.

cemartins:
Thanks for accepting my contribution :)

I looked into your JavaKeywordsAsync demo and saw that you are using Java8 streams to handle the change events more efficiently. It is really great, but I still need some more time to study it.

Also, if the text is changing because the user is typing, then I can check if the change was made inside an xml tag ( `< .... >` ) and recalculate the syntax highlighting only in this case and only for the affected xml tag.

This is just an idea. I don't quite know how to go about it yet.

TomasMikula:
They are [ReactFX](https://github.com/TomasMikula/ReactFX) `EventStream`s, not Java 8 Streams. Java 8 Streams are not suitable for the task because they are pull-based (demand driven), but UI events don't occur on demand. We need reactive (push-based) kind of streams.

cemartins:
Right. Thank you Tomas.
I must study about ReactFX as well.

JordanMartinez:
@cemartins If you're unfamiliar with Reactive Programming, this article ([The introduction to Reactive Programming you've been missing](https://gist.github.com/staltz/868e7e9bc2a7b8c1f754)) helped me understand the fundamental ideas behind it. That might help you better understand the ReactFX library.

cemartins:
Thank you so much Jordan. I will definetely read this article. I also found the course on reactive programming that Tomas mentioned (https://www.coursera.org/course/reactive <https://www.coursera.org/course/reactive>). Looks like I’m going to be busy learning new tricks for the next few weeks :)

Thank you again for your care and support.

Kind regards
Carlos

JordanMartinez:
Yes, you'll be busy learning some new tricks, but those tricks will save you hours of headaches. :-) 
I would take that class, but it's for Scala; I use Groovy. Still, I might do it just to get a deeper understanding of the ideas and get an idea for how Scala works.

cemartins:
Yep. I work in java, but am quite curious about scala, so this course is a bonus! I'm in! :)

PR #166, opened 2015-08-16
Implemented background colors for text

MewesK:
Just as you suggested the implementation is based on the same principles as the selection fill.
On top of that I had to extend Text to support CSS based colors.

I should mention that I have only implemented -fx-background-color and not anything else from -fx-background. JavaFX under the hood is so damn weird that it would be rather ugly to support the actual Background property. Using just the color property makes the code simple and shouldn't have a huge impact like having an actual Image in every Text instance.

One thing I would like to improve is the background node instantiation. Right now every Text instance has a matching background Path. Maybe it is possible to create only Path instances when there actually is a background color set.

Well I'm actually quite happy with the result. It'S not perfect yet but it works great. The missing background color feature was a real show stopper for me. I'm glad that I was able to fix that. (taught me a lot about JavaFX)

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/skin/TextExt.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/skin/TextExt.java
@@ -1,0 +1,95 @@
+package org.fxmisc.richtext.skin;
+
+import com.sun.javafx.css.converters.PaintConverter;
+import javafx.beans.property.ObjectProperty;
+import javafx.css.CssMetaData;
+import javafx.css.Styleable;
+import javafx.css.StyleableObjectProperty;
+import javafx.css.StyleableProperty;
+import javafx.scene.paint.Color;
+import javafx.scene.paint.Paint;
+import javafx.scene.text.Text;
+
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TextExt extends Text {
+
+    private ObjectProperty<Paint[]> backgroundColor = null;
+
+    public TextExt(String text) {
+        super(text);
+    }
+
+    @Override
+    public List<CssMetaData<? extends Styleable, ?>> getCssMetaData() {
+        try {
+            // Get inner class
+            Class<?> clazz = Class.forName("javafx.scene.text.Text$StyleableProperties");
+
+            // Get field and make it accessible
+            Field styleablesField = clazz.getDeclaredField("STYLEABLES");
+            styleablesField.setAccessible(true);
+
+            // Get list value and make it modifiable
+            List<CssMetaData<? extends Styleable, ?>> styleables = (List<CssMetaData<? extends Styleable, ?>>) styleablesField.get(null);
+            styleables = new ArrayList<>(styleables);
+
+            // Add new properties
+            styleables.add(StyleableProperties.BACKGROUND_COLOR);
+
+            // Set list value
+            return styleables;
+        } catch (ClassNotFoundException | NoSuchFieldException | IllegalAccessException e) {
+            e.printStackTrace();
+        }
+
+        return null;
+    }
+
+    public ObjectProperty<Paint[]> backgroundColorProperty() {
+        if (backgroundColor == null) {
+            backgroundColor = new StyleableObjectProperty<Paint[]>(null) {
+                @Override
+                public Object getBean() {
+                    return TextExt.this;
+                }
+
+                @Override
+                public String getName() {
+                    return "backgroundColor";
+                }
+
+                @Override
+                public CssMetaData<TextExt, Paint[]> getCssMetaData() {
+                    return StyleableProperties.BACKGROUND_COLOR;
+                }
+
+                @Override
+                public void invalidated() {
+                    // TODO
+                }
+            };
+        }
+        return backgroundColor;
+    }
+
+    private static class StyleableProperties {
+
+        private static final CssMetaData<TextExt, Paint[]> BACKGROUND_COLOR = new CssMetaData<TextExt, Paint[]>(
+                "-fx-background-color",
+                PaintConverter.SequenceConverter.getInstance(),
+                new Paint[]{Color.TRANSPARENT}) {
+            @Override
+            public boolean isSettable(TextExt node) {
+                return node.backgroundColor == null || !node.backgroundColor.isBound();
+            }
+
+            @Override
+            public StyleableProperty<Paint[]> getStyleableProperty(TextExt node) {
+                return (StyleableProperty<Paint[]>) node.backgroundColorProperty();
+            }
+        };
+    }
+}
--- a/richtextfx/src/main/java/org/fxmisc/richtext/skin/TextExt.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/skin/TextExt.java
@@ -10,7 +10,6 @@ import javafx.scene.paint.Color;
 import javafx.scene.paint.Paint;
 import javafx.scene.text.Text;
 
-import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -24,28 +23,22 @@ public class TextExt extends Text {
 
     @Override
     public List<CssMetaData<? extends Styleable, ?>> getCssMetaData() {
-        try {
-            // Get inner class
-            Class<?> clazz = Class.forName("javafx.scene.text.Text$StyleableProperties");
+        // Get list value and make it modifiable
+        List<CssMetaData<? extends Styleable, ?>> styleables = new ArrayList<>(super.getCssMetaData());
 
-            // Get field and make it accessible
-            Field styleablesField = clazz.getDeclaredField("STYLEABLES");
-            styleablesField.setAccessible(true);
+        // Add new properties
+        styleables.add(StyleableProperties.BACKGROUND_COLOR);
 
-            // Get list value and make it modifiable
-            List<CssMetaData<? extends Styleable, ?>> styleables = (List<CssMetaData<? extends Styleable, ?>>) styleablesField.get(null);
-            styleables = new ArrayList<>(styleables);
+        // Return list value
+        return styleables;
+    }
 
-            // Add new properties
-            styleables.add(StyleableProperties.BACKGROUND_COLOR);
+    public Paint[] getBackgroundColor() {
+        return backgroundColor.get();
+    }
 
-            // Set list value
-            return styleables;
-        } catch (ClassNotFoundException | NoSuchFieldException | IllegalAccessException e) {
-            e.printStackTrace();
-        }
-
-        return null;
+    public void setBackgroundColor(Paint[] backgroundColor) {
+        this.backgroundColor.set(backgroundColor);
     }
 
     public ObjectProperty<Paint[]> backgroundColorProperty() {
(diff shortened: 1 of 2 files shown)


richtextfx/src/main/java/org/fxmisc/richtext/skin/TextExt.java
```diff
@@ -0,0 +1,95 @@
...
+import javafx.scene.paint.Paint;
+import javafx.scene.text.Text;
+
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TextExt extends Text {
+
+    private ObjectProperty<Paint[]> backgroundColor = null;
```
TomasMikula: If only a single color is supported, the type should not lie about it. Let's just change it to `ObjectProperty<Paint>`.

richtextfx/src/main/java/org/fxmisc/richtext/skin/TextExt.java
```diff
@@ -0,0 +1,95 @@
...
+        try {
+            // Get inner class
+            Class<?> clazz = Class.forName("javafx.scene.text.Text$StyleableProperties");
+
+            // Get field and make it accessible
+            Field styleablesField = clazz.getDeclaredField("STYLEABLES");
+            styleablesField.setAccessible(true);
+
+            // Get list value and make it modifiable
+            List<CssMetaData<? extends Styleable, ?>> styleables = (List<CssMetaData<? extends Styleable, ?>>) styleablesField.get(null);
```
TomasMikula: Why all the reflection instead of just `styleables = Text.getClassCssMetaData()`?

richtextfx/src/main/java/org/fxmisc/richtext/skin/TextExt.java
```diff
@@ -0,0 +1,95 @@
...
+            // Set list value
+            return styleables;
+        } catch (ClassNotFoundException | NoSuchFieldException | IllegalAccessException e) {
+            e.printStackTrace();
+        }
+
+        return null;
+    }
+
+    public ObjectProperty<Paint[]> backgroundColorProperty() {
```
TomasMikula: In addition to this method, we should also add getter and setter.

richtextfx/src/main/java/org/fxmisc/richtext/skin/TextExt.java
```diff
@@ -0,0 +1,95 @@
...
+                    return "backgroundColor";
+                }
+
+                @Override
+                public CssMetaData<TextExt, Paint[]> getCssMetaData() {
+                    return StyleableProperties.BACKGROUND_COLOR;
+                }
+
+                @Override
+                public void invalidated() {
```
TomasMikula: No need to override if the implementation is empty.

TomasMikula:
Great job! See my comments above.

richtextfx/src/main/java/org/fxmisc/richtext/skin/TextExt.java
MewesK: That the way the JavaFX converter works. I was using Color instead of Paint[] before, but the CSS converter injected a Paint array anyway which led to casting errors along the way.

It's the same way in the internal implementation in the Background class. All I can think of is making the property private and expose a Paint property which hides the implementation details.

richtextfx/src/main/java/org/fxmisc/richtext/skin/TextExt.java
MewesK: I was trying to manipulate the internal StyleableProperties.STYLEABLES field in a static section before but it didn't work like I imagined. Overriding getCssMetaData was the final solution but I guess I was too lazy to clean the mess up. It's fixed in the latest commit.

richtextfx/src/main/java/org/fxmisc/richtext/skin/TextExt.java
MewesK: Done.

richtextfx/src/main/java/org/fxmisc/richtext/skin/TextExt.java
MewesK: I was not sure if there is something for me to implement. In the internal implementations this method is used to mark things as dirty etc. But I guess we don't need that for now.

richtextfx/src/main/java/org/fxmisc/richtext/skin/TextExt.java
TomasMikula: [StyleConverter.getPaintConverter()](https://docs.oracle.com/javase/8/javafx/api/javafx/css/StyleConverter.html#getPaintConverter--) should do the job.

MewesK: I tried the ColorConverter before with the mentioned behavior. But I will give this one a try too, thanks.

MewesK: Nope sorry, this one has the same behavior. Under the hood it still injects a Paint array which leads to casting errors later on:

Exception in thread "JavaFX Application Thread" java.lang.ClassCastException: [Ljavafx.scene.paint.Paint; cannot be cast to javafx.scene.paint.Paint

Maybe I'm just too stupid to make it work so don't hold back if you want to try your luck.

TomasMikula: Tried it. You are right, it injects an array. However, if I change the name of the property to something else than `-fx-background-color`, it works as expected. Seems like a bug to me.

MewesK: Just read the CSS reference. Seems like it's not a bug, just a kinda pointless feature in this case:

*Background fills are specified with the properties -fx-background-color, -fx-background-radius and -fx-background-insets. The -fx-background-color property is a series of one or more comma-separated <paint> values. The number of values in the series determines the number of background rectangles that are painted. Background rectangles are painted in the order specified using the given <paint> value. Each background rectangle can have different radii and insets. The -fx-background-radius and -fx-background-insets properties are series of comma-separated values (or sets of values). The radius and insets values used for a particular background are the found in the corresponding position in the -fx-background-radius and -fx-background-insets series. For example, suppose a series of three values is given for the -fx-background-color property. A series of three values should also be specified for the -fx-background-radius and -fx-background-insets properties. The first background will be painted using the first radius value and first insets value, the second background will be painted with the second radius value and second insets value, and so forth.*

TomasMikula: That paragraph applies to `Region`. `Text` is not a `Region`.

MewesK: Of course it does. The background stuff was never intended for Text nodes. I just borrowed everything from the Region implementation. That's why there is seemingly no other way available in JavaFX to handle backgrounds since they don't need it/support it. Seems like we would have to implement our own converter to make it work. Or implement the background feature like they have it intended to be which would be huge hassle and probably a performance killer.

So what are we going to do about it?

TomasMikula: A custom converter is not going to help, because it seems that the supplied converter is overridden anyway by some built-in converter for `-fx-background-color`. I think the easiest thing to do is to call it something else than `-fx-background-color`. E.g. `-fx-background-fill`?

TomasMikula: I reported a bug: https://bugs.openjdk.java.net/browse/JDK-8133685

MewesK: Ranming it might cause confusion for the user who expects -fx-background-color.
So we can either just keep it like it is and tolerate the unnecessary Paint array. Or we rename it.
-fx-background-fill would be a good choice since it follows -fx-highlight-fill. It's your call.

TomasMikula: There is another problem and that is that `com.sun.*` is not going to be visible from JDK 9 on. So unless you can implement it without `com.sun.*`, I would say go with `-fx-background-fill`. (Sure, for now you could probably get away with just giving `null` as the converter, since it is overridden anyway, but that would stop working when/if they fix the bug).

MewesK: I just changed the code based on your last comment. I will test the latest version tonight a bit more.

My next pull request will be about line styles (think red background for lines with break point or or yellow background for current selected line). Any suggestions to make it more likely to get merged upstream?

melkhaldi: A nifty customisation would be controlling line style and background offset
from/around text. This will enable highlighting text in different ways. Not
sure how hard it is to implement though.

In my mind, one of the things to consider is how the background will work
when making text selection. For example, if selection overlaps an already
coloured text background?

I look forward to plying with your addition :)

TomasMikula:
Looks good, thanks for the contribution!

@melkhaldi so you would want the equivalent of the [`-fx-background-insets`](http://docs.oracle.com/javase/8/javafx/api/javafx/scene/doc-files/cssref.html#region) property. It is not super difficult to implement, but is not completely trivial either. You would have to manipulate a closed `Path` to expand by the given insets.

@MewesK Regarding styling the whole lines, let's move the discussion to #6.

TomasMikula:
I have modified the RichText demo to support background color.

melkhaldi:
Hi Tomas, just tried the JAR to test, it does not seem to have means to apply color to background. Did you add a dedicated button/GUI control for it?

image

TomasMikula:
Hi Maher, I haven't uploaded new jars. You would have to build from source at this point. It is just 3 commands, if you have Gradle installed:

```
git clone https://github.com/TomasMikula/RichTextFX.git
cd RichTextFX
gradle RichText
```

TomasMikula:
I will make a new release later.

melkhaldi:
oh alright, i thought it was uploaded already. Don't have gradle set up yet
:) (did a reformat)

JordanMartinez:
I followed Tomas' instructions to set it up from the source. It works good! Great job guys! 

Here's one idea for a slight enhancement (though it isn't a high priority at all). When the background-fill is blue, it's hard to distinguish it from a selected text (see the highlighted, selected example below). This could be improved further if the selection color (which defaults to blue) uses a different color when selecting a text that has a similarly-colored background-fill. 

I also noticed that a white vertical line appears between two texts with different background-fills, but that line is only visible if the text is rather large. Again, not really that big of an issue.

rich text demo_004

MewesK:
Hi Jordan,

the white line is caused by the internal TextLayout class. We could of course just add 1px to the width of each BG shape to work around this "bug" but it may look "wrong" for small text.

The second issue with two similar fill colors for BG and selection could be dealt with by using a slightly transparent color for the selection. (that's the way Mac OS is doing it at least) I'll play around with this idea a little when I#M at home.

JordanMartinez:
@MewesK I think the white line isn't a big enough issue to do that. However, if the situation iss at least known here, then a person who truly needs that to be gone could talk to you about how to adjust it or try doing it themselves...

I'm not very familiar with Mac, so I'm not sure how that would work. I'm sure you'll come up with something :-)

melkhaldi:
I believe this might be something that can be controlled via an inset
property. For example, if char at end is so and so, set inset to +x.

TomasMikula:
Regarding the white gap, we could also try to construct a minimal example, where `TextLayout#getRange(0, 1)` ends at position `x` and `TextLayout#getRange(1, 2)` starts at `x+ɛ` and file it to the JavaFX team. Though I'm not very optimistic that they will resolve it, since `TextLayout` is private API.

PR #168, opened 2015-08-20
Implemented the proposed paragraph style feature (#6)

MewesK:
This PR is WIP and a more of a basis for discussion. Like you described it adds "PS".

In the second commit is the way I propose to set those paragraph styles. To demonstrate it I added it in kind of lazy way to the RichText demo. At the moment the paragraph style vanishes after a text or style update but otherwise it should be complete.

I'll try to fix that bug over the weekend. But early feedback would be appreciated.


MewesK:
The bug should be fixed now. (hopefully)

Let me know what you think.

TomasMikula:
Thanks for the PR. Here are some quick comments:

This pull request is based on an outdated version of this repository and as you can see there are merge conflicts. Can you please first merge branch `TomasMikula:master` into your branch? As you will see, you have duplicated much of the same work as me on the RichText demo.

If you find out that you need or want to do some preparatory refactoring (such as moving StyledTextAreaView into its own source file), please do so in a separate commit and potentially even submit a separate pull request for it first. That will then make it easier to see what was changed in the "feature" pull request. (I now for example cannot see what parts of StyledTextAreaView were modified and what was just copy/paste.)

Paragraph is immutable and I'm convinced it should stay so. The implication is that there can be no `paragraphStyle` _property_ on `Paragraph`, just a `final` field.

PR #190, opened 2015-10-14
Merged proposed paragraph style feature branch of MewesK and the current master branch

jobernolte:
Hi, I merged the current master branch into the feature branch with paragraph styles.


TomasMikula:
Thanks for bringing it up to date with master!

My comment from #168 remains, namely that `Paragraph` should stay immutable. I would also be happy to merge this piece-meal, e.g. if the refactoring of `StyledTextAreaView` into its own file came in as its own pull request.

jobernolte:
Don't know if moving StyledTextAreaView in its own source file is what you wanted. But I made paragraphStyle immutable as you requested.

richtextfx/src/main/java/org/fxmisc/richtext/skin/ParagraphBox.java
```diff
@@ -69,9 +70,10 @@ private CaretOffsetX(double value) {
     public void setIndex(int index) { this.index.setValue(index); }
     public int getIndex() { return index.getValue(); }
 
-    public ParagraphBox(Paragraph<S> par, BiConsumer<? super TextExt, S> applyStyle) {
+    public ParagraphBox(Paragraph<S, PS> par, BiConsumer<? super TextExt, S> applyStyle, PS initialParagraphStyle,  BiConsumer<TextFlow, PS> applyParagraphStyle) {
```
TomasMikula: The argument `initialParagraphStyle` is not necessary.

richtextfx/src/main/java/org/fxmisc/richtext/skin/ParagraphBox.java
```diff
@@ -69,9 +70,10 @@ private CaretOffsetX(double value) {
...
         this.getStyleClass().add("paragraph-box");
         this.text = new ParagraphText<>(par, applyStyle);
+        applyParagraphStyle.accept(this.text, initialParagraphStyle);
```
TomasMikula: The applied style should be `par.getParagraphStyle()` instead of `initialParagraphStyle`.

TomasMikula:
Looks mostly OK. See my inline comments. Also, have you tried it with the demo RichText editor app?

jobernolte:
Made the suggested changes. I also tried the RichText demo app and it seems to work fine.

TomasMikula:
Looks good. Merging.

TomasMikula:
Turns out I was a bit too optimistic. The demo actually doesn't work for styling paragraphs (when [`applyToParagraph`](https://github.com/TomasMikula/RichTextFX/blob/7a7df5b42cf6d6d7cc2afc0ab8f105dccc736eea/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/richtext/RichText.java#L53) is set to true). I "fixed" `updateParagraphStyleInSelection` in 7a7df5b4, but `StyledTextArea#setParagraphStyle` is still not working.

TomasMikula:
Fixed in 8289b0fe.

PR #198, opened 2015-11-02
Exposed Scrolling API for StyledTextArea

JordanMartinez:


TomasMikula:
Looking good. Maybe add comments that `total*Estimate` properties are only supposed to be _set_ by the skin, not the user, and that `estimatedScroll*` can be set in order to scroll the content.

JordanMartinez:
Should I add those comments in the class javadoc or the property's javadoc ?

TomasMikula:
I think property's javadoc is better.

JordanMartinez:
How's that?

TomasMikula:
:+1:

PR #199, opened 2015-11-02
Selection Fix

JordanMartinez:
Turns out I wrote this on top of the branch I used for exposing the scroll API. Do you want me to create a separate branch without the exposure and create a PR ? Not sure what best practices are in this regard.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/NavigationActions.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/NavigationActions.java
@@ -81,13 +81,12 @@ public interface NavigationActions<S, PS> extends TextEditingArea<S, PS> {
         }
     }
 
-
     /**
-     * Skips two word boundaries backwards.
+     * Skips n number of word boundaries backwards.
      * Based on the given selection policy, anchor either moves with
      * the caret, stays put, or moves to the former caret position.
      */
-    default void previousWord(SelectionPolicy selectionPolicy) {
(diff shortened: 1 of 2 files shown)


TomasMikula:
It's good to base independent PRs off the master branch, so that they can be dealt with independently. But we can now first resolve the other PR and once that is merged, the other commit will be cleared from this PR automatically.

JordanMartinez:
Thought so.

TomasMikula:
Why does Shortcut+Left/Right skip 2 boundaries, while Shift+Shortcut+Left/Right only 1?

JordanMartinez:
When I wrote the initial code, I realized that Shift+Shortcut+Left/Right would also select trailing/leading word boundaries, so I reduced it to 1. I wasn't sure if Shortcut+Left/Right should do the same.

TomasMikula:
At least they should be consistent. The problem with skipping just one boundary is that if the cursor is at the beginning of a word and you press (Shift+)Shortcut+Left, it will only move to the end of the previous word, not to the beginning of the previous word. So probably some more sophisticated logic is needed to fix that. But I'm OK if this PR just fixes the double-click case and leaves the rest as it was (i.e. skip 2 word boundaries for the rest).

JordanMartinez:
That makes sense.

JordanMartinez:
I opened an issue for the above inconsistency so it can be dealt with separately.

TomasMikula:
Cool, thank you!

PR #207, opened 2015-11-12
This branch is NOT intended to be merged, only to help with discussion.

JordanMartinez:
First attempt at removing Skin.

Commit notes:
- Flowless (scroll bar free version) was compiled using a local jar.
- StyledTextAreaView's contents were moved into StyledTextArea. If conflicting names of properties/methods were found, StyledTextArea's names stayed the same whereas StyledTextAreaView's names now had an additional "_" character in front of them ("exampleMethod()" -> "_exampleMethod()")
- Any other skin-related classes were moved to the main `org.fxmisc.richtextfx` package.
- StyledTextAreaBehavior's contents were not included in StyledTextArea. Rather, a modified version of the original object was created using the constructed StyledTextArea and its VirtualFlow.
- RichText demo works and embeds the StyledTextArea within a VirtualizedScrollPane. None of the other demos were tested.
- StyledTextAreaView & StyledTextAreaVisual were commented out, where needed, so that the demo could compile.

### Questions I have:
1. What can be optimized? (long methods chains that were needed for Control can be better streamlined, properties and other values can be implemented better, etc.)
2. How should Behavior be handled?
  - Include it in the StyledTextArea
  - Exclude it, but provide it in some other class (so that others can customize the behavior they want)
  - Something else?
3. How should the convenience method for creating and then embedding a StyledTextArea inside a VirtualizedScrollPane work?
  - Each subclass of StyledTextArea provides its own static method? Or overrides the super method?
  - Something else?


TomasMikula:
You did a good job, but I would break this down into smaller steps, which are testable independently. I can see at least 5 steps (order of which is somewhat flexible):

1. Port to new Flowless version (i.e. just replace `VirtualFlow` by `VirtualizedScrollPane<VirtualFlow>`).
2. Don't use Skin, instantiate `StyledTextAreaView` directly in `StyledTextArea` constructor and add it as the only child. Let `createDefaultSkin()` return null. (This can be done independently of 1.)
3. Don't include scrollbars, make wrappable by `VirtualizedScrollPane`. (That is, push `VirtualizedScrollPane` to the outside.) Requires answer to your question 3. above.
4. Inline `StyledTextAreaView` into `StyledTextArea`. (Depends only on 2.)
5. If we see no benefit of `StyledTextArea` extending `Control`, make it extend `Region`. (Depends only on 2.)

I released a snapshot of Flowless 0.5. You can use it if you temporarily add this

```groovy
repositories {
    maven {
        url 'https://oss.sonatype.org/content/repositories/snapshots/' 
    }
}
```

to [richtextfx/build.gradle](https://github.com/TomasMikula/RichTextFX/blob/master/richtextfx/build.gradle).

JordanMartinez:
> Question 3: How should the convenience method for creating and then embedding a StyledTextArea inside a VirtualizedScrollPane work?
>
> 1. Each subclass of StyledTextArea provides its own static method? 
> 2. Or overrides the super method?
> 3. Something else?

Similar to ReactFX's EventStreams class, I think there should be an additional class that has a factory methods that embed a StyledTextArea or one of its subclasses in a VirtualizedScrollPane. Its a matter of a process of elimination:
- If Option 1 is used, then  InlineStyleTextArea could also be used to create a StyledTextArea that's embedded in a VirtualizedScrollPane object. InlineCssTextArea would then be able to create its own area, an InlineStyleTextArea, and a StyledTextArea within a VirtualizedScrollPane. Besides this, InlineStyleTextArea's factory method would need to have a different name to distinguish it from StyledTextArea's factory method since they use the same arguments.
- Option 2 cannot be used because the return type will prevent super methods from being overridden properly.

JordanMartinez:
With that, I would propose a class with factory methods whose names are the name of the area it embeds in a VirtualizedScrollPane:
````java
public class AreaFactory {
    public <S, PS> VirtualizedScrollPane<StyledTextArea<S, PS>  StyledTextArea(args) {
        return new VirtualizedScrollPane<>(new StyledTextArea(args));
    }
    // inlineStyleTextArea
    // inlineCssTextArea
}
````

**Comment by JordanMartinez** [comment]
> Question 2. How should Behavior be handled?
> 
> 1. Include it in the StyledTextArea
> 2. Exclude it, but provide it in some other class (so that others can customize the behavior they want)
> 3. Something else?

The Behavior class was used to separate the Controller logic from the View in the MVC paradigm. However, with the skin now being exposed, the Behavior logic can also be changed. 
- I don't think Option 1 should be the default. Some have wanted to add their own customized behavior without needing to fork this project to do that. 
- Instead, I think a variant of Option 2 should be used. Each constructor found in StyledTextArea or one of its subclasses should have an additional `behavior` argument which handles the Controller logic for that area. Most people would want most of the logic found in the `StyledTextAreaBehavior` class with a few changes they make. Since all events are delegated to methods, those methods can be overridden easily. So, they could extend the class and make their changes. If this approach was used, I think this would further modulate StyledTextArea in a favorable way.

**Comment by TomasMikula** [comment]
> If Option 1 is used, then InlineStyleTextArea could also be used to create a StyledTextArea that's embedded in a VirtualizedScrollPane object. InlineCssTextArea would then be able to create its own area, an InlineStyleTextArea, and a StyledTextArea within a VirtualizedScrollPane. Besides this, InlineStyleTextArea's factory method would need to have a different name to distinguish it from StyledTextArea's factory method since they use the same arguments.

Static methods are not inherited. You can have a static method of the same name and argument types on a subclass, and it is not overriding the superclass method in any sense.

Behavior can already be customized, without the need to make StyledTextAreaBehavior public. See #86 or #91 for examples.

**Comment by JordanMartinez** [comment]
> Static methods are not inherited. You can have a static method of the same name and argument types on a subclass, and it is not overriding the superclass method in any sense.

Oh. I thought they did, but I guess that's because the static methods are available in the classes themselves. So, if I'm writing code in InlineStyleTextArea, StyledTextArea's static method is available.

> Behavior can already be customized, without the need to make StyledTextAreaBehavior public. See #86 or #91 for examples.

Ah... right. I forgot about that.

**Comment by TomasMikula** [comment]
Although you are able to call a static method on an instance

```java
Foo foo = ...;
foo.someStaticMethod();
```
you should not do that and instead do

```java
Foo.someStaticMethod();
```

The IDE even gives me a warning if I try to do the former.

JordanMartinez:
One issue I just ran into. VirtualizedScrollPane does not have a public getter for its content. Thus, if the factory methods are used to embed an area within a VirtualizedScrollPane, there's no access to the area once its created.
Obviously, one can create their own area and then add it to VirtualizedScrollPane, but that defeats the point of convenience that the factory methods would add.

JordanMartinez:
> Static methods are not inherited. You can have a static method of the same name and argument types on a subclass, and it is not overriding the superclass method in any sense.

For those following along, recall that `InlineCssTextArea` extends `InlineStyleTextArea`, which extends `StyledTextArea`. Here's the code I wrote for `StyledTextArea`, `InlineStyleTextArea`, and `InlineCssTextArea`:
````java
// StyledTextArea's factory methods
public static <S, PS> VirtualizedScrollPane<StyledTextArea<S, PS>> withScrollBars(S initialStyle,
      BiConsumer<? super TextExt, S> applyStyle, PS initialParagraphStyle,
      BiConsumer<TextFlow, PS> applyParagraphStyle, boolean preserveStyle) {
    return new VirtualizedScrollPane<>(new StyledTextArea<S, PS>(initialStyle, applyStyle, initialParagraphStyle, applyParagraphStyle, preserveStyle));
}
// difference from above: boolean preserveStyle is true.
public static <S, PS> VirtualizedScrollPane<StyledTextArea<S, PS>> withScrollBars(S initialStyle,
      BiConsumer<? super TextExt, S> applyStyle, PS initialParagraphStyle,
      BiConsumer<TextFlow, PS> applyParagraphStyle) {
    return withScrollBars(initialStyle, applyStyle, initialParagraphStyle, applyParagraphStyle, true);
}

// InlineStyleTextArea's factory method
public static <S, PS> VirtualizedScrollPane<InlineStyleTextArea<S, PS>> withScrollBars(S initialStyle,
       Function<S, String> styleToCss, PS initialParagraphStyle, Function<PS, String> paragraphStyleToCss) {
    return new VirtualizedScrollPane<>(new InlineStyleTextArea<>(initialStyle, styleToCss, initialParagraphStyle, paragraphStyleToCss));
}

// InlineCssTextArea's factory method
public static VirtualizedScrollPane<InlineCssTextArea> withScrollBars(String text) {
    return new VirtualizedScrollPane<>(new InlineCssTextArea(text));
}
````

Using `InlineCssTextArea`, I have access to all the above factory methods:

richtextfx - -github projects-richtextfx - richtextfx-demos - -github projects-richtextfx-richtextfx-demos-src-main-java-org-fxmisc-richtext-demo-richtext-teststatic java - intellij idea 15 0 1_001

**Comment by TomasMikula** [comment]
OK, let me rephrase: The static method of the subclass is not overriding the superclass method in the _usual_ sense

```java
class A {
	static String f() { return "A.f"; }
}

class B extends A {
	static String f() { return "B.f"; }
}

class StaticMethodOverriding {
		B b = new B();
		A a = b;

		System.out.println(a.f());
		System.out.println(b.f());
	}
}
```

Output:

```
A.f
B.f
```

event thought `a` and `b` are the same object.

But I see how you don't like the fact that you can invoke a static method of a superclass from the subclass. Ideally, in a language that lets you specify type aliases, I wouldn't have any descendants of `StyledTextArea`, just type aliases, to mask away some type arguments.

```scala
type StyleClassedTextArea = StyledTextArea<Collection<String>, Collection<String>>
```

Unfortunately we don't have that.

I am quite liberal where we put the factory methods. Keeping them with the respective classes might make them easier to find.

JordanMartinez:
> OK, let me rephrase: The static method of the subclass is not overriding the superclass method in the usual sense .... event though a and b are the same object.

Mm! Great explanation! That clarifies things.

> But I see how you don't like the fact that you can invoke a static method of a superclass from the subclass

Aye because if a developer isn't careful, they might use the wrong factory method. This makes me wonder about using the `AreaFactory` idea. Rather than having the three classes (and four if `CodeArea` is included), why not go the direction you went in ReactFX's EventStreams:
````java
public class AreaFactory {
    public static StyledTextArea styledTextArea(args) {}
    public static StyledTextArea inlineStyledTextArea(args) {}
    public static StyledTextArea inlineCssTextArea(args) {}
    public static StyledTextArea codeArea(args) {}

    // factory methods for the above which embed them in a VirtualizedScrollPane
}
````

TomasMikula:
Well, in fact, the split between `EventStream` and `EventStreams` is the pre-Java 8 way of doing things, because you could not have static methods on interfaces. When I started ReactFX, I was using early access builds of JDK 8, but I didn't realize I could already have those static methods directly on `EventStream`. When I was writing `Val`, I already knew that and there is not class `Vals`.

TomasMikula:
Of course the situation here is somewhat different. I'm not against having a separate class with static factory methods.

JordanMartinez:
>  the split between EventStream and EventStreams is the pre-Java 8 way of doing things... When I was writing `Val`, I already knew that and there is not class `Vals`.

Mm... ok.

> Of course the situation here is somewhat different. I'm not against having a separate class with static factory methods.

I think it would be best. Otherwise, there's two other options: 
- Put all factory methods in `StyledTextArea`. This is weird. Why would one go to `StyledTextArea` to get `InlineCssTextArea` or one of the other classes?
- Let each class have its own factory method. This isn't good either because a developer might use the wrong factory method and it creates unnecessary confusion.

JordanMartinez:
Should I deprecate `InlineStyleTextArea` and `InlineCssTextArea`  classes, but leave them to be deleted in a later commit?

JordanMartinez:
> Should I deprecate `InlineStyleTextArea` and `InlineCssTextArea` classes, but leave them to be deleted in a later commit?

I'm going to assume yes. And with that, I think I have my outcome now. So, going back to the order you listed at the top of this thread, I think these are the steps to take:

1. First Steps - order doesn't matter.
 - Port to new Flowless version (i.e. just replace VirtualFlow by VirtualizedScrollPane<VirtualFlow>).
 - Don't use Skin, instantiate StyledTextAreaView directly in StyledTextArea constructor and add it as the only child. Let createDefaultSkin() return null.
2. Inline StyledTextAreaView into StyledTextArea.
3. Make `StyledTextArea` extend `Region`, not `Control`.
 - I don't see what benefits extending Control would have. Still, I'll test out the differences and let you know what I find.
4. Don't include scrollbars: push VirtualizedScrollPane outside. Create initial `AreaFactory` class that has convenience methods for embedding a StyledTextArea inside a VirutalizedScrollPane.
5. Move Constructor-Wrapper classes to AreaFactory. Deprecate `InlineStyleTextArea` and `InlineCssTextArea` classes

PR #212, opened 2015-11-22
Skin removal

JordanMartinez:
Outcome:
- Skin (VirtualFlow) is accessible from `StyledTextArea`
- Scroll Bars are no longer a part of `StyledTextArea`. However, a factory class called `AreaFactory` can conveniently create `StyledTextArea` or any of its subclasses, and optionally embed that object inside a VirtualizedScrollPane so that scroll bars are included.
- Constructor-Wrapper classes (`InlineStyleTextArea` & `InlineCssTextArea`) are deprecated because `AreaFactory` can create their respective objects.

Steps to take:

1. First Steps - order doesn't matter. 
 - Port to new Flowless version (i.e. just replace VirtualFlow by VirtualizedScrollPane).
 - Don't use Skin, instantiate StyledTextAreaView directly in StyledTextArea constructor and add it as the only child. Let createDefaultSkin() return null.
2. Inline StyledTextAreaView into StyledTextArea.
3. Decide whether `StyledTextArea` extends `Region` or `Control` and implement it.
4. Don't include scrollbars: push VirtualizedScrollPane outside. Create initial AreaFactory class that has convenience methods for embedding a StyledTextArea inside a VirutalizedScrollPane.
5. Move Constructor-Wrapper classes to AreaFactory. Deprecate InlineStyleTextArea and InlineCssTextArea classes


JordanMartinez:
Step 3 is a lot of code to implement in one commit, so I'm not sure if this is related to something else. However, when `StyledTextArea` extends `Control` and its `createDefaultSkin()` method returns `null`, I get an error when I run the RichText demo. This error disappears as soon as STA extends Region:
````
Nov 21, 2015 4:25:41 PM javafx.scene.control.Control impl_processCSS
SEVERE: The -fx-skin property has not been defined in CSS for InlineStyleTextArea@39c4469e[styleClass=styled-text-area] and createDefaultSkin() returned null.
````

JordanMartinez:
In the demos, I'm getting an red squiggle error underneath any reference to a StyledTextArea.
The error reads: "Cannot access org.fxmisc.flowless.Virtualized"
Is that due to something related to the demos' "build.gradle" file?

JordanMartinez:
It also can't access VirtualizedScrollPane....?

JordanMartinez:
richtextfx - -github projects-richtextfx - richtextfx-demos - -github projects-richtextfx-richtextfx-demos-src-main-java-org-fxmisc-richtext-demo-richtext-richtext java - intellij idea 15 0 1_007

JordanMartinez:
Figured it out.

````
repositories {
    mavenCentral() // this was already there...

    // if I move this from richtextfx/build.gradle to build.gradle, it works
    maven {
        url 'https://oss.sonatype.org/content/repositories/snapshots/' 
    }
}
````

TomasMikula:
Regarding the null Skin problem, I was convinced that it is OK to return `null` from `createDefaultSkin`, thought I probably never tried it. That was at least my understanding of its [Javadoc](http://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/Control.html#createDefaultSkin--):

> If null then the control will have no skin unless one is provided by css.

JordanMartinez:
Yeah... One would probably need to add `-fx-skin: styled-text-area-view;` to the css file in order to make that work. I think I watched a video where the developer added a line like that.

JordanMartinez:
Yeah. Here's that [video @ 8:11](https://www.youtube.com/watch?v=7PPcM0E5yQw#t=08m11s)

TomasMikula:
One certainly _can_ add `-fx-skin` CSS property, but I thought we would be fine without it. If not, we will have to use Region as soon as we instantiate StyledTextAreaView directly in StyledTextArea constructor.

JordanMartinez:
Yup. That's what I did.

This PR says "This branch has conflicts that must be resolved." Is that because of the 2nd commit's failure? Or something else?

TomasMikula:
No, there are conflicts with what happened on the master branch in the meantime. (Although my last commit is from Nov 12, this branch is based on an older version still.) You will need to rebase this work on top of the current HEAD of the master branch.

JordanMartinez:
Dang...

JordanMartinez:
I guess the lesson I learned here is to fetch frequently...

Issue #152, opened 2015-06-17
Option to show a split view


JordanMartinez:
Hello Tomas,

How could I split a rich text area (whatever flavor) into two views, something that is typically done in a Word processor? 
There are two parts to this:
1. Two views display a portion of the same document.
2. A change in the document will update both views to display this update

As of now, I could do the first by adding two rich text areas to a Split Pane, but how would I do the second? If I add two areas to a Split Pane, I could have them listen to each other for any updates. This is probably inefficient if the area contains a lot of text, styles, etc, but it provides a work around until a better solution is provided.

melkhaldi:
Hi Jordan,
I did something like that too, by making the text area in focus be the one driving changes in the other. I listen to position of caret and push key entries from the one in focus to the other one. It did not give always the correct sequence of key characters though when I typed quickly .. Probably has something to do with threads (something above my level of programming).
It was OK to show an idea I wanted to explore. But certainly not desirable performance...
Maher
Sent from my Cyanogen phone
On Jun 16, 2015 9:29 PM, JordanMartinez <notifications@github.com> wrote:Hello Tomas,

How could I split a rich text area (whatever flavor) into two views, something that is typically done in a Word processor? 
There are two parts to this:
1. Two views display a portion of the same document.
2. A change in the document will update both views to display this update

As of now, I could do the first by adding two rich text areas to a Split Pane, but how would I do the second? If I add two areas to a Split Pane, I could have them listen to each other for any updates. This is probably inefficient if the area contains a lot of text, styles, etc, but it provides a work around until a better solution is provided.

TomasMikula:
Hi Jordan,

I think the best you can do today is as you suggest - have the two text areas listen to each other. It is not necessarily inefficient, if you observe and re-do the [individual changes](http://www.fxmisc.org/richtext/javadoc/org/fxmisc/richtext/StyledTextArea.html#richChanges--). [Here](https://github.com/TomasMikula/RichTextFX/blob/c0059e5628a523e8a9fdbc063a6a504f66cbcf69/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextArea.java#L779) you can see how to apply a change generated by one text area to the other one.
You will have to take care of ignoring changes from a text area while you are applying changes to it, so that you don't get an infinite loop.

[...]

PR #213, opened 2015-11-22
Skin removal

JordanMartinez:
This one's cleaner.


TomasMikula:
Is there something lost from `StyledTextAreaView` or `StyledTextAreaVisual`, when you keep them around?

JordanMartinez:
> Is there something lost from StyledTextAreaView or StyledTextAreaVisual, when you keep them around?

I don't think so. I just didn't see the need to have them anymore, so I deleted them. Should I leave them there and just deprecate them?

TomasMikula:
No, sorry, I was looking at the commit where you inlined StyledTextAreaView and was wondering why you kept the original file around. Didn't realize you deleted it in the next commit.

JordanMartinez:
Oh ok... 
Looking back at them now, if I don't delete them, I get issues in `StyledTextAreaView` when trying to access CharacterHit because it's no longer in the skin package and I'm guessing you don't want it public.

TomasMikula:
Seems like the `dispose()` method is no longer needed. The Skin had to be disposed to stop observing the Control. Now the "control" is only observing itself (its own properties and its own `EditableStyledDocument`), so disposal is not necessary.

(In the future I would like to be able to provide an external document. At that point, dispose will need to be reintroduced to break ties with the document, if the document outlives the area.)

If you don't see any catch, we can remove the method. That also means we don't need to keep the subscriptions that are unsubscribed in `dispose()`.

JordanMartinez:
> Seems like the dispose() method is no longer needed. 

Yeah, I was wondering about that...

Looking at its dependencies, would I remove all of the following?
- Subscription subscription = () -> {};
- manageSubscription()
- manageBinding()
- subscribeTo()

and then change all `subscribeTo()` to `stream.subscribe( /* code */);` and delete all `manageBinding()` references.

> (In the future I would like to be able to provide an external document. At that point, dispose will need to be reintroduced to break ties with the document, if the document outlives the area.)

Yeah, the idea proposed in #152.

TomasMikula:
One more thing. Seems like `StyledTextAreaBehavior` doesn't need a direct reference to `VirtualFlow`. In the two places it is used, `area` can be used just as well.

JordanMartinez:
I was thinking it would be more precise if `VirtualFlow` was used. But then again, if it's the only child, I guess it doesn't really matter....

TomasMikula:
This was a lot of work, thanks!

JordanMartinez:
Woohoo!

PR #221, opened 2015-12-04
Open up access to CharacterHit for subclasses of StyledTextArea

JordanMartinez:


TomasMikula:
Hmm, I haven't noticed that `ParagraphBox` became public. I don't think it should be. Then only the `hit(double x, double y)` method can be made public.

JordanMartinez:
It's public even before the skin removal process. Should I make both the class and its constructor package private in this branch or via a new one?

JordanMartinez:
It's changed to public in this commit: 0ae84c25936bdb285434ecc96564f0b1f946221d

TomasMikula:
Either way ;)

JordanMartinez:
`ParagraphBox`'s constructor is public long before this mistake. I'm guessing it's public so it could be accessed by the main richtextfx package outside the skin package?

However, now that it's inside the richtextfx package, it doesn't need to be public anymore.

richtextfx/src/main/java/org/fxmisc/richtext/StyledTextArea.java
```diff
@@ -677,7 +677,7 @@ CharacterHit hit(ParagraphBox.CaretOffsetX x, double y) {
         }
     }
 
-    CharacterHit hit(double x, double y) {
+    protected CharacterHit hit(double x, double y) {
```
TomasMikula: Don't we want this one public and keep the other two package private?

JordanMartinez: I guess that would be better overall....

JordanMartinez:
I added javadoc that gives an example of what it can do, too.

TomasMikula:
:+1:

Issue #223, opened 2015-12-09
Overriding Drag Events


JordanMartinez:
I want to be override `StyledTextAreaBehavrior#mouseReleased`'s dragSelection handling. Instead of moving the selected text to where the user releases the mouse, I want to copy that text there and style it underneath a different style.

Right now, my two options are:
- rewrite all of the functionality of this code (enum and overriding default event handling) just to fix this one instance
- inline dragSelection into `StyledTextArea` itself and override only the MouseReleased event handling.

JordanMartinez:
On second thought, it'd be simpler to have a reference to `StyledTextAreaBehavior` within `StyledTextArea` so that the area itself can return the drag state.

JordanMartinez:
Resolved via 87d564bec713109209ce1d7a100d02e229a817c2

PR #224, opened 2015-12-09
Exposes DragState so subclasses can more easily override drag-related event handling

JordanMartinez:
Proposed fix to #223

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/DragState.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/DragState.java
@@ -1,0 +1,15 @@
+package org.fxmisc.richtext;
+
+/**
+ * Possible dragging states.
+ */
+public enum DragState {
+    /** No dragging is happening. */
+    NO_DRAG,
+
+    /** Mouse has been pressed, but drag has not been detected yet. */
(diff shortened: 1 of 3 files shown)


JordanMartinez:
Hmm... If this gets merged, then we've already started down the path of not finishing the MVC refactoring. If it doesn't, then I'm stuck rewriting the code you've already written or dropping a necessary feature...

I guess that means I now have a reason to finish the refactoring, lol...

TomasMikula:
Sorry for not being responsive. I've been buried under lots of other work.

I came up with `DragState` et al. without giving it too much thought and I'm not quite comfortable putting it to the public API (without giving it additional thought). An API that maybe exposes less implementation details and gives you the feature that you need would be to have

```java
Property<IntConsumer> onSelectionDrop();
```

that let's you override what to do on mouse release when dragging the selection. The `int` would be the insertion index where the mouse was released. The default would be  `i -> area.moveSelectedText(i)`, but could be overridden. How does it sound?

JordanMartinez:
No worries man! I figured you had finals-related stuff or something going on...

Ok, I'll look into that more and get back to you.

JordanMartinez:
See PR 228

PR #226, opened 2015-12-13
Remove unneeded things

JordanMartinez:
Apparently, I forgot to remove the VirtualizedScrollPane instance.


JordanMartinez:
Turns out this was what cause the display bug in the `2+ StyledTextArea objects can share the same StyledDocument` PR. I implemented these changes in that PR, so now I'm closing this one.

PR #228, opened 2015-12-19
Added `onSelectionDrop ` which allows client to override the handling…

JordanMartinez:
… of an event where the user selects some text, drags it elsewhere, and "drops" it by releasing the mouse


TomasMikula:
:+1:

PR #229, opened 2015-12-19
Remove TODO reminder and last remnants of VirtualizedScrollPane

JordanMartinez:
I'm removing the TODO reminder as its irrelevant.

It looks like we missed the VIrtualizedScrollPane component that shouldn't be apart of this anymore. So, I'm removing that before anything else gets done.


TomasMikula:
:+1:

PR #231, opened 2015-12-19
Allow two or more StyledTextAreas (View) to share the same StyledDocument (Model)

JordanMartinez:
I implemented the approach I described in #152 and for the most part it works. There are a few problems.
- If the user types in the original area quickly or if the user types even one character in the second line in the original area, the original area making the modification throws a StringIndexOutOfBoundsException.

Issue #220, opened 2015-12-04
PopupDemo regression


TomasMikula:
There is a regression in the PopupDemo (`gradle PopupDemo`), introduced after 0.6.10: the position of the popup does not update as I type.

JordanMartinez:
The regression doesn't exist as of 7277f9f8a99b6a4933e344a13fefcd474a041aac (the last commit before the skin was removed), but shows up somewhere after I removed the skin. When my work was merged in 45145ff2f409d4ede84fa04eacb2540ce84e40eb, this regression appears.

I've narrowed it down to this. When user types/deletes a character, the following path occurs:
````java
layoutChildren() // calls positionPopup()
---
positionPopup() { // calls getSelectionBoundsOnScreen()
switch(alignment.getAnchorObject()) {
        case CARET:
            bounds = getCaretBoundsOnScreen(); break;
        case SELECTION:
            bounds = getSelectionBoundsOnScreen(); break;
    }
    // at this point, bounds is null when it should have a value to which
    // popup would position itself.
    bounds.ifPresent(b -> { /* code */ })
}
---
getSelectionBoundsOnScreen() { 
// calls getCaretBoundsOnScreen 
// since selection's length is 0
    IndexRange selection = getSelection();
        if(selection.getLength() == 0) {
            return getCaretBoundsOnScreen();
        }
}
---
getCaretBoundsOnScreen() {
    // virtualFlow.getCellIfVisible(getCurrentParagraph()).isPresent() == false
    return virtualFlow.getCellIfVisible(getCurrentParagraph())
        .map(c -> c.getNode().getCaretBoundsOnScreen());
}
````

JordanMartinez:
Tomas, when a user selects text from the left and going right in the demo, should the popup be (almost) directly underneath where the caret is as it progresses to the right?

When I positioned the caret somewhere, pressed and held Shift, and used the right arrow to further select characters on the right, the popup didn't line up vertically with the caret.

JordanMartinez:
Changing `getCaretBoundsOnScreen` to the following makes it work. So, is this a regression in `Flowless`?
````java
Cell<Paragraph<S, PS>, ParagraphBox<S, PS>> cell = virtualFlow.getCell(getCurrentParagraph());
    if (virtualFlow.visibleCells().contains(cell)) {
        return Optional.of(cell.getNode().getCaretBoundsOnScreen())
    }
````

JordanMartinez:
[...]

Issue #210, opened 2015-11-20
CodeArea horizontal scrollbar not working


msarmc:
Tomas,
I am using a CodeArea in a Dialog. When there is no vertical scrollbar the horizontal scrollbar does not work. I modified JavaKeywordsAsync to demonstrate this. See the attached txt file (it wouldn't let me add a .java file)
[ScrollBarTest.txt](https://github.com/TomasMikula/RichTextFX/files/40074/ScrollBarTest.txt)

Thanks,
Mike

TomasMikula:
Hi Mike,

thanks, I can confirm the bug. It is actually reproducible with the demos, `Dialog` does not make any difference. I assume you are using the latest sources, because I cannot reproduce the bug with the latest released version [0.6.10](https://github.com/TomasMikula/RichTextFX/releases/tag/v0.6.10).

msarmc:
Tomas,

That’s curious. I ran JavaKeywordsAsync to see if it had the problem but it didn’t. Maybe it is because I let it open at its default width of 600 instead of sizing it down so a horizontal scrollbar would be needed. If 0.6.10 doesn't have the bug I'll try that version.

I am using code I downloaded Nov 16th.

Thanks,
Mike

msarmc:
Tomas,

I changed to use 0.6.10. This corrected the problem in one dialog. In the other dialog the horizontal scrollbar still did not work. In the second case CodeArea is in a SplitPane.

Mike

msarmc:
I apologize. One more thing. I should have checked this before the previous comment. The vertical scroll bar does not work either when in a SplitPane.

TomasMikula:
Does it only happen in a Dialog, or whenever in SplitPane?

msarmc:
I only tested in a Dialog. If I have time later today I'll test it in a SplitPane without using a dialog.

msarmc:
Using the snapshot I downloaded Nov 16 I tested using a SplitPane not in a dialog. The horizontal scroll bar does not work if there is no vertical scroll bar. If there is a vertical scroll bar both work.

TomasMikula:
Thanks, but I think we already established that there is a problem in the snapshot, regardless of the Dialog. I was interested to hear about the problem with the 0.6.10 version.

msarmc:
My apologies. When using 0.6.10 the horizontal scroll bar seems to work fine with the vertical scroll bar when in a SplitPane

JordanMartinez:
[...]

PR #232, opened 2015-12-21
2+ Views (areas) share the same Model (document)

JordanMartinez:
This is the same result of the #231 but with a cleaner commit history.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextArea.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextArea.java
@@ -448,6 +448,12 @@ public class StyledTextArea<S, PS> extends Region
      * content model
      */
     private final EditableStyledDocument<S, PS> content;
+
+    /**
+     * Usually used to create another area (View) that shares
+     * the same document (Model).
+     * @return this area's {@link EditableStyledDocument}
+     */
     protected final EditableStyledDocument<S, PS> getContent() {
         return content;
     }
--- a/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextArea.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextArea.java
@@ -380,8 +380,9 @@ public class StyledTextArea<S, PS> extends Region
     }
 
     // beingUpdated
-    public ObservableBooleanValue beingUpdatedProperty() { return content.beingUpdatedProperty(); }
-    public boolean isBeingUpdated() { return content.isBeingUpdated(); }
+    private final SuspendableNo beingUpdated = new SuspendableNo();
+    public ObservableBooleanValue beingUpdatedProperty() { return beingUpdated; }
+    public boolean isBeingUpdated() { return beingUpdated.get(); }
 
     // total width estimate
     /**
@@ -643,7 +644,7 @@ public class StyledTextArea<S, PS> extends Region
                 internalSelection, content.getParagraphs()).suspendable();
 
         omniSuspendable = Suspendable.combine(
-                // must be first, to be the last one to release
+                beingUpdated, // must be first, to be the last one to release
                 text,
                 length,
                 caretPosition,
@@ -1046,7 +1047,7 @@ public class StyledTextArea<S, PS> extends Region
      * Sets style for the given character range.
      */
     public void setStyle(int from, int to, S style) {
-        try (Guard g = suspend(omniSuspendable)) {
+        try (Guard g = content.beingUpdatedProperty().suspend()) {
             content.setStyle(from, to, style);
         }
     }
@@ -1055,7 +1056,7 @@ public class StyledTextArea<S, PS> extends Region
      * Sets style for the whole paragraph.
      */
     public void setStyle(int paragraph, S style) {
-        try (Guard g = suspend(omniSuspendable)) {
+        try (Guard g = content.beingUpdatedProperty().suspend()) {
             content.setStyle(paragraph, style);
         }
     }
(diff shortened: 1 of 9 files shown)


richtextfx/src/main/java/org/fxmisc/richtext/StyledTextArea.java
```diff
@@ -446,26 +449,41 @@ public void setStyleCodecs(Codec<S> textStyleCodec, Codec<PS> paragraphStyleCode
      * content model
      */
     private final EditableStyledDocument<S, PS> content;
+    protected final EditableStyledDocument<S, PS> getCloneDocument() {
```
TomasMikula: I would probably just call it `getDocument()`.

TomasMikula:
`addSuspendable`, `removeSuspendable` looks a lot like adding/removing special kind of listeners. This suggests a question whether we can reuse some existing listener mechanism. We can! We can have a `SuspendableNo beingUpdatedProperty()` on `EditableStyledDocument` and add a listener that suspends the `StyledTextArea`'s `omniSuspendable`. I added [`Suspendable.suspendWhen(condition)`](https://github.com/TomasMikula/ReactFX/blob/da3f7f36aa4833ba864047d914604603e19b6a62/reactfx/src/main/java/org/reactfx/Suspendable.java#L86) to ReactFX to make it easier, so that one can now do something like

```java
manageSubscription(omniSuspendable.suspendWhen(content.beingUpdatedProperty()))
```

richtextfx/src/main/java/org/fxmisc/richtext/StyledTextArea.java
JordanMartinez: I would but there's already a method with that name (see Line 336).

JordanMartinez:
Aye... that would be much cleaner. Did you make a new release so I can update the code?

richtextfx/src/main/java/org/fxmisc/richtext/StyledTextArea.java
TomasMikula: Right. What about `getContent()`?

TomasMikula:
Yes, I updated the snapshot release (`2.0-SNAPSHOT`).

JordanMartinez:
Great. I've had my first run-through it. If `EditableStyledDocument` has a `beingUpdatedProperty()` now, should I remove the `StyledTextArea`'s `beingUpdated` SuspendableNo ? (I know we ported it to ESD in the previous PR's exploration, but I reported it back to STA because it didn't add any value at that time.)

JordanMartinez:
I ported `beingUpdated` to ESD. How's that?

richtextfx/src/main/java/org/fxmisc/richtext/StyledTextArea.java
JordanMartinez: Sounds good. I've updated the method name.

TomasMikula:
> Great. I've had my first run-through it. If EditableStyledDocument has a beingUpdatedProperty() now, should I remove the StyledTextArea's beingUpdated SuspendableNo ? (I know we ported it to ESD in the previous PR's exploration, but I reported it back to STA because it didn't add any value at that time.)

No, we need to keep another `beingUpdated` in STA, in order to guarantee the property that no change can be observed while `beingUpdated` is `false`. This way, the user can be sure that when `beingUpdated` changes from `true` to `false`, all other changes have been propagated and are thus consistent with each other (e.g. caret position, current paragraph, length, ...). Note that the order in which registered listeners are invoked is undefined (neither in JavaFX nor ReactFX). If STA observes the same `beingUpdated` as the user, the user could get the notification first, before STA releases the pending notifications.

JordanMartinez:
Ah... I see. Fixed it.

TomasMikula:
:+1:

JordanMartinez:
Cool :smiley:

TomasMikula:
So I merged this, but the supported use is limited. For example, if the document is edited from both areas, undo will not work very well. One area's undo will be regarded as a new edit in the other area. It is a complex problem on its own, so for now I would just say that such a use is not supported.

JordanMartinez:
Dang.... Yeah, that's quite a problem.

On another note, when will there be a new mainstream release (meaning, not a snapshot release)?

JordanMartinez:
> if the document is edited from both areas, undo will not work very well. One area's undo will be regarded as a new edit in the other area. It is a complex problem on its own, so for now I would just say that such a use is not supported.

I'll open an issue for this as well, just so it's on file.

TomasMikula:
I can make a new release as soon as

1. all known regressions introduced by these refactorings are resolved.
2. all SNAPSHOT-versioned dependencies are released :)

JordanMartinez:
> all known regressions introduced by these refactorings are resolved.

What are the known regressions? Would this include the Undo-related problem mentioned above?

> all SNAPSHOT-versioned dependencies are released :)
:) What else needs to be done in Flowless and ReactFX before they get their releases?

TomasMikula:
> What are the known regressions?

#210, #220, #234 

> Would this include the Undo-related problem mentioned above?

That's not really a regression, since that was not supported at all before.

>  What else needs to be done in Flowless and ReactFX before they get their releases?

They might be good to go.

JordanMartinez:
I've also noticed a problem in the richtextfx's build.gradle file. It lists UndoFX as a dependency twice: the first, with its transitive dependencies overrules the second, one without transitive dependencies. I've had to comment out the first instance or else I get compile warnings.

TomasMikula:
Yeah, the first one should have been deleted when the second one was added. This is a mistake. Anyway, the `transitive = false` options will be removed when all dependencies use the same version of ReactFX as RichTextFX.

PR #238, opened 2016-01-05
Fixes PopupDemo regression

JordanMartinez:
Popup wouldn't reposition itself when user typed/deleted text

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextArea.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextArea.java
@@ -17,6 +17,7 @@ import java.util.function.IntUnaryOperator;
 import java.util.function.UnaryOperator;
 import java.util.stream.Stream;
 
+import javafx.application.Platform;
 import javafx.beans.binding.Binding;
 import javafx.beans.binding.Bindings;
 import javafx.beans.binding.BooleanBinding;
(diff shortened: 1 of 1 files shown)


TomasMikula:
Here's my understanding of the problem: When `positionPopup` (indirectly) asks for `getCellIfVisible`, the cell is not yet visible, because the `virtualFlow` has not been laid out yet. This is a simplified view of how the layout proceeds in JavaFX (which I only found out by reading the [JavaFX sources](http://hg.openjdk.java.net/openjfx/8u-dev/rt/file/f1c3eb85af4d/modules/graphics/src/main/java/javafx/scene/Parent.java#l1070)):

```java
void layout() {
    layoutChildren();
    for(Node node: getChildren()) {
        if(node instanceof Parent) {
            ((Parent) node).layout();
        }
    }
}
```

`virtualFlow.getCellIfVisible` is called from `layoutChildren`, i.e. before `virtualFlow.layout()`.

To confirm my hypothesis that this is the cause of the problem, instead of your fix, you can try to call `virtualFlow.layout()` before `positionPopup()` and see if that fixes the problem as well.

As a general fix, though, I think methods of `VirtualFlow` whose result depends on the layout, should themselves call `layout()` as the first thing. Such methods would include `getCellIfVisible`, `visibleCells` and `hit`. The fact that they first call `layout()` should be documented in their Javadoc.

You should not be too concerned about `layout()` being called multiple times, because it is a no-op if the layout is not dirty (the `layout()` code above is simplified, in fact it also checks whether the layout is dirty).

TomasMikula:
Also, this means that the refactoring uncovered a deficiency in Flowless, rather than there being a problem with your code.

JordanMartinez:
You are correct in your hypothesis. 
````java
// changing this code
Platform.runLater(() -> positionPopup(popup, alignment, adjustment));
// to this code will correct the problem
virtualFlow.layout();
positionPopup(popup, alignment, adjustment);
````

JordanMartinez:
I've opened up an issue for this in Flowless. We'll fix it there.

TomasMikula:
:+1:

Issue #216, opened 2015-11-27
IllegalArgumentException when undo-ing then redo-ing


benedictleejh:
Steps to reproduce:

1. enter text "testtest"
2. Highlight the first "test"
2. style (Press the style button) the first "test"
3. Highlight the second "test"
3. style the second "test"
4. put selection point between the two "test"
5. type shift-space, space, space
6. undo, redo

Repo with code: https://github.com/benedictleejh/richtextfx-error

TomasMikula:
Thanks for reporting. This is an example of how stricter checking in UndoFX revealed a bug elsewhere.

A little analysis of the problem:

The problem comes down to the fact that an empty paragraph contains one text segment (`StyledText`) of length 0, but there is a style associated with that text segment nevertheless. The two change events, expected and received, differ in the style of that empty string, and are thus considered different, even though their effect is indistinguishable.

I think the right solution is to represent an empty `Paragraph` as having no text segments. Though there are probably places where we use the style of an empty segment, so we will have to find a way to not need that style.

JordanMartinez:
Here's the Java version of benedictleejh's code:
````java
import javafx.application.Application;
import javafx.scene.Scene;
import javafx.scene.control.Button;
import javafx.scene.input.KeyEvent;
import javafx.scene.layout.BorderPane;
import javafx.stage.Stage;
import org.fxmisc.richtext.InlineStyleTextArea;

import java.util.function.Function;

public class BugTest extends Application {

    public static void main(String[] args) {
        launch(args);
    }

    class BoldUnderline {
        boolean styled;
        String style = "-fx-font-weight: bold; -fx-underline: true;";
        String toCss() { return (styled) ? style : ""; }

        BoldUnderline(boolean styled) {
            this.styled = styled;
        }
    }
    
    InlineStyleTextArea<BoldUnderline, String> richTextBox = new InlineStyleTextArea<>(
            new BoldUnderline(false), BoldUnderline::toCss,
            "", Function.identity()
            );
    {
        richTextBox.setPrefSize(400, 300);
        richTextBox.addEventFilter(KeyEvent.KEY_TYPED, (KeyEvent e) -> {
            if (e.isShiftDown() && e.getCharacter().equals(" ")) {
                richTextBox.setUseInitialStyleForInsertion(true);
                richTextBox.insertText(richTextBox.getCaretPosition(), " ");
                richTextBox.setUseInitialStyleForInsertion(false);

                // consume event to prevent double space insertion
[...]

PR #240, opened 2016-01-15
Bug fixed 216

JordanMartinez:
I think the final commit (4d86628) could be done better. However, I'm not sure how to get around a plumbing issue. 
If a new `EmptyParagraph` is created anywhere outside of `EditableStyledDocument`, it doesn't have the correct style object for its `emptyTextStyle` (e.g. StyledDocumentBase, NormalParagraph, ReadOnlyStyledDocument). Additionally, its `paragraphStyle` isn't always correct either when it is created in some method in StyledDocument and ReadOnlyStyledDocument. 

I think the only good solution is to remove `EmptyParagraph`'s `emptyTextStyle` field and use the StyledDocument classes to check the Paragraph type before returning a style object, such as:
````java
class StyledDocumentBase {

    StyledDocumentBase(L paragraphs, S initialStyle, PS initialParagraphStyle) {
        // normal creation code
        this.initialStyle = initialStyle;
        this.initialParagraphStyle = initialParagraphStyle;
    }

    Paragraph someOtherMethodThatCreatesANewEmptyParagraph() {
        return new EmptyParagraph(initialParagraphStyle);
    }

    S getStyleOfChar/getStyleAtPosition(int position) {
        Paragraph<S, PS> par = paragraphs.get(position);
        return par instanceof EmptyParagraph
            ? initialStyle
            : par.getStyleOfChar/getStyleAtPosition(position);
    }
}
````
Doing the above would require another change. `ClipboardActions` has a default method that creates a `ReadOnlyStyledDocument`, but it doesn't have any reference within itself for getting the two styles for its new constructor. So, it'd also need to have two getters for the styles. Since it's name is ClipboardActions, I wasn't sure if something should further extend that. Possibly an interface called `StyledClipboardActions` ?

### DIFF
--- a/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/richtext/RichText.java
+++ b/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/richtext/RichText.java
@@ -35,7 +35,7 @@ import javafx.stage.Stage;
 
 import org.fxmisc.flowless.VirtualizedScrollPane;
 import org.fxmisc.richtext.AreaFactory;
-import org.fxmisc.richtext.Paragraph;
+import org.fxmisc.richtext.NormalParagraph;
 import org.fxmisc.richtext.StyleSpans;
 import org.fxmisc.richtext.StyledTextArea;
 import org.reactfx.SuspendableNo;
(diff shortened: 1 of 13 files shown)


JordanMartinez:
Once fully implemented, I think some tests should be written for the Paragraph methods that return a Paragraph to insure that EmptyParagraph is returned instead of a NormalParagraph whose length is 0 because I'm pretty sure that part isn't error-free yet in this branch.

JordanMartinez:
Closing as `EmptyParagraph`/`NonEmptyParagraph` approach to solving bug is "orthogonal" (#216).

PR #243, opened 2016-01-23
Extracted Model content from StyledTextArea (view) into its own model class (StyledTextAreaModel)

JordanMartinez:
So, I got to it sooner. (It's also has a somewhat cleaner commit history).


richtextfx/src/main/java/org/fxmisc/richtext/TextEditingArea.java
```diff
@@ -137,31 +137,6 @@
...
-     *                *
-     ******************/
-
-    /**
-     * Indicates whether this text area can be edited by the user.
-     * Note that this property doesn't affect editing through the API.
-     */
-    boolean isEditable();
-    void setEditable(boolean value);
-    BooleanProperty editableProperty();
```
TomasMikula: We should preserve the Javadoc comment by moving it to StyledTextArea.

richtextfx/src/main/java/org/fxmisc/richtext/TextEditingArea.java
```diff
@@ -137,31 +137,6 @@
...
-
-    /**
-     * When a run of text exceeds the width of the text region,
-     * then this property indicates whether the text should wrap
-     * onto another line.
-     */
-    boolean isWrapText();
-    void setWrapText(boolean value);
-    BooleanProperty wrapTextProperty();
```
TomasMikula: We should preserve the Javadoc comment by moving it to StyledTextArea.

richtextfx/src/main/java/org/fxmisc/richtext/TextEditingArea.java
JordanMartinez: Oops. Good point!

TomasMikula:
Looking good :+1: Thanks!

JordanMartinez:
☺

PR #246, opened 2016-01-25
Switched generic style objects order: paragraph comes before text (<S, PS> ==> <PS, S>)

JordanMartinez:


JordanMartinez:
I noticed on [EditableStyledDocument: Line 168](https://github.com/TomasMikula/RichTextFX/pull/246/files#diff-6059024eb727356aea7de04d69f6ef11R168) that I get a compile error now that `get(Paragraph)StyleForInsertionAt(int position)` method has been removed, so I switched to `getParagraphStyleAtPosition()` instead. Is that in accordance with your javadoc?

TomasMikula:
Great, thanks!

Issue #233, opened 2015-12-28
Add support for nonlinear undo/redo (clones don't currently handle undo/redo correctly)


JordanMartinez:
#232 (solution to #152) has been implemented. However, as noted by Tomas:

> If the document is edited from both areas, undo will not work very well. One area's undo will be regarded as a new edit in the other area. It is a complex problem on its own, so for now I would just say that such a use is not supported.

JordanMartinez:
I think the only way to account for this is to use one `UndoManager` that is shared between multiple `StyledTextAreaModel`s. However, the tricky part lies in making sure all the other Models' `getUndoManager()` returns the same `undoManager` after one of them sets it to something else. 

So, it seems like the best way to get around this is to wrap the UndoManager in wrapper class. See my [betterCloningUndoing](https://github.com/JordanMartinez/RichTextFX/commit/a02c3a7646d28ac1235a9ea22002316d13c68cff) branch for an example of this.

However, there is an additional problem to solve. As of now, when the undo manager is created, it is passed a lambda that calls the `replace()` method of the `StyledTextAreaModel` that initially constructed it when a change is re-done. 

So, if there are three `StyledTextAreaModel`s, two of which are clones of the original....
````
StyledTextAreaModel initial;
StyledTextAreaModel clone1;
StyledTextAreaModel clone2;
````
....when `clone1` or `clone2` call `getUndoManager().undo()`, it is `initial` whose `replace()` method is called, not either of the clones.

So, what happens when initial gets disposed? When `redo()` is called, the disposed and thus garbage-collected `initial` won't exist and the attempt to call its `replace()` method will result in a NullPointerException.

**Comment by JordanMartinez** [comment]
The only way around this second part is if the `apply` part of `createRich/PlainTextChange(UndoManagerFactory factory)` doesn't rely upon `StyledTextAreaModel`'s code in any way. In this way, it would need to actively call `EditableStyledDocument`'s code and any caret/selection values would need to be updated via the code that updates these values when `EditableStyledDocument`'s plainTextChange emits a new change.

**Comment by TomasMikula** [comment]
The same effect as by sharing an UndoManager can also be achieved by only one of the areas having a real UndoManager and the others having a fake one:

```java
area.setUndoManager(UndoManagerFactory.zeroHistoryFactory())
```

Then no undo/redo will be available on the other areas, but the keyboard shortcuts can be overridden to trigger undo/redo on the area with the true UndoManager.

[...]

PR #249, opened 2016-01-26
Make Clones share the same UndoManager

JordanMartinez:
See my comments in #233.


JordanMartinez:
With this approach, do [these lines that update the caret and selection post-replacement](https://github.com/TomasMikula/RichTextFX/blob/master/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextAreaModel.java#L646-L647) still need to be there?

JordanMartinez:
I'm closing this because I think #251 makes more sense in the long-term.

PR #250, opened 2016-01-26
Minor code refactoring to get rid of some compiler suggestions:

JordanMartinez:
- Replace "(arg1, arg2) -> arg1.method(arg2)" lambdas with method references
- Removed explicit type since it can be inferred (e.g. "<PS, S>" --> "<>")
- Removed unneeded semi-colons
- Removed unused Type parameter "<C>"
- Removed `initialParagraphStyle` from `createCell()` since it isn't used there.
- Made `omniSuspendable` a local variable since suspending EditableStyledDocument's `beingUpdatedProperty()` suspends all clones' Suspendables now instead of `omniSuspendable`


TomasMikula:
:+1:

PR #263, opened 2016-02-26
Allow Clone subclassing

JordanMartinez:
See 2nd commit's details as I had to do a lot in that one commit.


JordanMartinez:
Due to verbosity, should the Model's names be reduced by removing the `TextArea` part? For example:
- StyledTextAreaModel should remain the same
- StyleClassedTextAreaModel --> StyleClassedModel
- CodeAreaModel --> CodeModel
- InlineCssTextAreaModel --> InlineCssModel

JordanMartinez:
Since this PR assumed that `StyledTextAreaModel` would be subclass-able and since it is clear that that isn't the best approach, I'm closing this PR.

PR #269, opened 2016-03-17
Renamed confusing names for selection colors to more understandable names

JordanMartinez:
This will lead to some breaking changes in people's CSS files, but since we're already breaking stuff in the new release, might as well do it here.


TomasMikula:
I like your naming better, but maybe there is some value in keeping the names consistent with [JavaFX's](https://docs.oracle.com/javase/8/javafx/api/javafx/scene/doc-files/cssref.html#textinputcontrol)?

JordanMartinez:
Ah... that explains why you used those names in the first place.

PR #275, opened 2016-03-30
Add formal grammar to syntax highlighter

Groostav:
This is a proof-of-concept to integrate ANTLR and a grammar with the syntax highlighter.

My problem with the existing approach employed by the `JavaKeywords` demo is that it uses a very simple string scanner, making it difficult to specify that anything other than a particular occurrence of a word needs to be highlighted. For more complex cases you could use a regex, but getting that exactly correct is often subtly difficult; simple highlighting problems you might think you can solve with a regex are provably unsolvable. 

So the heavy-hitter is to use a full-bore LR or LL parser such as yacc, bison, or (as is the case in this PR), ANTLR.

A simple screen shot:

antlr grammar sample

From the commit message:

> Added runnable sample, looks pretty slick
> - with ANTLR you can make your syntax highlighter truely-IDE like. Since ANTLR has baked in its own concepts of text-ranges, this work represents a pretty direct mapping of ANTLR's text ranges for tokens to richtextfx's text ranges for CSS. Its rare that two API's map so directly onto each other, and it was a joy to work with.
> - I look at this as a proof-of-concept, and an attempt to drum-up interest. A huge number of todo's around validation, and I've heavily emphasized use of these types from FXML and then haven't included any.
> - I don't think richtextfx wants to deploy with a dependency on ANTLR, so I might try re-tooling this to do the nasty thing of having the dependency only resolved at runtime. In other words, use `import antlr.stuff`, but don't ship with the antlr runtime, and do a static check to see if the programmer using richtextfx has it, failing with a nice message if they don't. Maybe there's a better strategy.
> - Really, the most versatile thing we can do with the `highligher` type is to have it take a `Predicate<Node>` that simply tells you if this is the node you're looking for.
> - might try hacking that into event framework, such that you could do something like `<Highlighter predicate="#isWhatever"/>` and have a corresponding method on the controller, but it would likely be a very (_very_) serious hack.
> - for myself, I need to spend more time with gradle, its clear that its auto-magic can make my life easier, I just need to learn it. Needless to say I'm a gradle newbie, and I couldn't find any really established repo-layouts for generated code, so I rolled my own. I fully expect to get butchered for this in PR.
> - with respect to style, I'm not sure why I went so insane with `java.util.functional` this time around, I'm happy to refactor into more traditional if statements if so desired.

So, tasks to do before my next PR, assuming you guys are interested in this work:
- [ ] come up with an automated test plan
- [ ] add example for bracket highlighting
- [ ] add validation of filters such that if the user creates a highlight that cant hit anything we raise a warning
- [ ] replace heavy use of reflection with simpler (functional?) types, wrap that for FXML
- [ ] the extension hierarchy here is getting strange but I'm not much of a css pro so I'm not sure how we can refactor this toward some kind of composition or bridge. Will investigate.


TomasMikula:
This is pretty cool! I've been hoping someone would do this a long time ago.

The regex-based parsing in the JavaKeywords demo is meant just as a demonstration of what's (also) possible with RichTextFX, not to suggest people should do parsing that way.

I would like to keep the concern of parsing separate from RichTextFX, in order to
 - keep the scope of RichTextFX limited; and
 - not to favor this parser or that parser.

Would you like to create a stand-alone project with a dependency on RichTextFX? I sure would link to it from RichTextFX.

Groostav:
So, rather than be responsible and write tests and a working build system, I broke the build system, wrote no tests, and added a mountain of features and code. :+1: 

@TomasMikula, I think I'm coming up on forking this into a repo, I'd appreciate any input you're willing to give, in particular:

- any feedback on easy ways I can get this to be more ~~reactive~~ react-UI like is appreciated
- thoughts on names or needed features
- I had to check in my generated antlr code because I couldn't figure out how to get the gradle antlr task working, this annoys me to no end. Is there something obvious that I'm missing?

Groostav:
>- I'm very much a newbie to CSS, and the 'free' css-style-class generation scheme for tokens (and terminal nodes, if i implement it) is likely to generate _a fair number_ of css style classes. I'm wondering if there's an elegant way to alleviate some of the clobbering/no-such-style problems that are likely to result
> - would it be koscher (or is it even technically possible possible) to validate a css style name at runtime. In other words, before commiting a text range to a particular style class name, check to see if the style class actually exists, and throw it away if it doesn't. 
> - use a parenting relationship with style classes, such that each type might have its own css-ancestry, something like `lexer.[token]` vs `parser.[rule]` or `[grammer].lexer.[token]` (eg `javish-math.lexer.var` and `javish-math.parser.variable`) 

Looking at the [W3C's CSS syntax guide](http://www.w3schools.com/css/css_syntax.asp), I'm SOL. _briefly_ summarizing:

- identifier selection `#label` assumes that the value of `label` will be _unique on the page_, and nothing about my flow can make that guarantee.
- class selection `.label` works best, as per RichTextFX, however nested dots (like ip-style dotted-quad) would require [a hack on the language](http://stackoverflow.com/questions/3447329/styling-elements-with-a-dot-in-the-class-name).
- using an element selector in conjunction with a class selector, eg `label1.label2` (like `[language]-[type].[rule|token]` eg `babel-lexer.sum`) seems like the bestidea, but is tough to control programmatically. 

I could simply allow the users to specify a class-prefix through a string property. 

Hmm. More research here needs to be done, maybe a proof-of-concept for the above-point number 3.

Groostav:
Also, there was something I could write that was illegal under the supplied grammar that _didn't_ result in any nodes getting hilighted, when I thought they should've, unfortunately I wrote that input down in a comment and now I cant find it. didn't mark it with TODO for intelliJ either. Stupid. _Something about brackets..._

JordanMartinez:
@Groostav I'm closing this PR because it's outside the scope of this project. However, have you finished creating a separate repo that does the above? If so, we can update the ReadMe to link to it.

Issue #252, opened 2016-01-27
Cloning issues arise with subclasses


JordanMartinez:
I've run into another issue revolving around cloning.

Currently, each `StyledTextArea` (the View) creates its own `StyledTextAreaModel` (the Model). I'm guessing that the model itself is not passed to the View in order to prevent multiple views sharing the same model. (Each model is tasked with storing the data for where the caret position and selection values should be. If two views share the same model, it defeats the whole point).

However, this also creates another issue: `StyledTextAreaModel` cannot be subclassed. The following problems arise:
- Since `StyledTextAreaModel`'s constructor takes an `EditableStyledDocument` as an argument, which is currently package-private, not public, subclasses can't have constructors that are used to clone those subclassed models since they can't fulfill `super`'s constructor requirement.
- Since `StyledTextArea` constructs its own `StyledTextAreaModel`, even if one subclassed `StyledTextAreaModel`, they would not be able to create a `StyledTextArea` (a view) that uses such a model.

So, it seems like `EditableStyledDocument` should be made public, and a StyledTextArea would be updated to below:
````java
public class StyledTextArea<PS, S, Model extends StyledTextAreaModel<PS, S>> {
    private final Model model;
    StyledTextAreaModel(
        BiFunction<?, ?> applyStyle, 
        BiFunction<?,?> applyParagraphStyle, 
        Model model
    ) {
        this.model = model;
        // the remaining constructor code....
    }
}
````
With a subclass of `StyledTextAreaModel` like the following:
````java
public class AnotherModel<PS, S> extends StyledTextAreaModel<PS, S> { /* code */ }
````
...and a subclass of `StyledTextArea` like the following:
````java
public class AnotherArea<PS, S, AnotherModel<PS, S>> extends StyledTextArea<PS, S> { /* code */ }
````
To create an area, one would use the following code:
````java
AnotherModel<PS, S> model = new AnotherModel<>(/* args... */);
AnotherArea<PS, S, AnotherModel<PS, S>> area = new AnotherArea<>(model);
````
or use the following code:
````java
AnotherModel<PS, S> model = new AnotherModel<>(/* args... */);
StyledTextArea<PS, S, AnotherModel<PS, S>> area = new StyledTextArea<>(model);
````

I think this was the conclusion I reached when I was working on #243.

JordanMartinez:
[...]

Issue #244, opened 2016-01-23
Allow custom implementation of EditableStyledDocument


JordanMartinez:
Tomas originally said in #181
> I eventually want to support custom implementations of what is now called `EditableStyledDocument`....

JordanMartinez:
@TomasMikula 
Do you care to elaborate more on what would be customizeable?

TomasMikula:
My idea is that `EditableStyledDocument` should be an interface, something like

```java
public interface EditableStyledDocument<PS, S> {
    EventStream<RichTextChange<PS, S>> changes();
    ObservableList<Paragraph<PS, S>> getParagraphs();
    ReadOnlyStyledDocument<PS, S> subDocument(Position from, Position to);
    void replace(Position from, Position to, ReadOnlyStyledDocument<PS, S> replacement);
}
```

The implementation can use any represent of a document (e.g. a parse tree), as long as it publishes changes, provides a view as a list of paragraphs, and handles edits.

JordanMartinez:
That sounds good! Should `StyledDocumentBase` be public too then? Additionally, shouldn't the interface also include `EventStream<PlainTextChange> plainTextChanges();`?

TomasMikula:
> Should `StyledDocumentBase` be public too then?

For now I would say no.

> Additionally, shouldn't the interface also include `EventStream<PlainTextChange> plainTextChanges()`?

Yeah, but with a default implementation based on rich text changes. Currently the implementation tries to optimize for the case when plain text changes are observed, but rich text changes are not. The current implementation tries to avoid creating instances of `RichTextChange` at all when no-one is observing them. I don't think the actual savings make any substantial difference, but come with a price of higher complexity of the code.

JordanMartinez:
> Yeah, but with a default implementation based on rich text changes.

Ok.

JordanMartinez:
Just another thought: since there is now an interface for `EditableStyledDocument`, should there be a public interface for its counterpart: `ReadOnlyStyledDocument`? If we're going to allow custom ESD, it seems necessary to do the same for its counterpart.

TomasMikula:
> If we're going to allow custom ESD, it seems necessary to do the same for its counterpart.

Why?

JordanMartinez:
On second thought... I guess that isn't necessary. :-)

JordanMartinez:
[...]

PR #276, opened 2016-03-31
Allow subclasses of StyledTextArea (and its package's subclasses) to clone themselves

JordanMartinez:
I know we made `StyledTextAreaModel` package-private in the last commit. However, I came across an issue relating to cloning subclasses. Because the model is no longer public, a subclass cannot access the model to get the `EditableStyledDocument` as its clone's argument. The below approach would work if `StyledTextAreaModel` was public but `EditableStyledDocument` was not public:
````java
// assuming this constructor exists:
public SomeArea(SomeArea areaToClone) {
  // super = InlineCssTextArea(EditableStyledDocument<String, String> doc);
  super(
    area
      // since the model is package-private, "getModel()" isn't accessible!
      .getModel() 
      .getContent()
  );
}

// this code will not work if the model is package-private
SomeArea area = // area that extends InlineCssTextArea
SomeArea clone = new SomeArea(area);
````

However, if that approach is followed (`EditableSyledDocument` is not public) another issue arises: one cannot use it as a Constructor parameter....
````java
public class SomeArea extends InlineCssTextArea {
                // Compiler: cannot resolve object "EditableStyledDocument"
    public SomeArea(EditableStyledDocument<String,String> doc) {
        super(doc);
    }

}
````
... which means subclasses of subclasses cannot clone themselves
````java
public class AnotherArea extends SomeArea {

    public AnotherArea(EditableStyledDocument<String, String> doc) {
        // Compiler: "this constructor doesn't exist: 'SomeArea(EditableStyledDocument)' "
        super(doc);
}
````

So, I made `StyledTextAreaModel` **public** (to get around this issue) **and final** (so that it still can't be subclassed). Additionally, I made the methods relating to styles (e.g. `getInitialTextStyle()`) public so a proper clone can be created.
Finally, I made `EditableStyledDocument` public but did not change anything else (it was already `final` so there was no need to change anything there).

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocument.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocument.java
@@ -30,7 +30,7 @@ import org.reactfx.value.Var;
  * on styled text, but not worrying about additional aspects such as
  * caret or selection.
  */
-final class EditableStyledDocument<PS, S> extends StyledDocumentBase<PS, S, ObservableList<Paragraph<PS, S>>> {
+public final class EditableStyledDocument<PS, S> extends StyledDocumentBase<PS, S, ObservableList<Paragraph<PS, S>>> {
 
     /* ********************************************************************** *
      *                                                                        *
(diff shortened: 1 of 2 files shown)


JordanMartinez:
Closing due to recent comments in #252  & #244

PR #277, opened 2016-03-31
Custom EditableStyledDocument

JordanMartinez:
I used `EditableStyledDocument`'s methods as the base for the new interface.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocument.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocument.java
@@ -7,7 +7,22 @@ import org.reactfx.EventStream;
 import org.reactfx.SuspendableNo;
 import org.reactfx.value.Val;
 
+/**
+ * Content model for {@link StyledTextArea}. Implements edit operations
+ * on styled text, but not worrying about additional aspects such as
+ * caret or selection, which are handled by {@link StyledTextAreaModel}.
+ */
 public interface EditableStyledDocument<PS, S> extends StyledDocument<PS, S> {
+
+    /* ********************************************************************** *
+     *                                                                        *
+     * Observables                                                            *
+     *                                                                        *
+     * Observables are "dynamic" (i.e. changing) characteristics of an object.*
+     * They are not directly settable by the client code, but change in       *
+     * response to user input and/or API actions.                             *
+     *                                                                        *
+     * ********************************************************************** */
 
     String getText();
     ObservableValue<String> textProperty();
@@ -18,6 +33,12 @@ public interface EditableStyledDocument<PS, S> extends StyledDocument<PS, S> {
     ObservableList<Paragraph<PS, S>> getParagraphs();
 
     ReadOnlyStyledDocument<PS, S> snapshot();
+
+    /* ********************************************************************** *
+     *                                                                        *
+     * Event streams                                                          *
+     *                                                                        *
+     * ********************************************************************** */
 
     default EventStream<PlainTextChange> plainChanges() {
         return richChanges()
@@ -35,6 +56,15 @@ public interface EditableStyledDocument<PS, S> extends StyledDocument<PS, S> {
     StyledDocument<PS, S> subSequence(int start, int end);
 
     StyledDocument<PS, S> subDocument(int paragraphIndex);
+
+    /* ********************************************************************** *
+     *                                                                        *
+     * Actions                                                                *
+     *                                                                        *
+     * Actions change the state of the object. They typically cause a change  *
+     * of one or more observables and/or produce an event.                    *
+     *                                                                        *
+     * ********************************************************************** */
 
     void replace(int start, int end, StyledDocument<PS, S> replacement);
 
--- a/richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocument.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocument.java
@@ -42,7 +42,6 @@ public interface EditableStyledDocument<PS, S> extends StyledDocument<PS, S> {
 
     default EventStream<PlainTextChange> plainChanges() {
         return richChanges()
-                // map is used to prevent code repetition: StyledDocument#getText()
                 .map(c -> new PlainTextChange(c.position, c.removed.getText(), c.inserted.getText()))
                 // filter out rich changes where the style was changed but text wasn't added/removed
                 .filter(pc -> !pc.removed.equals(pc.inserted));
--- a/richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocument.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocument.java
@@ -19,7 +19,13 @@ public interface EditableStyledDocument<PS, S> extends StyledDocument<PS, S> {
 
     ReadOnlyStyledDocument<PS, S> snapshot();
 
-    EventStream<PlainTextChange> plainChanges();
+    default EventStream<PlainTextChange> plainChanges() {
+        return richChanges()
+                // map is used to prevent code repetition: StyledDocument#getText()
+                .map(c -> new PlainTextChange(c.position, c.removed.getText(), c.inserted.getText()))
+                // filter out rich changes where the style was changed but text wasn't added/removed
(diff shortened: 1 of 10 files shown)


JordanMartinez:
I didn't include `replaceText(int, int, String)` because that seems to be handled by `StyledTextAreaModel` now and you seem to only have it in `EditableStyledDocumentImpl` as a convenience method to use when testing that class

JordanMartinez:
I'd also like to rename `plainTextChanges` to `plainChanges` in `StyledTextAreaModel` and `StyledTextArea`. I know it introduces a breaking change, but it seems to be a more consistent naming scheme (e.g. plainChanges and richChanges).

richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocument.java
```diff
@@ -19,7 +19,12 @@
 
     ReadOnlyStyledDocument<PS, S> snapshot();
 
-    EventStream<PlainTextChange> plainChanges();
+    default EventStream<PlainTextChange> plainChanges() {
+        return richChanges()
+                // only allow changes that added/removed text; exclude style changes
+                .filter(c -> c.insertedLength() != 0 || c.removedLength() != 0)
+                .map(c -> new PlainTextChange(c.position, c.removed.getText(), c.inserted.getText()));
+    }
```
TomasMikula: If there is a rich-text change, then `insertedLength` or `removedLength` is non-zero already. You probably want to first `map` and then filter out plain-text changes where `removed.equals(inserted)`.

JordanMartinez: Wouldn't it be more efficient to filter it first? Otherwise, we're creating unneeded `PlainTextChange`es:
````java
default EventStream<PlainTextChange> plainChanges() {
    return richChanges()
            // only allow changes that added/removed text; exclude style changes
            .filter(c -> !c.inserted.getText().equals(c.removed.getText()))
            .map(c -> new PlainTextChange(c.position, c.removed.getText(), c.inserted.getText()));
}
````

Comment (TomasMikula) [inline richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocument.java]: Yes, slightly. On the other hand there is a slight repetition of code in calling `getText()`.

And if you want to avoid an intermediate `EventStream`, use `filterMap`.

Comment (JordanMartinez) [inline richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocument.java]: > Yes, slightly. On the other hand there is a slight repetition of code in calling getText().

Yeah... I noticed that, too. One way to get around it :stuck_out_tongue_winking_eye: 
````java
    default EventStream<PlainTextChange> plainChanges() {
        return richChanges()
                // exclude style changes
                .map(c -> {
                    String inserted = c.inserted.getText();
                    String removed = c.removed.getText();
                    if (!inserted.equals(removed)) {
                        return new PlainTextChange(c.position, removed, inserted);
                    } else {
                        return null;
                    }
                })
                .filter(c -> c != null);
    }
````

Comment (TomasMikula) [inline richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocument.java]: I don't care so much, do what you think is right ;)

Comment (JordanMartinez) [inline richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocument.java]: I'll do the most recent approach and explain in comments why it's used rather than the other approaches: less memory usage and less code repetition

Comment (TomasMikula) [inline richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocument.java]: I wouldn't care too much about optimizing one object allocation, since there are already other much higher costs paid anyway.

Comment (JordanMartinez) [inline richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocument.java]: With the most recent approach used, does that lead to higher costs due to the EventStreams?

Comment (TomasMikula) [inline richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocument.java]: An extra `EventStream` adds a few stack frames to the call stack when propagating the event. But even by the time an edit reaches the `plainChanges()` stream, so many objects have been allocated and dismissed that one more or less doesn't make any difference. Therefore, I would go for code clarity.

Comment (JordanMartinez) [inline richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocument.java]: So you do care :wink: 

Ok. I didn't factor that into consideration. Code clarity it is.

Comment (TomasMikula) [inline richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocument.java]: I guess I do care, when I see how much noise 4e6ce6b9 adds in the name of saving one object allocation per change. Also, code comments can go out of sync with code very easily, so self-documenting code is better :)

**Comment by JordanMartinez** [comment]
On another note, does `StyledTextArea#getModel()` need to be protected? I feel like it should be package-private now (it was used to get the underlying document, but we have a better API for that now)

**Comment by TomasMikula** [comment]
Agreed.

File: richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocumentImpl.java [hunk 2]
```diff
@@ -26,9 +26,7 @@
 import org.reactfx.value.Var;
 
 /**
- * Content model for {@link StyledTextArea}. Implements edit operations
- * on styled text, but not worrying about additional aspects such as
- * caret or selection.
+ * Provides an efficient implementation of {@link EditableStyledDocument}
```
TomasMikula: I don't think it is very efficient :) Could be improved.

JordanMartinez:
Whoops... that commit message should be "package-private" not private...

richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocumentImpl.java
JordanMartinez: Lol... that was my assumption since you wrote it :smile:

JordanMartinez:
If there's nothing else that needs to be fixed, I'd like to open a new PR that fixes the last erroneous commit.

TomasMikula:
You can also

    git rebase -i HEAD~14

to reorder and squash some of the last 14 commits and then

    git push -f

to rewrite the history of this PR.

JordanMartinez:
Sorry that it took so long to update the PR. I almost finished updating the history, but then I had to leave for something and wasn't able to complete the update.

TomasMikula:
No worries, my reaction periods tend to be much longer ;)

richtextfx/src/main/java/org/fxmisc/richtext/EditableStyledDocument.java
```diff
@@ -19,7 +19,13 @@
 
     ReadOnlyStyledDocument<PS, S> snapshot();
 
-    EventStream<PlainTextChange> plainChanges();
+    default EventStream<PlainTextChange> plainChanges() {
+        return richChanges()
+                // map is used to prevent code repetition: StyledDocument#getText()
```
TomasMikula: Out of context of our previous discussion, this comment does not make much sense, does it? ;)

JordanMartinez: I suppose not... should it be removed entirely?

JordanMartinez:
Is there anything else to do before merging this PR? I still haven't heard back from you about the "out of context" mapping comment.

TomasMikula:
Yeah, I think that comment is better removed entirely.

JordanMartinez:
The commit message gives more context to the comment removal. So, that should suffice for now.

TomasMikula:
:+1: 

Going to merge now. Though `EditableStyledDocument` as it stands now is a hell of an interface asking someone to implement. It should be gradually simplified.

JordanMartinez:
> Though EditableStyledDocument as it stands now is a hell of an interface asking someone to implement. It should be gradually simplified.

That is true.... On one hand, `StyledTextAreaModel` does seem to make use of all of them at some point or another (whether directly or through some other call within a method call). Still, I didn't check to insure that it was as simple as `StyledTextAreaModel` would currently allow.
On another hand, making `StyledDocumentBase` public (or perhaps an `EditableStyledDocumentBase`) could help with that.

JordanMartinez:
When you have time, could you update the snapshot release?

TomasMikula:
Updated.

JordanMartinez:
Thanks! :+1:

PR #281, opened 2016-04-06
Fixed the command to run the RichText demo

adrapereira:


TomasMikula:
Hmm, I guess I updated the README prematurely. In the new snapshot version, the RichText demo _is_ in sub-package `demo.richtext`, but in 0.6.10, it is not.

TomasMikula:
Version 0.7-M1 has been released, and the command now matches the class name in that version.

Issue #259, opened 2016-02-20
What should be in the 0.7 release and ReadMe Updates


JordanMartinez:
`0.6.10` was released on Oct 12. 4 months later, we've made the following changes to RichTextFX since that release:
- Skin related
  - Implemented paragraph styles (text alignment, line background color, copy & paste styles)
  - Exposed estimated scrolling API
  - Allow better access to the skin (no longer using the Control, Skin, Behavior paradigm)
  - Separated the `ScrollPane` aspect of `StyledTextArea` from the actual area itself
  - Allow access to `CharacterHit` (so can determine which character is at point x, y)
  - Implemented better caret-blinking (visible when user moves the caret, visible immediately upon showing the area)
  - Added pseudo classes for line which has the caret
  - Added pseudo classes for handling 
- Clone related
  - Allow multiple views of the same `EditableStyledDocument`
- Other
  - Converted all default `EventHandlers` to `EventHandlerTemplates`
  - Made double-clicking a word select only the word (not the leading/trailing whitespace char)
  - Removed unused and useless font property
  - Implemented `onSelectionDrop` to override default handling when moving a selection elsewhere in the area.
  - Fixed undo bug (styles from deleted text would cause a mismatch between the undone document and the one to redo)

With the numerous issues still open, what should be included in the `0.7` release and what should be postponed until `0.7.1`?

**Edit:** Include WellBehavedFX experimental code
**Edit 3/16/16:** Exclude #252 as it's no longer relevant. Added checkboxes.

Of the issues we have, I think the following should be included in the `0.7` release:
- [x] #233 because it will cause problems with cloning and overall stability.
    - This issue was closed due to being outside the scope of RTFX
- [x] Use the `WellBehavedFX` experimental InputMap API for setting default UI handlers that can be overridden (see the last few comments of #257 after it was closed) 

**Edit 9/7/16:** Removed ideas for `0.7.1` release

JordanMartinez:
Also note that if one searches the `0.7` milestone, the only issue that arises is #210, which seems to have been fixed now.

JordanMartinez:
I'm going to add a note here to keep track of what needs to be updated in the ReadMe when the new release is made:
- [ ] Rename deprecated `InlineStyleTextArea` to `StyledTextArea` and update related content
- [ ] Add a link to the RichtTextFX CSS Guide (optional?)

firm1:
Hello,
[...]

PR #292, opened 2016-04-19
Add a demo that shows how to override the default behavior and some…

JordanMartinez:
… quirks of the new approach:

- Shows how to override the default behavior of a node by installing an `InputMap`
- Shows the bug (or feature?) that can arise when one adds a handler to the node's `on[EventType]Property`.

### DIFF
--- a/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/OverrideBehaviorDemo.java
+++ b/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/OverrideBehaviorDemo.java
@@ -1,0 +1,98 @@
+package org.fxmisc.richtext.demo;
+
+import static javafx.scene.input.KeyCode.*;
+import static javafx.scene.input.KeyCombination.*;
+import static org.fxmisc.wellbehaved.event.EventPattern.*;
+
+import javafx.application.Application;
+import javafx.event.Event;
+import javafx.event.EventHandler;
+import javafx.geometry.Insets;
+import javafx.scene.Scene;
+import javafx.scene.control.CheckBox;
+import javafx.scene.control.Label;
+import javafx.scene.input.KeyEvent;
+import javafx.scene.input.MouseEvent;
+import javafx.scene.layout.VBox;
+import javafx.stage.Stage;
+
+import org.fxmisc.richtext.InlineCssTextArea;
+import org.fxmisc.wellbehaved.event.InputMap;
+import org.fxmisc.wellbehaved.event.Nodes;
+
+/**
+ * This demo shows how to override the default behavior of a StyledTextArea via an InputMap.
+ * It also demonstrates the bugs that can arise if one adds a handler to one of the
+ * {@code on[EventType]Property}.
+ */
+public class OverrideBehaviorDemo extends Application {
+
+    public static void main(String[] args) {
+        launch(args);
+    }
+
+    @Override
+    public void start(Stage primaryStage) {
+        InlineCssTextArea area = new InlineCssTextArea();
+
+        InputMap<Event> preventSelectionOrRightArrowNavigation = InputMap.consume(
+                anyOf(
+                        // prevent selection via (CTRL + ) SHIFT + [LEFT, UP, DOWN]
+                        keyPressed(LEFT,    SHIFT_DOWN, SHORTCUT_ANY),
+                        keyPressed(KP_LEFT, SHIFT_DOWN, SHORTCUT_ANY),
+                        keyPressed(UP,    SHIFT_DOWN, SHORTCUT_ANY),
+                        keyPressed(KP_UP, SHIFT_DOWN, SHORTCUT_ANY),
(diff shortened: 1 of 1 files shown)


JordanMartinez:
I've updated this PR to be rebased on the master branch (and a few additional changes).

JordanMartinez:
@TomasMikula Is there an issue preventing you from merging this?

TomasMikula:
This demo is quite elaborate, demonstrating many things at once. I can see how 200 lines of code can be a little too much to process for someone who just wants to override a single key combination. Maybe sometimes less is more...

TomasMikula:
It shouldn't require internet connection to work.

TomasMikula:
Using a template only makes sense if it is going to be reused many times. For a one-off use, using InputMap directly is effectively the same, and probably easier.

JordanMartinez:
> This demo is quite elaborate, demonstrating many things at once. I can see how 200 lines of code can be a little too much to process for someone who just wants to override a single key combination. Maybe sometimes less is more...

Good point. However, if someone does indeed subclass a `StyledTextArea`, doesn't the demonstration show the basic template one would follow to do that?

> It shouldn't require internet connection to work.

I only included that to insure the link was valid. If it isn't and we don't catch it, then someone will file an issue and we could update it immediately. However, That is a good point.

> Using a template only makes sense if it is going to be reused many times. For a one-off use, using InputMap directly is effectively the same, and probably easier.

Perhaps this demo should be broken up into 2-3 demos? One could show how to write a template for a subclass of `StyledTextArea`. The second could show how to override one part of the handler via an InputMap. Third (or part of the second one) would show what happens if one also adds an EventHandler to the area's `on[EventType]Property`.

TomasMikula:
> I only included that to insure the link was valid. If it isn't and we don't catch it, then someone will file an issue and we could update it immediately. However, That is a good point.

We can just include the link in the README instead of the demo, and the same still applies (i.e. someone will file an issue).

A template can really only pay off when both
 1. it contains _many event handlers_ (such as the one in `StyledTextAreaBehavior`), and
 2. there are _many instances_ (e.g. if someone is implementing an IDE which will have many text areas).

When only a handful of event handlers is being defined/overridden, condition 1. does not hold. So I probably would not bother explaining templates in RichTextFX documentation.

Note that even when subclassing `StyledTextArea`, one can use `InputMap` directly; and even when not subclassing, one could use `InputMapTemplate`.

JordanMartinez:
> We can just include the link in the README instead of the demo, and the same still applies (i.e. someone will file an issue).

Works for me. BTW, are you keeping track of all the work to be done on the ReadMe? I think the things we've discussed span 3-4 topics, but I could only recall one (explaining that STA is the base of the other flavors) when I thought about it.

> So I probably would not bother explaining templates in RichTextFX documentation.

Ok, that makes sense. On another point, the API is the same for `InputMap` and `InputMapTemplate`, so most should be able to figure it out and, if not, it won't take long to explain.

JordanMartinez:
I've updated the PR.

richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/OverrideBehaviorDemo.java
```diff
@@ -0,0 +1,98 @@
...
+                "The added handler will insert a newline character at the caret's position when [Enter] is pressed.",
+                "If checked, the default behavior and added handler will both occur: ",
+                "\tthus, two newline characters should be inserted when user presses [Enter].",
+                "When unchecked, the handler will be removed."
+        ));
+        checkBoxExplanation.setWrapText(true);
+
+        EventHandler<KeyEvent> insertNewlineChar = e -> {
+            if (e.getCode().equals(ENTER)) {
+                area.insertText(area.getCaretPosition(), "\n");
```
TomasMikula: You would want to call `e.consume()` here, to prevent the event from bubbling up the scene graph. The peculiarity is that other handlers on the same Node will still be invoked. 
From [Handling JavaFX Events](http://docs.oracle.com/javafx/2/events/processing.htm):

> However, if the node that consumes the event has more than one filter or handler registered for the event, the peer filters or handlers are still executed.

TomasMikula:
> BTW, are you keeping track of all the work to be done on the ReadMe?

No, but I know someone is doing that [here](https://github.com/TomasMikula/RichTextFX/issues/259#issuecomment-213857924) ;)

JordanMartinez:
> No, but I know someone is doing that here ;)

Ah! I thought I opened an issue for that specifically, but I couldn't find it. So, I thought I had decided against doing that. :grinning:

richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/OverrideBehaviorDemo.java
JordanMartinez: Good point. It's fixed now.

TomasMikula:
:+1:

PR #296, opened 2016-04-23
Finish migration to stable release (WellBehavedFX 0.3) of new InputMa…

JordanMartinez:
…p behavior-handling approach


TomasMikula:
Thanks!

TomasMikula:
Looking at it now, `StyledTextAreaBehavior#dispose()` is not necessary any more. It was meant to be called when changing a skin. But now the same behavior is there for the whole lifetime of the text area, so there is no need to dispose it. It will go away the area goes.

JordanMartinez:
Ok. I'll add a new PR for that.

PR #304, opened 2016-05-10
Allow modification of area before embedding it into a VirtualizedScrollPane

JordanMartinez:
AreaFactory did three things:
- convenience methods that embed an area into a VirtualizedScrollPane (basically, gave it scroll bars)
- cloned an area
- due to the above two usages, also provide each area's constructor so that all necessary methods are found in the same class

Unfortunately, the structure prevented subclasses from cloning themselves. Thus, EditableStyledDocument was made public as an interface. Now that cloning is not restricted to AreaFactory (and the constructors were added), it seemed that AreaFactory's constructor methods could be removed and the embedding methods could be moved to their corresponding area's class.

However, this also had its problems.
1. the source code for each of the three flavors (which is relatively small) doubled in size due to the above methods being moved to their location. StyledTextArea, which is already large as it is, would be filled with even more lines.
2. using the same name for all areas (embeddedInstance) led to naming conflicts between StyleClassedTextArea and CodeArea
2. moving the `clone[AreaType]` to each area pollutes subclasses with unnecessary API.

So, the methods will stay where they are but Consumers will allow editing of the constructed area before it is embedded into a VirtualizedScrollPane. Note: sometimes this is helpful (see JavaKeywords demo as an example. Other times it's just easier to write it the normal way: construct an area, modify it, then wrap it in a VirtualizedScrollPane

### DIFF
--- a/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/JavaKeywords.java
+++ b/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/JavaKeywords.java
@@ -11,6 +11,7 @@ import javafx.scene.layout.StackPane;
 import javafx.stage.Stage;
 
 import org.fxmisc.flowless.VirtualizedScrollPane;
+import org.fxmisc.richtext.AreaFactory;
 import org.fxmisc.richtext.CodeArea;
 import org.fxmisc.richtext.LineNumberFactory;
 import org.fxmisc.richtext.StyleSpans;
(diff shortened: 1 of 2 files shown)


TomasMikula:
I don't know, did JavaKeywords really improve? Maybe we should just remove AreaFactory without replacement...

JordanMartinez:
Not really. I just updated it to use the new API because it could use it.

> Maybe we should just remove AreaFactory without replacement...

Having gone through this PR and seen once again the issues we have, I think that would be good. It's just easier to wrap an area in a VSP and go from there than anything else...

PR #309, opened 2016-05-13
Simple esd

JordanMartinez:


TomasMikula:
Thanks! We probably shouldn't add any more public methods than those declared in `EditableStyledDocument`. That means we should remove the `replaceText` method, which would require to rewrite the tests that use it.

JordanMartinez:
How about now?

TomasMikula:
👍

PR #313, opened 2016-05-17
Moved model classes in model package

afester:
This commit moves the model related classes into a separate "model" package. Meanwhile, there are plenty of classes in the "org.fxmisc.richtext" package, both UI related (with a strong dependency on JavaFX Node classes to create the scene nodes) and model related. Separating out the model classes would definitely help to understand the overall structure of the editor, specifically when getting started.
The drawback might be that some classes and methods had to be made public rather than package scoped, but I think this is reasonable.
Any comments are welcome ...


TomasMikula:
The change makes sense to me. I think it is a much better separation than what we used to have before: the main package and the `skin` sub-package, where in the main package we had a mixture of model and some parts of the view, and `skin` contained the rest of the view.

Can you also move model-related test classes to the `model` subpackage?

We should try to avoid circular dependencies between packages. Can we make `model` independent of the main package (i.e. not import anything from there).

You will also need to resolve the merge conflict.

afester:
Hi Tomas,
I have updated this branch to the head of master. I had already solved the merge conflict, and made sure that there are no dependencies from classes in the model package to other RichTextFX classes. The model package is now self-contained. I also moved the model tests to the `org.fxmisc.richtext.model` package. Please let me know if anything additional is required ...

richtextfx/src/main/java/org/fxmisc/richtext/model/SuperCodec.java
```diff
@@ -0,0 +1,51 @@
+package org.fxmisc.richtext.model;
+
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+public interface SuperCodec<S, T extends S> extends Codec<T> {
```
TomasMikula: Why did this need to become public?

TomasMikula:
Other than the `SuperCodec` made public, looks good to me 👍

richtextfx/src/main/java/org/fxmisc/richtext/model/SuperCodec.java
afester: Its used in `StyleClassedTextArea`: 

    SuperCodec.upCast(SuperCodec.collectionListCodec(Codec.STRING_CODEC)),
    SuperCodec.upCast(SuperCodec.collectionListCodec(Codec.STRING_CODEC))

TomasMikula:
👍

PR #314, opened 2016-05-17
Custom object

afester:
This commit is the basis for adding custom objects such as images to the rich text editor (based on the earlier modelPackage pull request). It also extends the rich text demo to allow the insertion of an image.
This is still work in progress, but I would at least like to get some feedback on the used approach. I still would like to make the custom object support more flexible, by allowing to register factories for specific object types so that the Nodes can be created individually based on the custom object type.
Any comments are welcome ...


TomasMikula:
Just a general remark: I am a big fan of type safety, so I would like to explore some generic approach that avoids runtime `instanceof` checks, or rules out the possibility (at compile time) that we get an object type that we don't know how to render.

PR #315, opened 2016-05-17
Load save

afester:
This commit adds a simple persistence feature (load/save) to the rich text demo. It is based on the earlier "customObject" pull request and supports Images as custom objects.


TomasMikula:
You should be able to re-use the provided `Codec`s (used for clipboard functionality) to write/read a binary representation of a document. That will save you from having to parse CSS, or XML.

afester:
Closing this pull request since load/save demo based on the Codecs is now included in #356

PR #324, opened 2016-06-10
Use the -rtfx prefix for RichTextFX specific underline properties, us…

afester:
…e the underline width as the toggle to enable/disable the underline

### DIFF
--- a/richtextfx-demos/src/main/resources/org/fxmisc/richtext/demo/spellchecking.css
+++ b/richtextfx-demos/src/main/resources/org/fxmisc/richtext/demo/spellchecking.css
@@ -1,6 +1,6 @@
 .underlined {
-    -fx-underline-color: red;
-    -fx-underline-dash-array: 2 2;
-    -fx-underline-width: 1;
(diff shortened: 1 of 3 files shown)


TomasMikula:
Cool, thanks!

JordanMartinez:
I've updated the CSS guide to use the updated prefix.

TomasMikula:
Great, thanks @JordanMartinez !

JordanMartinez:
No problem!

Quick question @TomasMikula `TextExt` also has the CSS property `-fx-background-fill`. Could you clarify: is this used internally (and thus shouldn't be included in the guide) or another part of the text that can be styled (and thus should be in the guide and also use the `-rtfx` prefix)?

TomasMikula:
Good catch! It _is_ meant to be used by users. So it should be renamed to `-rtfx` and included in the guide. At that occasion, we might as well rename it to `-rtfx-background-color`, since the only reason it was called `-fill` instead of `-color` was [JDK-8133685](https://bugs.openjdk.java.net/browse/JDK-8133685).

PR #325, opened 2016-06-10
Update another RichTextFX-specific CSS property to use the -rtfx pref…

JordanMartinez:
…ix; "-fill" reverted back to "-color"

Note: "-fill" was initially used  instead of "-color" because of JDK-8133685. Now that a different prefix is used, this is no longer necessary


TomasMikula:
There are more places in that file that need renaming from "fill" to "color".

JordanMartinez:
Good catch. How's that?

TomasMikula:
👍

PR #331, opened 2016-06-21
Add variants of editing/navagiting methods that take relative index argument (paragraph index, column index)

JordanMartinez:


richtextfx/src/main/java/org/fxmisc/richtext/model/NavigationActions.java
```diff
@@ -58,6 +83,43 @@ else if(pos >= sel.getEnd())
...
+                if(pos <= sel.getStart())
+                    anchor = sel.getEnd();
+                else if(pos >= sel.getEnd())
+                    anchor = sel.getStart();
+                else
+                    anchor = getAnchor();
+                selectRange(anchor, pos);
+                break;
+        }
+    }
```
TomasMikula: Seems a little repetitive wrt. the method above.

JordanMartinez:
I chose `StyledDocument` as the place to implement the base code for `getAbsoluteCharPosition` and not `TwoDimensional` because then that method would "pollute" its implementors with a confusing method name and parameters since it doesn't use `int paragraphIndex` and `int columnIndex` as its parameters but `int major` and `int minor`.

richtextfx/src/main/java/org/fxmisc/richtext/model/NavigationActions.java
JordanMartinez: Ah... I'll fix that.

TomasMikula:
Looks, good, just not sure if I want to use "Char" in the method names. I try to be consistent in using the term _position_ when pointing to a place between characters, and _character index_ when pointing to a character.

```
 ┌ character 0
 | ┌ character 1
 | |   ┌ character 3
 | |   |
 v v   v

|_|_|_|_|

^ ^     ^
| |     |
| |     └ position 4
| └ position 1
└ position 0
```

JordanMartinez:
I agree with keeping it consistent.

At the same time, I don't know if this distinction between _position_ and _character index_ is even stated in the javadoc or is as clear as the above picture. 
Should it be in the javadoc? Or somewhere else in the project pages (ReadMe, Wiki, etc.)?

JordanMartinez:
Well, since so many methods point to `getAbsolutePosition()`, whose javadoc now includes your picture, I think that takes care of the previous issue I raised.

richtextfx/src/main/java/org/fxmisc/richtext/model/TextEditingArea.java
```diff
@@ -197,4 +265,39 @@ default void replace(IndexRange range, StyledDocument<PS, S> replacement) {
      */
     @Deprecated
     public void positionCaret(int pos);
+
+    /**
+     * Returns the absolute position (e.g. the spot in-between characters) to the left of the given column in the given paragraph.
```
TomasMikula: You mean "i.e." instead of "e.g." ;)

richtextfx/src/main/java/org/fxmisc/richtext/model/TextEditingArea.java
```diff
@@ -197,4 +265,39 @@ default void replace(IndexRange range, StyledDocument<PS, S> replacement) {
...
+     *
+     * <p>For example, given a text with only one line {@code "text"} and the characterIndex value of {@code 1}, "position 1" would be returned:</p>
```
TomasMikula: `columnIndex` instead of `characterIndex`

richtextfx/src/main/java/org/fxmisc/richtext/model/TextEditingArea.java
```diff
@@ -161,11 +198,42 @@
...
+    /**
      * Replaces a range of characters with the given rich-text document.
      */
     void replace(int start, int end, StyledDocument<PS, S> replacement);
 
     /**
+     * Replaces a range of characters with the given rich-text document.
+     *
+     * <p><b>Warning:</b> uses {@link #getAbsolutePosition(int, int)}. Be aware of possible bugs that
+     * can arise by using this method.</p>
```
TomasMikula: This sounds as if the method is buggy.

JordanMartinez: Yeah.....
I wasn't sure how to basically say: "Hey, this method might do some quirky stuff because _you_ gave it an index that was off by one. So, keep an eye out!"

Do you have any suggestions for how to rephrase that to avoid that unwanted connotation?

richtextfx/src/main/java/org/fxmisc/richtext/model/TextEditingArea.java
JordanMartinez: I used "i.e." a lot until I saw you use "e.g." a lot. So, I thought they were the same and just used that whenever it related to you or this project. :smiley: Having looked it up, I see the difference now.

TomasMikula:
👍

PR #349, opened 2016-08-23
Get exact LinesCount!

Mulham-Raee:
- new method **getParagraphLinesCount**, returns an integer of linesCount in a specific Paragraph using its index as parameter.  
- **getCaretBoundsOnScreen** method in class StyledTextArea is now public, used to determine if the caret is outside the screen or not.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextArea.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextArea.java
@@ -1229,7 +1229,11 @@ public class StyledTextArea<PS, S> extends Region
         });
     }
 
-    private Optional<Bounds> getCaretBoundsOnScreen() {
+    public int getParagraphLinesCount(int index){
+        return virtualFlow.getCell(index).getNode().getLineCount();
+    }
+
+    public Optional<Bounds> getCaretBoundsOnScreen() {
         return virtualFlow.getCellIfVisible(getCurrentParagraph())
                 .map(c -> c.getNode().getCaretBoundsOnScreen());
     }


JordanMartinez:
What exactly is "linesCount" ? Is that the number of Text objects stored a TextFlow's list of children? I'm not familiar with that API.

Mulham-Raee:
@JordanMartinez a paragraph represents a Text (or TextField) object in a TextFlow (CodeArea), but still a TextField can store more than one line that is not necessarily separated by `\n` in case of WrapText.
Using this new method you can get the linesCount for each Paragraph (Text object) and the sum would be linesCount for the whole TextFlow (CodeArea).

JordanMartinez:
Oo! I can see why that would be very helpful.

Putting it into my own words... If you have a really long line of text, and the wrap text feature is on, and the StyledTextArea (or one of its flavors) is not wide enough to display the entire long line, then this feature would tell you how many lines it takes to display that entire long line.

> Using this new method you can get the linesCount for each Paragraph (Text object) and **the sum would be linesCount for the whole TextFlow (CodeArea)** **_(emphasis mine)_**.

But wouldn't this only work for the content that is actually displayed? I may be misunderstanding you, but since most of the document isn't displayed, you still wouldn't be able to know how many lines it takes to display the entire document. Each line itself is a TextFlow; CodeArea is a VirtualFlow with many TextFlow objects (cells) as needed to display the content that fits within the viewport.

Mulham-Raee:
Yes that is exactly what I meant.
I am not sure if I can get the linesCount of the text that is not displayed, but still it is very helpful to track any new lines inserted in StyledTextArea when the wrapText feature is on.
That was the case when I needed this feature, as I wanted the CodeArea Height to fit its content instead of scrolling to see the rest.

JordanMartinez:
@TomasMikula any comments here? This PR has been open for a while now...

DaveJarvis:
Consider also a `getTotalLineCount()` that iterates over the paragraphs to run the total sum of lines for the text area.

sherzfel:
This branch returns either 1 is wrapText is false or if wrapText is true it returns the number of signs. Am I doing something wrong or has this branch an error?

JordanMartinez:
Both features of this PR have been implemented in other PRs. The only one that hasn't is the one that @DaveJarvis mentioned: `area.getLineCount()`, which would return the sum of all paragraph's `lineCounts`. @DaveJarvis if you still want that feature, please open a new issue. It won't take long to resolve, but there may be a few other factors we need to take into consideration.

PR #356, opened 2016-09-09
Custom object support

afester:
This is an enhanced proposal to support custom objects which can be rendered by user specific nodes.

In the model layer, all occurrences of `StyledText`have been replaced by the new interface `Segment`. `StyledText`is now just one flavor of segment types. `CustomObject`is an abstract class which implements the `Segment`interface and which provides a base class for user defined objects. Out-of-the-box, RichTextFX would support images to be inserted into the document, but the user can define any other custom objects by registering a factory method for the JavaFX node and a decode method for deserializing the object through a  Codec.  See `CustomObjectDemo`for an example how to register and use custom objects.

The code already supports copy&paste (if the selection contains a custom object, it is serialized when copying). Selecting one single custom object by clicking on it is not supported yet.

Any feedback/comments are welcome - especially registering the decode method is very unclean yet, see the SegmentFactory class. Essentially, decoding must be delegated to the custom object implementation since only the custom object knows how to recreate itself from a DataInputStream, but the method in ReadOnlyStyledDocument which does the decoding is static which makes it difficult to access a delegate. I tried to make those methods non-static, but this looked like a huge bunch of refactoring is necessary then.


richtextfx/src/main/java/org/fxmisc/richtext/model/CustomObject.java
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * This is the base class for custom objects in the model layer.
+ * Its String representation is always one character long and contains
+ * the "object replacement character" (\ufffc). 
+ */
+public abstract class CustomObject<S> implements Segment<S> {
+
+    private S style;
+    private SegmentType typeId;
```
JordanMartinez: A minor detail, but this should probably be renamed to `type` now, not `typeId` as it was when you were using an `int` rather than your current enum/interface approach

richtextfx/src/main/java/org/fxmisc/richtext/model/Paragraph.java
```diff
@@ -90,23 +103,33 @@ public String substring(int from) {
...
         } else {
-            List<StyledText<S>> segs = new ArrayList<>(segments.size() + p.segments.size());
+            List<Segment<S>> segs = new ArrayList<>(segments.size() + p.segments.size());
             segs.addAll(segments);
             segs.addAll(p.segments);
             return new Paragraph<>(paragraphStyle, segs);
         }
     }
 
+    private boolean canJoin(Segment<S> left, Segment<S> right) {
```
JordanMartinez: This doesn't seem to be extensible... What happens if we or developers in their code include additional features that can be joined together? Wouldn't it make more sense to have some sort of segment-joining strategy?

For example:
- Strategy 1: inherit the left side's style
 - 1 plain Text + 1 plain Text = 1 plain Text
 - 1 BulletPointText + 1 plain Text = 1 BulletPointText
 - 1 plain Text + 1 BulletPointText = 1 plain Text
- Strategy 2: inherit the right side's style
 - 1 Empty Text + 1 Rich Text = 1 Rich Text
- Strategy 3: keep styles/objects distinct
 - 1 bolded text + 1 italicized text = 1 bolded text + 1 italicized text 
 - 1 Text + 1 Image = 1 Text + 1 Image
 - 1 Image + 1 Text = 1 Image + 1 Text
 - 1 Table + 1 Text = 1 Table + 1 Text
 - 1 Table + 1 Table = 1 Table + 1 Table
- Strategy 4: keep as two distinct objects, but change the style of latter
 - 1 Table + 1 BulletPointText = 1 Table + 1 Text

Having written out these examples, I'm thinking that a bullet-point style should be implemented as a paragraph style, not a text style.

TomasMikula:
Thanks for the work you have done on this so far!

My main question (for everyone) is: Can we find a way to turn [this `RuntimeException`](https://github.com/afester/RichTextFX/blob/7d0d23e081c165a617b7404b88f5cfbf45b0461b/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextArea.java#L1209) into a **compile time** error? For any application, the types of custom objects are known in advance (at compile time). It's not like they are being defined dynamically or loaded dynamically into the JVM at runtime. So in principle, it should be possible. I know Java is not great at these kinds of abstractions, but can we give it some thought?

JordanMartinez:
> Can we find a way to turn this RuntimeException into a compile time error?

Couldn't we store the `NodeFactory` getter inside the object implementing the `Segment` interface? For example:
````java
public interface Segment<S> {
    public <T extends Segment<S>> NodeFactory<S, T> getFactory();
}

@FunctionalInterface
public interface NodeFactory<S, T extends Segment<S>> {
    public Node apply(T seg);
}

// implementations
class StyledText<S> extends Segment<S> {
    public NodeFactory<S, StyledText<S>> getFactory() {
        return DefaultNodeFactories.getStyledTextFactory();
    }
}

public class DefaultNodeFactories<S> {
    private static final NodeFactory<S, StyledText<S>> STYLED_TEXT_FACTORY = 
            seg -> { /* view construction code here... */ };

    public static final StyledTextFactory<S, StyledText<S>> getStyledTextFactory() { 
           return STYLED_TEXT_FACTORY; 
    }
   
   // other default factories here (e.g. table, shapes, etc.)
}
````

**Comment by TomasMikula** [comment]
Something like that would probably work, but isn't a very good separation
between model an view.

**Comment by JordanMartinez** [comment]
> it isn't a very good separation between model an view.

True... It's almost like we'll need to add another generic to StyledTextArea with a class that does know how to display everything. For example:
````java
public class StyledTextArea<PS, S, NF extends NodeFactory> {
    private final NF nodeFactory;
}
public interface NodeFactory {
    public Node apply(Segment<?> seg);
}
````
However, won't that still lead to `instanceof` checks?
```java
public Node apply(Segement<?> seg) {
    if (seg instaceof StyledText<?>) {
        // styled text factory applied here...
    else if (seg instaceof Table) {
        // table factory here
    } else {
        // some other factory here
    }
}
````

**Comment by TomasMikula** [comment]
What if we abstract over the type of segments in a type parameter:

```java
/**
 * Segment operations on type Seg.
 * Notice that each method takes the segment as an argument.
 */
public interface SegmentOps<Seg> {
    int length(Seg segment);
    char charAt(Seg segment, int index);
    /* ... */
}

public class StyledTextArea0<PS, Seg, S> {
    public StyledTextArea0(
            PS initialParagraphStyle, BiConsumer<TextFlow, PS> applyParagraphStyle, // as before
            S initialTextStyle, // as before
            SegmentOps<Seg> segmentOps,
            Function<Seg, Node> nodeFactory,
            BiFunction<Seg, S, Seg> applyStyle // Given a segment, update its style. The result is a new segment.
    ) {
        /* ... */
    }
}
```

Note that any type `Foo` can now represent a segment, without having to implement any interface, as long as we provide an implementation of segment operations for that type, i.e. an instance of `SegmentOps<Foo>`. This is similar to how you can create a `TreeSet<E>` for any type `E` whatsoever, given that you [provide a `Comparator<E>`](https://docs.oracle.com/javase/8/docs/api/java/util/TreeSet.html#TreeSet-java.util.Comparator-).

Our original `StyledTextArea` is then

```java
public class StyledTextArea<PS, S> extends StyledTextArea0<PS, StyledText<S>, S> {}
```

If one wanted to embed images in addition to `StyledText`, they could define their own

```java
public class MyArea<PS, S> extends StyledTextArea0<PS, Either<StyledText<S>, LinkedImage<S>>, S> {}
```

Extending is not even really necessary, it is just to provide a shorthand for the rather long type `StyledTextArea0<PS, Either<StyledText<S>, LinkedImage<S>>, S>`.

JordanMartinez:
> What if we abstract over the type of segments in a type parameter
That seems like it would work. Additionally, it looks like it would address the `canJoin` strategy issue I raised earlier since that would be user-defined.

afester:
Hi Tomas and Jordan,

thanks for your feedback!

> Something like that would probably work, but isn't a very good separation
between model an view.
Right, that is what I also observed while I tried to implement the proposal from @JordanMartinez ... I will check if the additional type parameter is feasible (even though I dont feel completely comfortable with adding yet another type parameter ...) 
Besides the Node factory, I am more concerned about the factory which creates a specific segment type from its serialized representation (decode). I dont think that the SegmentFactory I am currently using is a very clean solution - do you have any idea on that one`?

TomasMikula:
> even though I dont feel completely comfortable with adding yet another type parameter

I feel the pain from the user point of view. As an implementor, I'm completely comfortable with having lots of type parameters and a very generic implementation. I even think abstracting something away as a type parameter reduces the potential for bugs in the implementation, since you don't have access to the specifics of the concrete type.

From the user point of view, it is a major pain if you have to keep specifying 3 type arguments, especially when they are the same 3 concrete types throughout your whole application. I believe this issue could be resolved satisfactorily by using a _type alias_. Unfortunately, [Java doesn't have type aliases](http://stackoverflow.com/questions/5604390/how-do-i-create-some-variable-type-alias-in-java). Type aliases exist e.g. in

C++:

```cpp
using Style = Collection<String>
using MyArea = StyledTextArea0<Style, StyledText<Style>, Style>
```

Scala:

```scala
type Style = Collection[String]
type MyArea = StyledTextArea0[Style, StyledText[Style], Style]
```

and other languages. What we can do in Java is

```java
class MyArea extends StyledTextArea0<Collection<String>, StyledText<Collection<String>>, Collection<String>> {
    /* define constructors */
}
```

but then `StyledTextArea0<Collection<String>, StyledText<Collection<String>>, Collection<String>>` cannot be used where `MyArea` is required, because they are not the _same_ type.

I'm somewhat reluctant to compromise what otherwise seems like a reasonable design due to a deficiency in the language.

> I am more concerned about the factory which creates a specific segment type from its serialized representation (decode).

`decode` in `StyledText` would produce `StyledText`, `decode` in `LinkedImage` would produce `LinkedImage`. Then if we were to decode `Either<StyledText<S>, LinkedImage<S>>`, we need a `Codec<Either<StyledText<S>, LinkedImage<S>>>`. Such codec can be defined easily, if we already have `Codec<StyledText<S>>` and `Codec<LinkedImage<S>>`. (In the same spirit as [`Codec<Optional<T>>` is defined based on `Codec<T>`](https://github.com/TomasMikula/RichTextFX/blob/290a4e72b416becd20404bdd555de53e64a7e0a4/richtextfx/src/main/java/org/fxmisc/richtext/model/Codec.java#L72).) Does this answer your concern?

JordanMartinez:
> I'm somewhat reluctant to compromise what otherwise seems like a reasonable design due to a deficiency in the language.

I'd stick with the good design since it will lead to less bugs and only Java-language users are affected. Java will hopefully add that feature in a later release, but until then one can get work around this annoyance through ways that work for them. For example, typing in a placeholder text (e.g. `Style`) that they then later programmatically replace in the IDE's editor with the correct concrete class (e.g `Collection<String>`), or defining a keyboard shortcut that will paste the correct code in when needed.

TomasMikula:
Those IDE tricks help with writing those programs, but not with reading. Code is much more often read than written. Otherwise, I agree with the rest.

JordanMartinez:
> Those IDE tricks help with writing those programs, but not with reading. Code is much more often read than written. Otherwise, I agree with the rest.

Good point. :disappointed:

JordanMartinez:
How long would it take for a feature like this to make it into Java if a JSR was started for it (I'm not sure if there already is one / has been one)?

kuc:
>How long would it take for a feature like this to make it into Java if a JSR was started for it (I'm not sure if there already is one / has been one)?

Probably no earlier than Java 11.

JordanMartinez:
> Probably no earlier than Java 11.

Lame....

Let's put it this way. There have been four approaches so far: 
- afester's approach (which has a runtime exception), 
- my "include a NodeFactory in the implemented Segment" approach (which has a design issue as the model not separated from view), 
- my second approach (which leads to a runtime exception eventually), or 
- Tomas' approach (which leads to a reading issue). 

It seems like we've exhausted all the approaches we could take to this problem, correct? So, which issue is the least problematic holistically?
- My first approach is a design issue, so it should be excluded immediately to prevent a greater complexity of bugs arising from that design issue
- Afester's approach or my second approach's flaws can be minimized via user's unit testing
- Tomas' approach's flaws can only be fixed when Java does(sometime in the next decade), or if users decide to take it upon themselves to understand the code, or if we have good documentation that happens to not change frequently.

afester:
> I feel the pain from the user point of view. As an implementor, I'm completely comfortable with having lots of type parameters and a very generic implementation. I even think abstracting something away as a type parameter reduces the potential for bugs in the implementation, since you don't have access to the specifics of the concrete type.

Definitely, using appropriate language features to make the implementation more robust is the right approach. However, its also important to consider the API design - complex APIs prevent people from using the component. When evaluating a component, the first things people look at are probably the code examples and the JavaDoc - for the time being, the component itself is a black box for them.

Anyway, @JordanMartinez thanks for summarizing the various approaches - I will see if I can come up with a proposal for Tomas' generics approach especially to see how the API "feels" with that approach.

afester:
@TomasMikula 

> Note that any type Foo can now represent a segment, without having to implement any interface, as long as we provide an implementation of segment operations for that type

Would Foo not be required to at least extend from some abstract base class? Otherwise, where would you store the style information, that is (today in StyledText)

    private final S style;

    public S getStyle() {
        return style;
     }

JordanMartinez:
@afester 

> Would Foo not be required to at least extend from some abstract base class? Otherwise, where would you store the style information, that is (today in StyledText)
>
> private final S style;
>
> public S getStyle() {
>    return style;
> }

I believe you're right. As it stands, we can't store that info in `SegOp` due to its generic type:
````java
public class Segment {}
public interface SegOp<Segment> {
    public int length(Segment segment);
    public S getStyle(Segment segment); // compilation error
}
````
So we'd need another interface like `Styleable` (or some better-named interface to prevent confusion with paragraph styles) for that. For example:
````java
public interface Styleable<S> {
    public S getStyle();
}
public class Foo implements Styleable<String> {}
Segment<Foo> seg = //
assert seg.getStyle().getClass() == String.getClass();
````

**Comment by TomasMikula** [comment]
> complex APIs prevent people from using the component. When evaluating a component, the first things people look at are probably the code examples and the JavaDoc

I believe we could preserve current `StyledTextArea` API as is, even maintaining source compatibility. We will just make it extend a more generic class (which in my example I named `StyledTextArea0`).

> Otherwise, where would you store the style information

Good catch! In my sketch, I intended the `applyStyle` function passed to the constructor to update style of a segment:

```java
BiFunction<Seg, S, Seg> applyStyle
```

But I forgot about extracting style back (`Function<Seg, S> getStyle`). We could either make `SegmentOps` style-aware, i.e.

```java
public interface SegmentOps<Seg, Style> {
    Style getStyle(Seg segment);

    // and then `setStyle` can go here as well
    Seg setStyle(Set segment, Style style);
}
```

Or add the setter to the parameter list of the `StyledTextArea0` constructor.

Btw, a pair of getter and setter functions is also called a [`Lens`](http://stackoverflow.com/questions/8307370/functional-lenses#8312477), so we could package it up in an interface

```java
public interface Lens<T, U> {
    U get(T t); // getStyle
    T set(T t, U u); // setStyle

    // factory method to create a Lens given two lambdas
    static <A, B> Lens<A, B> create(Function<A, B> getter, BiFunction<A, B, A> setter) {
        return new Lens<A, B> {
            // implementation trivial
        }
    }
}
```

so that it is a single constructor parameter.

Also, I think in the most generic version of `StyledTextArea` (say, `StyledTextArea00` 😄), we could even do without any notion of a segment style at all. Then `StyledTextArea0` introduces the notion of segment style and since it has the `Lens` to get/set styles of segments, it can add segment style-related methods to its API. But this is just a thought for a potential next iteration :)

JordanMartinez:
> But I forgot about extracting style back (Function<Seg, S> getStyle). We could ... make SegmentOps style-aware

I didn't list your suggestion in my previous comment because I thought it would be easier for users if they did not need to keep another generic type in mind when doing segment operations. Thinking about it now, that's probably not true. 

Would there be any advantage to adding it to the `StyledTextArea0` constructor parameter list as opposed to the `SegOps` interface? The latter makes more sense in my mind because any style-related operations (e.g. merge) could also be contained there, though that could be stored in another interface `StyleOps` that is later stored in a `SegOps` implementation.

> StyledTextArea00 :smile:

:-) I feel like that could get confusing after a while...  Could we just call `StyledTextArea00` `RichAreaBase` / `RichTextAreaBase` instead (since it would no longer have 'styles' associated with it and there can be more than just text displayed) and call `StyledTextArea0` `StyledRichAreaBase` (because that's where styles would be introduced), and keep `StyledTextArea` as is?

TomasMikula:
And then add a `GenericRichAreaBase`, which will be just a subclass of `AbstractGenericRichAreaBase` 😄 

Just kidding, naming is hard. `StyledTextArea0` was just a working name :)

> Would there be any advantage to adding it to the `StyledTextArea0` constructor parameter list as opposed to the `SegOps` interface? The latter makes more sense in my mind

Yeah, makes more sense to me as well. If we then decide to generalize to `StyledTextArea00`/`RichTextAreaBase`, we will need a `SegOps0` (:wink:) that doesn't talk about styles.

JordanMartinez:
> And then add a GenericRichAreaBase, which will be just a subclass of AbstractGenericRichAreaBase :smile:

:laughing: Lol! 

> If we then decide to generalize to StyledTextArea00/RichTextAreaBase, we will need a SegOps0 (:wink:) that doesn't talk about styles.

Or we could have a base interface with no style, `SegOps` (my idea for your `SegOps0`), and possibly an extension of that interface that adds style, `StyledSegOps` (my idea for our current `SegOps`).

JordanMartinez:
@afester Can you give us an update? Or have you been working on other things recently?

afester:
I started with the Generics approach (you can have a look at it at https://github.com/afester/RichTextFX/tree/genericSegment), but I found that the separate implementation of the operations unnecessarily bloats up the code, since the operations have to be looped through to many classes and in some cases also to static methods. If we still want to persue a Generics based approach, I would suggest to at least provide a base class for the generic segment, like <SEG extends Segment>.

Anyway, I would like to make another proposal which I just committed: Instead of adding an `enum` for each segment type, this approach uses the class name as the type id for the segment types. This is similar to what Java serialization does and allows to recreate the Segment objects through reflection. It avoids the mapping between type id and factory, and thus the RuntimeException. Naturally, a bunch of other errors might occur during instantiation of the Segments through reflection, but these are really severe runtime errors like invalid classpaths etc. The only place where the objects need to be recreated is the decode method of the Codec, and all those errors are thrown as IOException with the actual error as nested exception.

To create the JavaFX nodes, the segments now provide a `createNode()` method. The little drawback is that there is now a small dependency from the model to the JavaFX Node, but I think that dependencies to the Java runtime are fine. The Node factories for the default segment types (currently StyledText and LinkedImage) are injected into the model package to avoid dependencies from the model layer up to the view layer.

The segments now also have a method `canJoin` which checks if a particular segment can be joined with another segment - this is more flexible than the earlier approach.

Finally, I had to rework the algorithm which restyles part of the document. The earlier approach split the corresponding text and re-created the StyledText objects for each style span, but this is not possible anymore since there might be other Segment types involved. I created some Unit tests for that and the new approach works fine (including the RichText demo when selecting and restyling a range which includes an image), but there might still be corner cases and/or better approaches.

Please let me know any feedback ...

JordanMartinez:
Thanks for looking into the generic approach!

> I found that the separate implementation of the operations unnecessarily bloats up the code, since the operations have to be looped through to many classes and in some cases also to static methods.

I went through your code in my IDE and I think I see your point about how the `SegOps` bloats the code. In my edits to your `genericApproach` branch (see my PR), it starts with `StyledDocument`: I added a `getSegOps` method to the interface so that `ReadOnlyStyledDocument` and `EditableStyledDocument` could return their related object for various parameters in methods. `ParagraphText` (and thus `ParagraphBox` which creates it) needs the `SegOps` object which an area's `EditableStyledDocument` can provide through the aforementioned interface's new method, so that it can extract the text and style from the `Segment`. Then there is the model class, `Paragraph`, that needs to know how to operate on its segments.
Finally, `StyledTextArea`'s most generic constructor is the only one that needs the object as a parameter so it can properly construct a `SimpleEditableStyledDocument`. If the ESD s already constructed, the rest of the constructors' parameter lists never change.

In my PR to your branch, I also added a `SegmentOps` object to the static methods of `ReadOnlyStyledDocument`. I'm not sure of all the implications of this, but it seems like it could cause a release-dependent issue depending on how one implements the code. What happens if the `SegmentOps` class is updated/modified later on? Will that screw up anything?

Part of the issue here is also readability, correct? For example, the repetitive calls of `segmentOps.[method]` looks weird as shown in a portion of the method `getStyleSpans`:
````java
        if(startSegIdx == endSegIdx) {
            SEG seg = segments.get(startSegIdx);
            builder.add(segmentOps.getStyle(seg), to - from);
        } else {
            SEG startSeg = segments.get(startSegIdx);
            builder.add(segmentOps.getStyle(startSeg), segmentOps.length(startSeg) - start.getMinor());

            for(int i = startSegIdx + 1; i < endSegIdx; ++i) {
                SEG seg = segments.get(i);
                builder.add(segmentOps.getStyle(seg), 
                            segmentOps.length(seg));
            }

            SEG endSeg = segments.get(endSegIdx);
            builder.add(segmentOps.getStyle(endSeg), end.getMinor());
        }
````
We could add method wrappers to make it easier to read, but I'm not sure how that would affect performance:
````java
class Paragraph {
    private final SegmentOps<Seg, S> segOps

    // method wrappers for "segOps.length(seg)"
    private int segLength(Seg segment) { return segOps.length(segment); }
}
````
Then the above method might look like this:
````java
        if(startSegIdx == endSegIdx) {
            SEG seg = segments.get(startSegIdx);
            builder.add(segStyle(seg), to - from);
        } else {
            SEG startSeg = segments.get(startSegIdx);
            builder.add(segStyle(startSeg), segLength(startSeg) - start.getMinor());

            for(int i = startSegIdx + 1; i < endSegIdx; ++i) {
                SEG seg = segments.get(i);
                builder.add(segStyle(seg), segLength(seg));
            }

            SEG endSeg = segments.get(endSegIdx);
            builder.add(segStyle(endSeg), end.getMinor());
        }
````

**Comment by JordanMartinez** [comment]
I've updated my PR to your generic approach (once you merge it) to now compile completely. The only demo that actually works (I used `SegOps` place holders on the rest of the demos) is the RichText demo.

**Comment by TomasMikula** [comment]
> but I found that the separate implementation of the operations unnecessarily bloats up the code, since the operations have to be looped through to many classes and in some cases also to static methods.

Can you point me to this problematic code?

> I would suggest to at least provide a base class for the generic segment, like `<SEG extends Segment>`.

That limits the flexibility somewhat. Without requiring inheritance from a base class (or implementing an interface), even classes that were not originally intended to represent a segment can be used as such.

> Anyway, I would like to make another proposal which I just committed: Instead of adding an enum for each segment type, this approach uses the class name as the type id for the segment types. This is similar to what Java serialization does and allows to recreate the Segment objects through reflection. 

I'm afraid we are asking for a lot of trouble down the road if we use reflection. How would your [`decode` method](https://github.com/afester/RichTextFX/blob/41cf7032a585f425fc51c9c8381e2197c5ceb68d/richtextfx/src/main/java/org/fxmisc/richtext/model/ReadOnlyStyledDocument.java#L176-L187) for `Segment`s work for a segment that has an additional type parameter, like

```java
class MySegment<T, S> implements Segment<S> {
    private T t;
    public void decode(DataInputStream is, Codec<S> styleCodec) throws IOException {
        // how to implement this???
        // We have no way to decode T (because we have no Codec<T>)
    }
}
```

Not to mention that all `Segment`s now need to know how to serialize/deserialize themselves. I don't feel comfortable requiring users to implement serialization for their segments even when they are not at all interested in saving the content of the text area.

Btw, you are addressing two concerns in this PR: custom objects and serialization. I think we can address them separately, which would make your work on this PR easier.

> To create the JavaFX nodes, the segments now provide a `createNode()` method.

My objection is the same as before, that this doesn't allow for separation of model from view.

JordanMartinez:
> Btw, you are addressing two concerns in this PR: custom objects and serialization. I think we can address them separately, which would make your work on this PR easier.

But can we implement custom objects without it also affecting the serialization? Or are you suggesting we export the 'serialization' part to be outside the area? For example, putting it into `EditableStyledDocument` and allowing the developer to determine whether or not this method exists.

Wouldn't the "custom object" feature also require some API for how/when to display the caret (e.g. tables)?

TomasMikula:
Currently, we say:

 - _"If you tell me how to serialize `PS` and `S`, then we can serialize `ReadOnlyStyledDocument<PS, S>`."_

This would just change to

 - _"If you tell me how to serialize `PS` and `Segment<S>`, then we can serialize `ReadOnlyStyledDocument<PS, Segment<S>, S>`."_

> Wouldn't the "custom object" feature also require some API for how/when to display the caret (e.g. tables)?

That's right, but let's not worry about that now. Also, this PR is about inline content, while a table would be more like a custom paragraph.

TomasMikula:
Where _"tell me how to serialize `X`"_ means _"give me `Codec<X>`"_.

afester:
> Not to mention that all Segments now need to know how to serialize/deserialize themselves. I don't feel comfortable requiring users to implement serialization for their segments even when they are not at all interested in saving the content of the text area.

It is also required for copy&paste. How should we handle this? I could be wrong, but I think that each Segment type **must** support serialization/deserialization. Otherwise you could not copy&paste an area containing a segment which does not support it.

> To create the JavaFX nodes, the segments now provide a createNode() method.
> > My objection is the same as before, that this doesn't allow for separation of model from view.

That is right, and I am a big supporter of that separation ;-) However I dont know yet how to handle this - currently, the ParagraphText constructor creates the JavaFX nodes like

```
    for(SEG segment: par.getSegments()) {
        TextExt t = new TextExt(segmentOps.getText(segment));
        ...
        getChildren().add(t);
```

Here we have the generic `segment` and the `SegmentOps` interface. A natural solution would be

```
    for(SEG segment: par.getSegments()) {
        Node t = segmentOps.createNode(segment);
        ...
        getChildren().add(t);
```
but since SegmentOps (and StyledTextOps) are part of the model layer, this would again lead to a dependency between model and view.

> That's right, but let's not worry about that now. Also, this PR is about inline content, while a table would be more like a custom paragraph. 

It could also be implemented as a segment type - imagine a Paragraph which contains an image, followed by a table, followed by some text. I already implemented a simple table segment type while working on the custom objects, to see if it is feasible at all...

TomasMikula:
> It is also required for copy&paste.

True. Still, I think serialization should be a separate concern. Segments, ReadOnlyStyledDocument, etc. theoretically make sense on their own, without being used in StyledTextArea.


In my first sketch above that introduced `SegmentOps`, I also introduced additional `StyledTextArea` constructor parameter for creating segment nodes, `Function<Seg, Node> nodeFactory`. This would need to be passed to `ParagraphText` constructor.

> It could also be implemented as a segment type - imagine a Paragraph which contains an image, followed by a table, followed by some text. I already implemented a simple table segment type while working on the custom objects, to see if it is feasible at all...

Fair enough, my comment missed the point. I think Jordan meant that a custom object might be focusable and be editable by itself, in which case, we need to have a way to transfer focus to that object.

JordanMartinez:
> I think Jordan meant that a custom object might be focusable and be editable by itself, in which case, we need to have a way to transfer focus to that object.

Yes, I was thinking of tables that have a couple of columns/rows...

JordanMartinez:
> Still, I think serialization should be a separate concern. Segments, ReadOnlyStyledDocument, etc. theoretically make sense on their own, without being used in StyledTextArea.

We mentioned this idea once before, but what if we added a Builder-like API for reconstructing a `StyledDocument`? Wouldn't that allow a user to design their own custom segments without needing to implement code that (de)serializes the object?

TomasMikula:
@JordanMartinez My comment was about the [encode/decode methods on `Segment`](https://github.com/afester/RichTextFX/blob/41cf7032a585f425fc51c9c8381e2197c5ceb68d/richtextfx/src/main/java/org/fxmisc/richtext/model/Segment.java#L31-L33) introduced in this PR (builder API seems unrelated). The generic approach, on the other hand, leaves serialization completely to Codecs.

JordanMartinez:
@TomasMikula Gotcha.

With that being said, I had a question about where the codecs should be. For example, it seems appropriate for [the document and paragraph codecs](https://github.com/TomasMikula/RichTextFX/blob/master/richtextfx/src/main/java/org/fxmisc/richtext/model/ReadOnlyStyledDocument.java#L97-L141) to stay inside `ReadOnlyStyledDocument`. Although they could be moved into a `Codecs` factory class, it would only add clarity in where to get them, nothing else. 
However, I'm not so sure about the [StyledText codec](https://github.com/TomasMikula/RichTextFX/blob/master/richtextfx/src/main/java/org/fxmisc/richtext/model/ReadOnlyStyledDocument.java#L143-L165). Shouldn't that be moved elsewhere now because it's no longer just `StyledText` objects that act as segments in the document? If so, putting it into the aforementioned `Codecs` factory class might make more sense. Additionally, we could put the codecs code of any new custom objects we later add to this project (e.g. tables) inside that class.

TomasMikula:
I would have codecs for existing external types, like `String`, `List`, `Optional`, `Either`, etc., as static methods on `Codec`, and package codecs for RichTextFX classes with the respective classes.

This would mean moving `StyledText` and `Paragraph` codecs to the respective classes.

JordanMartinez:
Sounds good

afester:
This is the current state of the Generics based approach, including the work done by @JordanMartinez. The unit tests are all succeeding, the demos compile (and should run - I have not tried them all). The RichText demo can load the `sample.rtfx` file with some images - inserting an image is not implemented yet. Also the latest comments by @TomasMikula regarding Codecs are not considered yet - there might be additional changes necessary. I have renamed `StyledTextArea` to `GenericStyledArea` and `StyledTextArea` is essentially a facade for a `GenericStyledArea` with a `StyledText<S>` segment type. With that, the demo's required only very few changes. This is still work in progress, but any feedback is welcome :-)

richtextfx/src/main/java/org/fxmisc/richtext/model/SegmentOps.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @param <SEG>
+ * @param <S>
+ */
+public interface SegmentOps<SEG, S> {
+    public int length(SEG seg);
+
+    public char charAt(SEG seg, int index);
+
+    public String getText(SEG seg);
```
JordanMartinez: Why don't we make `getText()` and `length()` default methods that return the single `/uffc` (or whatever it is) character and a length of 1? Then, only text-related custom objects need to override that, which is expected anyway.

richtextfx/src/main/java/org/fxmisc/richtext/model/SimpleEditableStyledDocument.java
```diff
@@ -22,14 +22,14 @@
 /**
  * Provides an implementation of {@link EditableStyledDocument}
  */
-public final class SimpleEditableStyledDocument<PS, S> implements EditableStyledDocument<PS, S> {
+public final class SimpleEditableStyledDocument<PS, SEG, S> implements EditableStyledDocument<PS, SEG, S> {
```
JordanMartinez: Since we're maintaining source compatibility with `StyledTextArea`, why not do the same for this one? Unfortunately, I'm not sure what such a name would be called.

JordanMartinez:
I'll keep looking at it and give more feedback, but thanks for your work @afester!

JordanMartinez:
See [my PR that addresses the immutability issue](https://github.com/afester/RichTextFX/pull/4)

richtextfx/src/main/java/org/fxmisc/richtext/model/ReadOnlyStyledDocument.java
```diff
@@ -67,143 +66,124 @@ public Summary reduce(Summary left, Summary right) {
...
-     * @param paragraphStyle The paragraph style to use for the paragraph which contains the segment.
-     *
-     * @return A ReadOnlyStyledDocument with the given segment.
-     */
-    public static <PS, S> ReadOnlyStyledDocument<PS, S> from(Segment<S> seg, PS paragraphStyle) {
-        List<Paragraph<PS, S>> res = new ArrayList<>(1);
-        Paragraph<PS, S> content = new Paragraph<PS, S>(paragraphStyle, Arrays.asList(seg));
-        res.add(content);
-        return new ReadOnlyStyledDocument<>(res);
-    }
```
TomasMikula: Why is this method removed?

richtextfx/src/main/java/org/fxmisc/richtext/model/SegmentOps.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public SEG subSequence(SEG seg, int start, int end);
+
+    public SEG subSequence(SEG seg, int start);
+
+    public SEG append(SEG seg, String str);
```
TomasMikula: Since this method is not possible to implement in general, we should probably replace it with something else.

TomasMikula: Applies to some other methods as well.

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
```diff
@@ -0,0 +1,1371 @@
...
+        Duration duration = Duration.ofMillis(Math.round(javafxDuration.toMillis()));
+        EventStream<?> ticks = EventStreams.restartableTicks(duration, restartImpulse);
+        return StateMachine.init(false)
+                .on(restartImpulse.withDefaultEvent(null)).transition((state, impulse) -> true)
+                .on(ticks).transition((state, tick) -> !state)
+                .toStateStream();
+    }
+
+
+    public static <SEG, S> Node createStyledTextNode(SEG seg, SegmentOps<SEG, S> segOps, BiConsumer<? super TextExt, S> applyStyle) {
```
TomasMikula: This methods should be specialized to `StyledText`, since it doesn't work for and shouldn't be used with any other segments, right?

richtextfx/src/main/java/org/fxmisc/richtext/model/SegmentOps.java
JordanMartinez: Should there be a separate "TextOps" that adds these methods so that other text-related methods can still implement it?

JordanMartinez: Meaning, if SegmentOps was only:
````java
public interface SegmentOps<SEG, S> {
    char charAt(SEG seg, int index);
    int length(SEG seg);
    String getText(SEG seg);
    boolean canJoin(SEG currentSeg, SEG nextSeg);
    S getStyle(SEG seg);
    SEG setStyle(SEG seg);
}
````
then `TextOps` could be
````java
public interface TextOps<SEG, S> extends SegmentOps<SEG, S> {
    SEG subsequence(SEG seg, int start, int end);
    SEG append(SEG seg, String text);
    // other text-related methods
}
````
and then `StyledText` could implement `TextOps` as well as other classes like `Hyperlink` or some other custom class that is text-based.

.... But, having written all of that, I don't see why we would do that since `Hyperlink` would have a separate model (`displayedText` and `linkText`) and use a separate `SegmentOps` implementation anyway... :sweat_smile: :laughing:

Comment (TomasMikula) [inline richtextfx/src/main/java/org/fxmisc/richtext/model/SegmentOps.java]: I meant, how would `StyledTextArea` or `ReadOnlyStyledDocument` use `TextOps`? All they have access to is `SegmentOps`.

Comment (JordanMartinez) [inline richtextfx/src/main/java/org/fxmisc/richtext/model/SegmentOps.java]: Good point

File: richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java (hunk 7 above)
Comment (afester) [inline richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java]: That is correct. I moved it to `StyledTextArea` and replaced the type parameter for `SEG` with `StyledText<S>`.

File: richtextfx/src/main/java/org/fxmisc/richtext/model/ReadOnlyStyledDocument.java (hunk 5 above)
Comment (afester) [inline richtextfx/src/main/java/org/fxmisc/richtext/model/ReadOnlyStyledDocument.java]: I am not completely sure which diff that is - this should be the method which creates a paragraph from an arbitrary segment, it might have been removed when merging from the genericSegment branch, but it has been re-added in https://github.com/TomasMikula/RichTextFX/pull/356/commits/ef989df7fd38d24293cd9f1efa7dbb79727a8bf3


```
public static <PS, SEG, S> ReadOnlyStyledDocument<PS, SEG, S> fromSegment(SEG segment,  PS paragraphStyle, S style, SegmentOps<SEG, S> segmentOps) {
    List<Paragraph<PS, SEG, S>> res = new ArrayList<>(1);
    res.add(new Paragraph<>(paragraphStyle, segmentOps, segment));
    return new ReadOnlyStyledDocument<>(res, segmentOps);
}
```

I have adjusted it to the original code using Arrays.asList(), probably thats even better ....

File: richtextfx/src/main/java/org/fxmisc/richtext/model/SimpleEditableStyledDocument.java (hunk 4 above)
Comment (afester) [inline richtextfx/src/main/java/org/fxmisc/richtext/model/SimpleEditableStyledDocument.java]: SimpeEditableStyledTextDocument, probably - inherited from GenericEditableStyledDocument ...
I was thinking the same already, let me see what consequences that would have ;-)

File: richtextfx/src/main/java/org/fxmisc/richtext/model/SegmentOps.java (hunk 6 above)
Comment (JordanMartinez) [inline richtextfx/src/main/java/org/fxmisc/richtext/model/SegmentOps.java]: Getting back to your previous point, what should the method (and other methods) be changed to?

Comment (TomasMikula) [inline richtextfx/src/main/java/org/fxmisc/richtext/model/SegmentOps.java]: See [my PR](https://github.com/afester/RichTextFX/pull/5).

And funny enough, I did add `TextOps extends SegmentOps` 😄 . Although `ReadOnlyStyledDocument` still sees only `SegmentOps`. The only additional method is 

```java
public SEG create(String text, S style);
```

In `SegmentOps, the original methods `append` and `canJoin` are replaced by a single method

```java
public Optional<SEG> join(SEG currentSeg, SEG nextSeg);
```

afester:
@TomasMikula thanks for the code cleanup! One final question on the overall approach: say we want to define an area which supports StyledText, LinkedImage, Rectangle and Hyperlinks as custom objects - that would end up in something like 

```
private final GenericStyledArea<ParStyle, Either<StyledText<TextStyle>, 
                                                 Either<LinkedImage<TextStyle>, 
										                Either<Rectangle<TextStyle>, 
                                                               HyperLink<TextStyle>>>>>, 
                                    TextStyle> area =
	...
```

correct? I have just added a `Rectangle ` object as a proof of concept, and it was basically a text replacement (inserting another level of `Either` in a couple of places within `RichText.java`), and implementing the classes for `Rectangle` and `RectangleOps`, but since Java, as we know, has no type aliases, there is no simple shortcut to the above, right? (besides implementing again a polymorphic hierarchy of `CustomObjects` and use only the base `CustomObject` as the right part of the Either container, and then again implementing serialization through reflection or similar) My concern really is that the code becomes quite difficult to read when there are a lot of custom object types. On the other hand, I definitely see how that approach provides a type-safe implementation :+1: ...

richtextfx/src/main/java/org/fxmisc/richtext/model/SegmentOps.java
JordanMartinez: :smile: ironic...

JordanMartinez:
> One final question on the overall approach: say we want to define an area which supports StyledText, LinkedImage, Rectangle and Hyperlinks as custom objects - that would end up in something like [nested Either approach], correct?

@afester probably not because the nesting gets confusing and tedious after a while:
````java
Either<A, Either<B, Either<C, D>>> firstLevelEither = //
if (firstLevelEither.isRight()) {
    secondLevelRight = firstLevelEither.getRight();
    if (secondLevelRight.isRight()) {
        thirdLevelRight = secondLevelRight.getRight();
        return thirdLevelRight.isRight()
            ? thirdLevelRight.getRight()
            : thirdLevelRight.getLeft()
    } else {
        return secondLevelRight.getLeft();
    }
} else {
    return firstLevelEither.getLeft();
}
````

We should design a new `Either`-like class that can take more generics in a single line:
````java
public interface OneOf<A, B, C, D> {
    boolean is1();
    boolean is2();
    boolean is3();
    boolean is4();

    OneOf<A, B, C, D> as1(A a);
    OneOf<A, B, C, D> as2(B b);
    OneOf<A, B, C, D> as3(C c);
    OneOf<A, B, C, D> as4(D d);

    // etc.
}
````
Then you would have
````java
private final GenericStyledArea
                   <
                    ParStyle, 
                    OneOf<
                        StyledText<TextStyle>, 
                        LinkedImage<TextStyle>, 
                        Rectangle<TextStyle>, 
                        Hyperlink<TextStyle>
                      >
                   >
````
To distinguish `OneOf` from other variants of the class that have more or less generics, we could add a number after it:
````java
public interface OneOf3<A, B, C> { /* code */ }
public interface OneOf4<A, B, C, D> { /* code */ }
public interface OneOf5<A, B, C, D, E> { /* code */ }
public interface OneOf6<A, B, C, D, E, F> { /* code */ }
````
since adding a spelled out number just takes up more space:
````java

public interface OneOfThree<A, B, C> { /* code */ }
public interface OneOfFour<A, B, C, D> { /* code */ }
public interface OneOfFive<A, B, C, D, E> { /* code */ }
public interface OneOfSix<A, B, C, D, E, F> { /* code */ }
````
Then the question becomes "should `Either<L, R>` just become `OneOf2<A, B>` so that it can be quickly changed if another generic is desired?

**Comment by TomasMikula** [comment]
> but since Java, as we know, has no type aliases, there is no simple shortcut to the above, right? (besides implementing again a polymorphic hierarchy of CustomObjects and use only the base CustomObject as the right part of the Either container, and then again implementing serialization through reflection or similar)

Lack of type aliases is a real pain. Nevertheless, the user is welcome to define their hierarchy of custom objects extending from `CustomObject`. They don't even need to resort to reflection or such, because they can make their hierarchy _**closed**_, i.e. don't allow implementations outside of the package. The codec can then use a few first bits to distinguish which type of object is being (de)serialized, similar how `Codec<Either<L, R>>` uses one boolean to distinguish between left/right.

The `CustomObject` can even be just a wrapper for

```java
Either<StyledText<TextStyle>, 
       Either<LinkedImage<TextStyle>, 
              Either<Rectangle<TextStyle>, 
                     HyperLink<TextStyle>>>>>
```

and the `Codec` and `TextOps` can just defer to that. To make the definition of `SegmentOps`, `TextOps`, `Codec` easier for such wrappers, we could provide methods like

```java
interface SegmentOps<SEG, S> {

    public <T> SegmentOps<T, S> translate(
            Function<T, SEG> to,
            Function<SEG, T> from
    ) {
        // straightforward but tedious implementation
    }

}
```

TomasMikula:
This should also alleviate the need for disjunctions of higher arity that @JordanMartinez proposes.

afester:
> Nevertheless, the user is welcome to define their hierarchy of custom objects with CustomObject as the root.

Exactly, that was my thinking also.

TomasMikula:
The key point is that user's custom object hierarchy can be _closed_, while ours would have to be _open_. It is this openness that causes problems.

afester:
Quick intermediate status: I am currently working on the comment by @JordanMartinez (source code compatibility with SimpleEditableStyledDocument - removing the SEG type parameter is straight forward, but it seems that we will still need the additional parameter for the segment operations in the constructors). 
@TomasMikula besides that, what would be required in addition to accept this pull request? :smile:

JordanMartinez:
> it seems that we will still need the additional parameter for the segment operations in the constructors

Mm... I was hoping we could have better source compatibility than that. However, I'm guessing the issue lies in keeping `SimpleEditableStyledDocument` `final`. 

We could 

1. make the class extend-able by not making it final, but that isn't the original direction Tomas wanted
2. make a non-final package-private base class, which the two final variants extend by adding a constructor that handles that code: the first is the SESD as it stands in your current commit history as of this comment (SEG type parameter is included); the second is the SESD as it stands now in the stable version of this project (SEG type parameter is specified as (`StyledText<S>`). This has a small amount of duplicate code in constructors
3. or design another class that wraps the SESD, constructs it with the `StyledText<S>` as the SEG type, implements `EditableStyledDocument` and delegates calls to the SESD (tedious and adds a small but unnecessary amount of overhead).

How would the second approach sound?

iblader:
发送自我的 BlackBerry - 最安全的移动设备
发件人： zhangchunnan@yzhtech.com
发送时间： 2016年10月19日 下午5:06
收件人： huangkai@yzhtech.com; lixiaoliang@yzhtech.com; qiudongyue@yzhtech.com; cailingxia@yzhtech.com
主题： 枪弹安全防控系统DEMO版本报价单--2016.10.19



demo版本的系统只包含枪支跟踪相关功能。



需求、设计、开发、测试总时间为1.5月




报价单已经传到confluence





----
张春男
北京英智兴达信息技术有限公司
地址： 北京市海淀区朱房路临甲66号顺事嘉业创业园B栋D单元二层
办公电话： 86-10-62385098-805
移动电话： 86-18600404627
传真： 86-10-62351196
电邮： z.cn@yzhtech.com

iblader:
发送自我的 BlackBerry - 最安全的移动设备
发件人： zhangchunnan@yzhtech.com
发送时间： 2016年10月19日 下午11:14
收件人： huangkai@yzhtech.com
主题： 枪弹安全防控系统报价单（全功能）--2016.10.19




----
张春男
北京英智兴达信息技术有限公司
地址： 北京市海淀区朱房路临甲66号顺事嘉业创业园B栋D单元二层
办公电话： 86-10-62385098-805
移动电话： 86-18600404627
传真： 86-10-62351196
电邮： z.cn@yzhtech.com

JordanMartinez:
@iblader I think you wrote your message in the wrong thread....?

afester:
I was able to remove the loop-through of the segment operations through `ReadOnlyStyledDocument`, so that `ReadOnlyStyledDocument` does not need to store the `segOps` - this bloats the code a bit less. After all, the segment operations are only required by `Paragraph`, i.e. where new `Paragraph` instances are created.

Then, I also moved `SimpleEditableStyledDocument` to `GenericEditableStyledDocument` and made it non-final. This could be an issue as @JordanMartinez mentioned above - after all, `SimpleEditableStyledDocument` itself is still final, but it is now possible to subclass `GenericEditableStyledDocument` of course. If that is indeed problematic, we should consider the second approach described above.

JordanMartinez:
@afester Sorry that I wasn't clearer. We only made `EditableStyledDocument` public only because it was also `final`. Tomas wants to encourage good practices via composability and prevent inheritance in this matter. So, I was laying out our options in my previous comment, but I believed that we should take the second approach. The first approach was already dismissed in my mind, but my phrasing of that wasn't as clear as it should have been. The third approach seemed unnecessary in my mind because the second one is better.

Also, I don't think we need both a `GenericStyledTextAreaModel` and a `StyledTextAreaModel`, only one should suffice as these are package-private classes, correct?

JordanMartinez:
[See my PR](https://github.com/afester/RichTextFX/pull/6) that addresses these two things.

JordanMartinez:
I tested out all the demos to make sure they work (they do).

However, I found an issue with the RichText demo. Follow these steps to reproduce the bug:

- Run demo
- insert some text by typing (e.g. "hello")
- position caret to the right of text
- insert image to the right of text using the insert image button (I inserted a PNG & a JPEG file)
- move caret to the right of the image
- type in some text (e.g. "aaa")
- BUG: image is inserted a second time to the right of first character and caret is positioned to the right of that first character

This bug reappears every time a text is inserted via user typing to the right of an image, but does not occur when text is inserted to left of it.

JordanMartinez:
Aye, the bug is caused due to `LinkedImageOps#subsequence` returning back its own segment

afester:
> The bug might be caused by LinkedImageOps#subsequence returning back its own segment

Yes, indeed ... I remember that I had a check for the start/end parameters in earlier code, since the method should require to return an empty segment - but, what is an empty `LinkedImage` segment?

The following code fixes the issue - if that is reasonable I can prepare a "real" fix, probably using some `LinkedImage.empty()` method which creates an empty `LinkedImage` segment:

```
    public LinkedImage<S> subSequence(LinkedImage<S> seg, int start) {
        if (start > 0) {
            return new LinkedImage<>("", getStyle(seg));
        }
        return seg;
    }
```

JordanMartinez:
> Yes, indeed ... I remember that I had a check for the start/end parameters in earlier code, since the method should require to return an empty segment if the range is outside the segment's length - but, what is an empty LinkedImage segment?

Perhaps we could wrap the returned object in an `Optional` and then be able to check if that optional contains anything in the package-private code. Thus, `SegmentOps` would look like:
````java
public interface SegmentOps<SEG, S> {
    public Optional<SEG> subSequence(SEG seg, int start);
}
````

and the implementation of it would be:
````java
public Optional<LinkedImange<S>> subSequence(LinkedImage<S> seg, int start) {
    return length() < start
             ? Optional.empty()
             : seg;
    }
````

File: richtextfx/src/main/java/org/fxmisc/richtext/model/SegmentOps.java [hunk 8]
```diff
@@ -68,15 +68,19 @@ public String getText(Either<L, R> seg) {
...
 
     @Override
-    public Either<L, R> subSequence(Either<L, R> seg, int start, int end) {
-        return seg.map(l -> lOps.subSequence(l, start, end),
-                       r -> rOps.subSequence(r, start, end));
+    public Optional<Either<L, R>> subSequence(Either<L, R> seg, int start, int end) {
+        return seg.unify(ll -> seg.unify(l -> lOps.subSequence(l, start, end).map(Either::left),
+                                         e -> Optional.empty()),
+                         rr -> seg.unify(e -> Optional.empty(),
+                                         r -> rOps.subSequence(r, start, end).map(Either::right)));
```
TomasMikula: This code can be simplified to

```java
return seg.unify(l -> lOps.subSequence(l, start, end).map(Either::left),
                 r -> rOps.subSequence(r, start, end).map(Either::right));
```

richtextfx/src/main/java/org/fxmisc/richtext/model/SegmentOps.java
```diff
@@ -68,15 +68,19 @@ public String getText(Either<L, R> seg) {
...
 
     @Override
-    public Either<L, R> subSequence(Either<L, R> seg, int start) {
-        return seg.map(l -> lOps.subSequence(l, start),
-                       r -> rOps.subSequence(r, start));
+    public Optional<Either<L, R>> subSequence(Either<L, R> seg, int start) {
+        return seg.unify(ll -> seg.unify(l -> lOps.subSequence(l, start).map(Either::left),
+                                         e -> Optional.empty()),
+                         rr -> seg.unify(e -> Optional.empty(),
+                                         r -> rOps.subSequence(r, start).map(Either::right)));
```
TomasMikula: Similarly here.

richtextfx/src/main/java/org/fxmisc/richtext/model/Paragraph.java
```diff
@@ -132,7 +132,8 @@ public String substring(int from) {
             int segIdx = pos.getMajor();
             List<SEG> segs = new ArrayList<>(segIdx + 1);
             segs.addAll(segments.subList(0, segIdx));
-            segs.add(segmentOps.subSequence(segments.get(segIdx), 0, pos.getMinor()));
+            segmentOps.subSequence(segments.get(segIdx), 0, pos.getMinor())
+                            .ifPresent(seg -> segs.add(seg));
```
TomasMikula: I'm afraid that this has the potential of breaking the invariant that [each Paragraph has at least one segment](https://github.com/afester/RichTextFX/blob/e40cf78eba2c3adcca77c52554a3fa2e26ee6c94/richtextfx/src/main/java/org/fxmisc/richtext/model/Paragraph.java#L42).

richtextfx/src/main/java/org/fxmisc/richtext/model/Paragraph.java
```diff
@@ -146,7 +147,8 @@ public String substring(int from) {
             Position pos = navigator.offsetToPosition(start, Forward);
             int segIdx = pos.getMajor();
             List<SEG> segs = new ArrayList<>(segments.size() - segIdx);
-            segs.add(segmentOps.subSequence(segments.get(segIdx), pos.getMinor()));
+            segmentOps.subSequence(segments.get(segIdx), pos.getMinor())
+                            .ifPresent(seg -> segs.add(seg));
```
TomasMikula: Same here, could result in no segments.

richtextfx/src/main/java/org/fxmisc/richtext/model/Paragraph.java
JordanMartinez: So, does the way `Paragraph` handle the code need to be changed? Or is the `Optional` approach itself flawed?

richtextfx/src/main/java/org/fxmisc/richtext/model/SegmentOps.java
afester: Getting a `Type mismatch: cannot convert from Object to Optional<Either<L,R>>` with that - since `subSequence()` now returns an `Optional<>`, don't I have to consider `lOps/rOps.subSequence()` being empty, similar to what `SegmentOps.join()` operation already does:

    public Optional<Either<L, R>> join(Either<L, R> left, Either<L, R> right) {
        return left.unify(ll -> right.unify(rl -> lOps.join(ll, rl).map(Either::left), 
                                                           rr -> Optional.empty()),
                                  lr -> right.unify(rl -> Optional.empty(), 
                                                           rr -> rOps.join(lr, rr).map(Either::right)));
    }

Still getting started with reactfx, so sorry if that is a trivial question :smile:

TomasMikula: Unlike the `join` method, you are calling `unify` twice on the same value (`seg`). Therefore, the `e -> Optional.empty()` is never executed, because we already know it is the other case.

Seems like type inference cannot figure out the types. Try replacing `Either::left` with `Either::<L, R>left`.

TomasMikula: The `map` method on `Optional` already handles the empty case&mdash;keeps it empty.

richtextfx/src/main/java/org/fxmisc/richtext/model/Paragraph.java
TomasMikula: I don't think it is fundamentally flawed. We just need to somehow preserve at least one segment.

JordanMartinez: I tried this out, but it doesn't work:
````java
Position pos = navigator.offsetToPosition(length, Backward);
int segIdx = pos.getMajor();
List<SEG> segs = new ArrayList<>(segIdx + 1);
segs.addAll(segments.subList(0, segIdx));
segmentOps.subSequence(segments.get(segIdx), 0, pos.getMinor())
                .ifPresent(segs::add);

return segs.isEmpty()
        ? this
        : new Paragraph<>(paragraphStyle, segmentOps, segs);
````
Even debugging the application did not go through this section. My guess is the bug is caused at the ReadOnlyStyledDocument level

Comment (afester) [inline richtextfx/src/main/java/org/fxmisc/richtext/model/Paragraph.java]: I'll go ahead and create a unit test which triggers the assertion mentioned by @TomasMikula (or, which checks if the resulting Paragraph indeed still has at least one segment - with the new code, this should fail  until the issue is fixed)

Comment (afester) [inline richtextfx/src/main/java/org/fxmisc/richtext/model/Paragraph.java]: @JordanMartinez I have added a test case which, obviously, fails now - https://travis-ci.org/TomasMikula/RichTextFX/builds/171061050

The relevant area is in `Paragraph.subSequence(int start)` (the `trim()` method also needs to be fixed, but that should be trivial once we have the solution for `subSequence()`):
```
...
            segmentOps.subSequence(segments.get(segIdx), pos.getMinor())
                            .ifPresent(segs::add);
            segs.addAll(segments.subList(segIdx + 1, segments.size()));
            if (segs.isEmpty()) {
                segs.add(????);
            }
            return new Paragraph<>(paragraphStyle, segmentOps, segs);
...
```
I am unclear what to add to the `segs` list, in case the result list is empty - we would require an object of type `S` with a length of `0`, essentially ...
Alternatively we might decide that a Paragraph can now also contain an empty list of segments, but since the segment list is exposed through the API, that might influence existing applications ...

Comment (JordanMartinez) [inline richtextfx/src/main/java/org/fxmisc/richtext/model/Paragraph.java]: @afester Since my above approach didn't work, I think we will have to do something at the `ReadOnlyStyledDocument` level. We might even need to make `Paragraph#subsequence` return an `Optional<Paragraph<PS, SEG, S>>`. However, I'm not as familiar with how that code in `ReadOnlyStyledDocument` works since it uses `Tuple3`, so it's not easy to follow the logic as to what each argument is. Then again, I only glanced it and haven't really studied it.

Comment (JordanMartinez) [inline richtextfx/src/main/java/org/fxmisc/richtext/model/Paragraph.java]: If we wrap the returned value of subsequence in an `Optional`, it adds boilerplate to the code (and is honestly so confusing that I'm not even sure if I've implemented it correctly, nor would I want to ever have to read through this code). See [here](https://github.com/JordanMartinez/RichTextFX/commit/a6b130124de4133763e7b5756e13d9d2e4fb7418) for my work (Note: this does not compile as ReadOnlyStyledDocument would need to account for the new changes).

Comment (JordanMartinez) [inline richtextfx/src/main/java/org/fxmisc/richtext/model/Paragraph.java]: [See this PR](https://github.com/afester/RichTextFX/pull/7) to see what updating the code to use Empty and NonEmpty variants of `Paragraph` would look like.

Comment (afester) [inline richtextfx/src/main/java/org/fxmisc/richtext/model/Paragraph.java]: > If we wrap the returned value of subsequence in an Optional, it adds boilerplate to the code (and is honestly so confusing that I'm not even sure if I've implemented it correctly, nor would I want to ever have to read through this code).

Indeed - I also tried that last week and gave up at the `Paragraph.restyle()` methods - I **was** pretty sure that I would not be able to implement that correctly :joy: 

I'll look into your PR and let you know my progress...

**Comment by JordanMartinez** [comment]
I manually tested the demos and they all work. Still, I'd like @TomasMikula to look over the code change of the Paragraph variants commit. No doubt he'd be able to spot any issues/optimizations in the code as he knows that area of the code better than I do.

**Comment by JordanMartinez** [comment]
@afester I'm not sure how much more this code has to go before its finished and merged into Tomas' project, but would it be too much to ask for you to clean up the commit history a bit? This PR will (currently) merge 95 commits into the branch. Multiple commit messages are unclear as to what was done in that commit and other commits override previous ones. 

There are the two views of commit history: one sees it as development, so however it was developed—warts and all—gets merged into the code as is while the other sees it as a story, so the development process gets cleaned up before merging so others can see the development in a neater way. Although Tomas has not specified which view "wins" in his project, I think it'd be more beneficial for others if it was cleaned up.

**Comment by JordanMartinez** [comment]
We should also add a few more tests to `SimpleEditableStyledDocument` to account for the usage of the two variants being used within the same document. At the moment, I'm not sure what those additional tests should be.

**Comment by afester** [comment]
> but would it be too much to ask for you to clean up the commit history a bit?

Not at all @JordanMartinez :+1: 

My understanding is that there is no way to remove "intermediate" commits within a branch which has already been pushed. So, the approach would be to create a new branch, commit the code from this PR to it in chunks which make sense (resulting in probably 3 to 4 commits which reflect the larger pieces of work we did), then submit a new PR for that new branch and close this PR. Is that correct?

**Comment by JordanMartinez** [comment]
Great!

Yes on everything but the new PR. Just force push the new branch to this
PR's branch and it'll override the commits and their history with the
updated branch. No need to create a new PR and disassociate all our
conversation here from the work done.

**Comment by JordanMartinez** [comment]
@afester After thinking about it more, I'm pretty sure there is a bug in the Paragraph commit I wrote. If a string with newline characters is inserted, I don't think there is a check to determine which Paragraph variant to create. In other words, since it defaults to creating only `NonEmptyParagraph`, it will create that kind rather than the `EmptyParagraph` when a blank line is the argument.

**Comment by JordanMartinez** [comment]
@afester The Paragraph variant approach is also flawed because the code that handles the `EmptyParagraph` creation does not know what the `paragraphStyle` should be. (This was probably the reason we did not implement this last time, but it's been so long that I couldn't remember that.)

For example, in the `NonEmptyParagraph`, calling either `trim(0)` or `subsequence(length())` should create an `EmptyParagraph`. However, the only style at hand for the creation is the non-empty paragraph's own paragraph style. This minor point screws up ReadOnlyStyledDocument's `testUndo`:
````java
org.fxmisc.richtext.model.ReadOnlyStyledDocumentTest > testUndo FAILED
    java.lang.AssertionError at ReadOnlyStyledDocumentTest.java:23
Expected :Par[X; ]
Actual   :Par[Y; ]
````
because the change that is undone cannot know what the paragraph style of what it replaced is, nor that it should handle the code differently in this specific instance.

A "default style" for the paragraph style won't work either because some empty paragraphs can and should be styled differently (e.g. bullet points).

**Comment by JordanMartinez** [comment]
As an alternative, we can go back to the non-variant Paragraph way of handling things but include another method in `SegmentOps` that creates an empty segment with a default style. This segment can then be added to the list of segments that are then passed into the Paragraph's constructor. I'm testing this approach out now.

**Comment by JordanMartinez** [comment]
@afester I'm not sure if you saw [my previous comment](https://github.com/TomasMikula/RichTextFX/pull/356#issuecomment-259593957) about the flawed approach of using an Empty and NonEmpty `Paragraph`s. If you see [this PR](https://github.com/afester/RichTextFX/pull/9), which constructs the non-empty paragraph when it should, the `ReadOnlyStyledDocumentTest#testUndo` fails.

As an alternative, I suggested in my [last comment](https://github.com/TomasMikula/RichTextFX/pull/356#issuecomment-259595497) that we don't use this approach and instead include another method in `SegmentOps` that creates an empty segment with a default style. Then, we can construct that empty segment to add to the list of segments when it is empty. This will avoid that Paragraph's assertion that the passed-in segments are not empty.

**Comment by afester** [comment]
@JordanMartinez Yes I did - I just wanted to finish the commit squashing.  Will check your proposals next ...

**Comment by JordanMartinez** [comment]
@afester Ah... ok. That makes sense.

**Comment by afester** [comment]
@JordanMartinez Did you remove https://github.com/JordanMartinez/RichTextFX/commits/alternativeApproach? I get a 404 when I click the link from your comment ...

**Comment by JordanMartinez** [comment]
@afester My apologies. Yes, I mistakenly did so because I thought it was the `Optional` subsequence approach.

Do you want me to redo that work? Or will you implement it yourself?

**Comment by afester** [comment]
@JordanMartinez if you can still remember what you implemented, it would be great if you could redo this work ... :smiley:

**Comment by JordanMartinez** [comment]
See my work [here](https://github.com/JordanMartinez/RichTextFX/tree/emptySeg).

**Comment by afester** [comment]
@JordanMartinez I am getting a NullPointerException in the RichTextEditor demo when I insert an image into an empty document, set the cursor before the image and enter more than two letters:

```
Exception in thread "JavaFX Application Thread" java.lang.NullPointerException
	at org.fxmisc.richtext.demo.richtext.RichText.lambda$18(RichText.java:79)
	at org.fxmisc.richtext.StyledTextArea.createStyledTextNode(StyledTextArea.java:104)
	at org.fxmisc.richtext.demo.richtext.RichText.createNode(RichText.java:292)
	at org.fxmisc.richtext.demo.richtext.RichText.lambda$1(RichText.java:79)
	at org.fxmisc.richtext.ParagraphText.<init>(ParagraphText.java:102)
	at org.fxmisc.richtext.ParagraphBox.<init>(ParagraphBox.java:77)
	at org.fxmisc.richtext.GenericStyledArea.createCell(GenericStyledArea.java:1176)
        ...
```

I'll continue to investigate this issue, but if you have any additional feedback please let me know :smile:

JordanMartinez:
@afester if it's a null style, the fix is very easy. If you look at `StyledText`'s textOps (which returns the SegmentOps for StyledText), you'll see I passed in a `null` value for the style when creating an empty segment. 
I didn't know whether this would have any adverse affects, but I thought it would be fine. It also made the code easier to write because one wouldn't have to pass in an empty style when calling that method. 

I've changed it on my side and the NPE goes away now.

However, if I follow the steps you said above

1. Insert image in empty area
2. Move caret to be in front of image
3. Type "j"
4. Type "s"

An odd thing occurs on step 3. Once "j" is typed, the caret moved left one position, so that the next letter I insert ("s") is then inserted in front of the "j". Thus I have "sj[image]".

Do you know why this might occur? It only occurs when the image is the first segment in a paragraph and the caret is at position 0 in the line (i.e. to the left of the image).

JordanMartinez:
We'll definitely need to write a guide to help developers implement this code correctly as there are a number of things to work out.

JordanMartinez:
@TomasMikula I believe this custom object PR is now ready to be merged.

JordanMartinez:
@afester One thought that came to mind today that I haven't thought about / investigated more: will this PR break undo/redo?

Jugen:
Provided the following three (3) conditions are met undo/redo generally works:

1.  The preserveStyle parameter must be TRUE when invoking the GenericStyledArea constructor.

2.  A properly implemented Codec for your SEG must be set via GenericStyledArea.setStyleCodecs

3.  The Object.equals( Object obj ) method must be overridden and properly implemented by your SEG

afester:
@Jugen Thanks for that analysis - I believe that this is the kind of information we need to add to the developer's guide which was proposed by @JordanMartinez some time ago.
I did some basic testing with undo/redo and did not observe any issue with the RichText demo (using the LinkedImage segment), but probably more testing is required. I am currently evaluating how to implement the concrete component I have in mind, so I will probably find some additional issues there ...

JordanMartinez:
Let's move any discussion about the developer guide to the issue I just opened.

TomasMikula:
Thanks guys for moving things forward! @afester @JordanMartinez are you both happy with the current state of this PR?

JordanMartinez:
@TomasMikula Glad to see you're back again! Yes, I think this PR is good.

afester:
@TomasMikula Me too !

Happy holidays

TomasMikula:
🎉

Issue #360, opened 2016-09-18
Image Info next to the lines


GweltazCalori:
Hi,
I wanted to know if it's possible to have image info next to line numbers. For example in Intellij you can have an info next to the line.
![intel](https://cloud.githubusercontent.com/assets/12122821/18617493/f21a5940-7dd0-11e6-823f-3cdfc777f6ec.PNG)

I used LineNumberFactory to add line but I can't find abything about adding infos.

TomasMikula:
See [this StackOverflow answer](http://stackoverflow.com/questions/28659716/show-breakpoint-at-line-number-in-richtextfx-codearea/28660913#28660913). Perhaps we should add it to the demos.

JordanMartinez:
Closed via #364.

PR #364, opened 2016-09-19
Demonstrate via demo the usage of StyledTextArea#paragraphGraphicFactory

JordanMartinez:
Extracted this demo from Tomas' StackOverflow answer

Since you mentioned it in #360, I thought it would be trivial to do and very helpful for others to have.


TomasMikula:
Thanks! 👍

Issue #350, opened 2016-08-23
MenuBar Mnemonics are ignored


jakob-erdmann:
When focus is set onto the RichtText area and a `MenuBar` is added to the same `Stage`, pressing ALT will cause the `MenuBar` to highlight the first element, yet hitting any mnemonic key will not trigger the event to open the menu.

The mnemonic is working when focus is on any other element, so it's not a problem with JavaFX menu mnemonics. Also I tried overriding the event handler for `onKeyTypedProperty` to not consume any event which did not change anything on the bugged behavior. So besides the text area printing keys when ALT key is hold down and thus not letting mnemonics kick in, the event is somewhere caught in the middle and thus not propagated to menu properly.

TomasMikula:
> The mnemonic is working when focus is on any other element

Can you confirm it works on `javafx.scene.control.TextArea`?

> Also I tried overriding the event handler for `onKeyTypedProperty`

Versions `0.7-*` use `addEventHandler` to register event handlers. Therefore, setting `onKeyTypedProperty` doesn't override anything. You could use `addEventFilter` to prevent handling (and consuming) a specific event type.

> So besides the text area printing keys when ALT key is hold down and thus not letting mnemonics kick in

I would say this is the correct behavior. <kbd>Alt</kbd>+<kbd>?</kbd> is a valid way of typing certain letters.

jakob-erdmann:
> Can you confirm it works on `javafx.scene.control.TextArea`?

Indeed i can, check out this code: https://gist.github.com/jakob-erdmann/be3eb9a658135f610ea971e983bde098

The upper area is a `TextArea` with <kbd>Alt</kbd>+<kbd>F</kbd> working to open the menu, the lower one is a `CodeArea` where <kbd>ALT</kbd>+<kbd>F</kbd> is printing an 'f' character.

> Versions 0.7-* use `addEventHandler` to register event handlers. Therefore, setting `onKeyTypedProperty` doesn't override anything. You could use `addEventFilter` to prevent handling (and consuming) a specific event type.

I tested this with version `0.6.10` as we saw this as your latest stable release. But even in version `0.7-M2` I am able to reproduce the described behavior.

> I would say this is the correct behavior. <kbd>Alt</kbd>+<kbd>?</kbd> is a valid way of typing certain letters.

I can agree on this as long as the combination should not be used for mnemonics or similar.

[...]

PR #365, opened 2016-09-22
Changed StyledTextAreaBehavior to not catch mnemonic combinations

ghost:
Fix for https://github.com/TomasMikula/RichTextFX/issues/350

On Unix systems Alt Gr will trigger neither Alt nor Ctrl to be pressed, on Windows it triggers both. As you want to let Alt Gr be a captured by `StyledTextArea` on Windows I changed the condition for it to let Alt be passed through and Alt Gr be captured. Although I still think that Alt Gr should also not be caught on Windows as it is not used to print characters there.

PS.: this is the private account of @jakob-erdmann

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextAreaBehavior.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextAreaBehavior.java
@@ -153,9 +153,9 @@ class StyledTextAreaBehavior {
 
         Predicate<KeyEvent> noControlKeys = e ->
                 // filter out control keys
-                (!e.isControlDown() && !e.isMetaDown())
+                (!e.isControlDown() && !e.isMetaDown() && !e.isAltDown())
                 // except on Windows allow the Ctrl+Alt combination (produced by AltGr)
-                || (isWindows && !e.isMetaDown() && (!e.isControlDown() || e.isAltDown()));
+                || (isWindows && !e.isMetaDown() && e.isControlDown() == e.isAltDown());
 
         Predicate<KeyEvent> isChar = e ->
                 e.getCode().isLetterKey() ||
--- a/richtextfx/src/test/java/org/fxmisc/richtext/StyledTextAreaBehaviorTest.java
+++ b/richtextfx/src/test/java/org/fxmisc/richtext/StyledTextAreaBehaviorTest.java
@@ -1,0 +1,52 @@
+package org.fxmisc.richtext;
+
+import javafx.scene.input.KeyCode;
+import javafx.scene.input.KeyEvent;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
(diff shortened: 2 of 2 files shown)


TomasMikula:
Thanks, looks good! Although I don't see the Windows case (both Ctrl and Alt pressed) being tested by the test cases.

We should be able to test the behavior without instantiating a `StyledTextArea`. Creating `StyledTextAreaBehavior` should suffice.

> Although I still think that Alt Gr should also not be caught on Windows as it is not used to print characters there.

My thinking is that since it produces a `KEY_TYPED` event, the corresponding `KEY_PRESSED` event should be consumed. Am I misunderstanding something?

Deroin:
> We should be able to test the behavior without instantiating a `StyledTextArea`. Creating `StyledTextAreaBehavior `should suffice.
I think I had problems getting the event handled when only having the behavior. But I can double check and extend for Alt + Ctrl case.

> My thinking is that since it produces a `KEY_TYPED` event, the corresponding `KEY_PRESSED` event should be consumed. Am I misunderstanding something?
I don't know if the `KEY_PRESSED` is automatically consumed, but yes a `KEY_TYPED` event is thrown with those modifiers. On Windows only a bunch of keys are able to create new characters (e.g. @) by this. I guess when matching the typed key with the text property of the event we should be able to identify if this was used for printing a key or not. But anyway there are key combinations like CTRL + ALT + S should be usable by some menu actions in my opinion. I will also check on this tomorrow.

TomasMikula:
> I don't know if the `KEY_PRESSED` is automatically consumed

It is certainly not consumed automatically. That's why we are writing this code to consume it, and it is hard to determine when we should and when not.

Deroin:
Well basically you only want to consume the event when it is used to print a character. And from the event object you can take the text that would be printed. So if there is no text given it probably should not be consumed. Will check the exact behavior for Windows machines, on Linux I can confirm that the event holds this data.

TomasMikula:
> Well basically you only want to consume the event when it is used to print a character.

Exactly, that's what we are trying to determine: Does this key press produce a character input?

> And from the event object you can take the text that would be printed.

There are two methods, [`getCharacter()`](http://docs.oracle.com/javase/8/javafx/api/javafx/scene/input/KeyEvent.html#getCharacter--) and [`getText()`](http://docs.oracle.com/javase/8/javafx/api/javafx/scene/input/KeyEvent.html#getText--). `getCharacter()` is defined only for key typed events, `getText()` is only defined for key pressed and key released events. We are inspecting a `KEY_PRESSED` event, so we can use only the `getText()` method. However, the string returned from `getText()` is a

> String describing the key code, such as "HOME", "F1" or "A"

A textual description of the _key code_ has nothing to do with whether character input is produced by the key press. This is why it is so difficult.

JordanMartinez:
Closing due to inactivity and otherwise unresolved questions.

PR #370, opened 2016-10-04
(Shift) + Shortcut + Left/Right skips only 1 word boundary, not 2

### LINKED ISSUE #200: (Shift)+Shortcut+Left/Right skips one too many word boundaries.
Given the sentence below where "|" represents the caret:
`I love cho|colate ice cream.`

Pressing Shift+Shortcut+Left/Right will select the remaining word and either the leading or trailing word boundary (angle brackets used to show selection "<selection>"):
- Right once: `I love cho<colate >ice cream.` instead of `I love cho<colate> ice cream.`
- Right twice: `I love cho<colate ice >cream.`
- Left once: `I love< cho>colate ice cream.`
- Left twice: `I< love cho>colate ice cream.`

Similarly, pressing Shortcut+Left/Right:
- Right once is  `I love chocolate |ice cream.` instead of `"I love chocolate| ice cream.`
- Right twice: `I love chocolate ice |cream.`
- Left once: `I love| chocolate ice cream.`
- Left twice: `I| love chocolate ice cream.`

JordanMartinez:
This fixes #200. I could write a test for it, but that would require adding a TestFX test dependency in headless mode so that the travis build still works correctly.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextAreaBehavior.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextAreaBehavior.java
@@ -1,5 +1,6 @@
 package org.fxmisc.richtext;
 
+import static java.lang.Character.isWhitespace;
 import static javafx.scene.input.KeyCode.*;
 import static javafx.scene.input.KeyCombination.*;
 import static org.fxmisc.richtext.model.TwoDimensional.Bias.*;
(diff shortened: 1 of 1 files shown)


JordanMartinez:
Any comments on this?

TomasMikula:
I'm afraid this fix would cause the caret to move like this, i.e. sometimes skipping just the space:

```
|foo bar baz
foo| bar baz
foo |bar baz
foo bar| baz
foo bar |baz
foo bar baz|
```

JordanMartinez:
.... isn't that kind of the point? If it's something you don't want in RTFX, then that issue should have been closed a long time ago.

TomasMikula:
I think this is the desired behavior:

When moving to the right:
 - If there is a space to the right of the caret, then skip two boundaries.
 - If there is a word character to the right of the caret, then skip just one boundary.

Analogously when moving to the left.

This is different from both the current state and this PR.

JordanMartinez:
The new commit should do it.

JordanMartinez:
Shouldn't this also factoring in tabs?

TomasMikula:
I would use [isWhitespace](http://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#isWhitespace-char-).

JordanMartinez:
How's that?

On another hand, it doesn't work if there are multiple blank lines between one word and the next. The caret `|` by `word1` would stop twice before reaching `word2`:
````
word1|

|

|word2
````

Is that desired?

TomasMikula:
Well, that is also the current behavior. I think it is a problem of [BreakIterator.getWordInstance()](https://docs.oracle.com/javase/8/docs/api/java/text/BreakIterator.html#getWordInstance--) that it considers a line break as a word boundary.

PR #371, opened 2016-10-04
Handle drag events only if mouse release occurred inside of view

### LINKED ISSUE #321: Unexpected text drag behavior on abort.
Hi Tomas

When dragging text and the mouse is release outside of the control then the selected text is moved generally to the beginning of the text area or some earlier location in the text.

The expected behavior I believe is that the drag should be aborted and the text should not move.

Thanks, regards
Jurgen

JordanMartinez:
This should fix #321


TomasMikula:
Shouldn't the test be inside of the `DRAG` case?

JordanMartinez:
> Shouldn't the test be inside of the DRAG case?

The caret will still move then, but that's what happens in a TextArea, too, (just tested it). So, yeah, it should.

JordanMartinez:
I've updated the PR

TomasMikula:
👍

PR #372, opened 2016-10-04
Prevent unneeded emission by filtering out negative values

### LINKED ISSUE #357: Some MouseEvent behaviors cannot be safely overridden because hooks for them do not yet exist
First, thanks for this very promising component.

Now, the observed behavior. In my code (a console widget) I wanted to reposition the cursor after a text selection was made (on mouse released, using moveTo), it worked well except when the drag ended out of the component area. Annoyed by the thing, I digged, and ended finding that it was somehow linked to hit computation and drag code ; when the drag (of selection) ends within the component all is fine, but when the drag ends outside the component the hit methods is called infinitelly and the moveTo doesnt work as expected.

Version used : the last one (7-M2), but the behavior was observed pre-7.

To reproduce/observe the problem, a "crude" approach :
- create a subclass of  StyleClassedTextArea , let's call it  StyleClassedTextAreaExt, it only overrides the hit method to trace the calls and stack

```
@Override
    public CharacterHit hit(double x, double y) {
        CharacterHit hit = super.hit(x, y);
        System.out.println("HIT CALLED : ");
        (new Exception()).printStackTrace();
        return hit;
    }
```
- create a basic fx app, border pane, and in center the StyleClassedTextAreaExt
- launch the app, not in full screen, and keep an eye on console output
- type some text in the text area
- start a text selection in the text, keep the mouse pressed, move the cursor out of the text area to the left (same to the right), and the release the mouse
- you then observe infinite calls in the logs

The observed stacks of the continuous calls are :

>   at fr.mmu.st2.console.ui.widget.console.StyleClassedTextAreaExt.hit(Console.java:246)
>   at org.fxmisc.richtext.StyledTextAreaBehavior.dragTo(StyledTextAreaBehavior.java:471)
>   at org.reactfx.value.Val.ifPresent(Val.java:151)
>   at org.fxmisc.richtext.StyledTextAreaBehavior.lambda$new$38(StyledTextAreaBehavior.java:259)
>   at org.reactfx.util.NonAccumulativeStreamNotifications.lambda$head$0(NotificationAccumulator.java:134)
>   at org.reactfx.ObservableBase.notifyObservers(ObservableBase.java:68)
>   at org.reactfx.ObservableBase.notifyObservers(ObservableBase.java:57)
>   at org.reactfx.ProperEventStream.emit(ProperEventStream.java:18)
>   at org.reactfx.util.NonAccumulativeStreamNotifications.lambda$head$0(NotificationAccumulator.java:134)
>   at org.reactfx.ObservableBase.notifyObservers(ObservableBase.java:68)
>   at org.reactfx.ObservableBase.notifyObservers(ObservableBase.java:57)
>   at org.reactfx.ProperEventStream.emit(ProperEventStream.java:18)
>   at org.reactfx.MappedStream.lambda$observeInputs$0(MappedStream.java:25)
>   at org.reactfx.util.NonAccumulativeStreamNotifications.lambda$head$0(NotificationAccumulator.java:134)
>   at org.reactfx.ObservableBase.notifyObservers(ObservableBase.java:68)
>   at org.reactfx.ObservableBase.notifyObservers(ObservableBase.java:57)
>   at org.reactfx.ProperEventStream.emit(ProperEventStream.java:18)
>   at org.reactfx.EmitBothOnEachStream.lambda$observeInputs$1(EmitBothOnEachStream.java:31)
>   at org.reactfx.util.NonAccumulativeStreamNotifications.lambda$head$0(NotificationAccumulator.java:134)
>   at org.reactfx.ObservableBase.notifyObservers(ObservableBase.java:68)
>   at org.reactfx.ObservableBase.notifyObservers(ObservableBase.java:57)
>   at org.reactfx.ProperEventStream.emit(ProperEventStream.java:18)
>   at org.reactfx.MappedStream.lambda$observeInputs$0(MappedStream.java:25)
>   at org.reactfx.util.NonAccumulativeStreamNotifications.lambda$head$0(NotificationAccumulator.java:134)
>   at org.reactfx.ObservableBase.notifyObservers(ObservableBase.java:68)
>   at org.reactfx.ObservableBase.notifyObservers(ObservableBase.java:57)
>   at org.reactfx.ProperEventStream.emit(ProperEventStream.java:18)
>   at org.reactfx.AccumulatingStream.lambda$observeInputs$0(AccumulatingStream.java:34)
>   at org.reactfx.util.NonAccumulativeStreamNotifications.lambda$head$0(NotificationAccumulator.java:134)
>   at org.reactfx.ObservableBase.notifyObservers(ObservableBase.java:68)
>   at org.reactfx.ObservableBase.notifyObservers(ObservableBase.java:57)
>   at org.reactfx.ProperEventStream.emit(ProperEventStream.java:18)
>   at org.reactfx.EventStreams$20$1.handle(EventStreams.java:421)
>   at javafx.animation.AnimationTimer$AnimationTimerReceiver.lambda$handle$484(AnimationTimer.java:57)
>   at java.security.AccessController.doPrivileged(Native Method)
>   at javafx.animation.AnimationTimer$AnimationTimerReceiver.handle(AnimationTimer.java:56)
>   at com.sun.scenario.animation.AbstractMasterTimer.timePulseImpl(AbstractMasterTimer.java:357)
>   at com.sun.scenario.animation.AbstractMasterTimer$MainLoop.run(AbstractMasterTimer.java:267)
>   at com.sun.javafx.tk.quantum.QuantumToolkit.pulse(QuantumToolkit.java:506)
>   at com.sun.javafx.tk.quantum.QuantumToolkit.pulse(QuantumToolkit.java:490)
>   at com.sun.javafx.tk.quantum.QuantumToolkit.lambda$runToolkit$404(QuantumToolkit.java:319)
>   at com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:95)
>   at com.sun.glass.ui.win.WinApplication._runLoop(Native Method)
>   at com.sun.glass.ui.win.WinApplication.lambda$null$148(WinApplication.java:191)
>   at java.lang.Thread.run(Thread.java:745)

P.S.
If it is of any value, one point I observed in the code of your component while digging the problem :
- in debug, while reproducing the problem, one can see that if you put breakpoints in StyledTextArea.hit(x,y) in the two first conditions (that should be triggered, I guess, once the mouse goes out of the component area) ; they are never triggered during the problem :

```
public CharacterHit hit(double x, double y) {
        VirtualFlowHit<Cell<Paragraph<PS, S>, ParagraphBox<PS, S>>> hit = virtualFlow.hit(x, y);
        if(hit.isBeforeCells()) {
            return CharacterHit.insertionAt(0);     <----- (breakpoint here) never true when mouse goes out
        } else if(hit.isAfterCells()) {
            return CharacterHit.insertionAt(getLength());      <----- (breakpoint here) never true when mouse goes out
        } else {
            int parIdx = hit.getCellIndex();
            int parOffset = getParagraphOffset(parIdx);
            ParagraphBox<PS, S> cell = hit.getCell().getNode();
            Point2D cellOffset = hit.getCellOffset();
            CharacterHit parHit = cell.hit(cellOffset);
            return parHit.offset(parOffset);
        }
    }
```
- what is triggerred though (and should not in this case, I guess) is in StyledTextAreaBehavior, the start auto scroll :

```
private void mouseDragged(MouseEvent e) {
        // don't respond if disabled
        if(view.isDisabled()) {
            return;
        }

        // only respond to primary button alone
        if(e.getButton() != MouseButton.PRIMARY || e.isMiddleButtonDown() || e.isSecondaryButtonDown()) {
            return;
        }

        Point2D p = new Point2D(e.getX(), e.getY());
        if(view.getLayoutBounds().contains(p)) {
            dragTo(p);
            autoscrollTo.setValue(null); // stops auto-scroll
        } else {
            autoscrollTo.setValue(p);    // starts auto-scroll     <------ HERE : triggered twice before infinite loop starts
        }

        e.consume();
    }

```

JordanMartinez:
This fixes #357. `emitBothOnEach` would continuously fire due to `distance` having a negative X value in @MichelMunoz's gif example. Since the point of origin is the top-left corner of the area, any negative values can be excluded. Once this is done, the `emitBothOnEach` EventStream, which only emits an event when both `animationFrames()` and `distance` emit a new event, only emits a new event when `distance` emits a non-negative x/y point.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextAreaBehavior.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextAreaBehavior.java
@@ -249,6 +249,7 @@ class StyledTextAreaBehavior {
                 projection,
                 Point2D::subtract);
         EventStream<Point2D> deltas = nonNullValuesOf(distance)
+                .filter(dist -> dist.getX() >= 0 && dist.getY() >= 0)
                 .emitBothOnEach(animationFrames())
                 .map(t -> t.map((ds, nanos) -> ds.multiply(nanos / 100_000_000.0)));
         valuesOf(autoscrollTo).flatMap(p -> p == null


TomasMikula:
But IIRC, negative distance is used for continuous scrolling while dragging.

Continuous firing is fine, it just needs to be terminated when mouse is released, which is not correctly detected at the moment, when the release happens outside the area.

JordanMartinez:
> Continuous firing is fine, it just needs to be terminated when mouse is released, which is not correctly detected at the moment, when the release happens outside the area.

So #371 would fix this, correct?

TomasMikula:
The problem with #357 seems like the release outside the area is never detected. #371 ignores a mouse release. To fix #357, we need to detect a mouse release that we seem to miss right now, not ignore one.

JordanMartinez:
> To fix #357, we need to detect a mouse release that we seem to miss right now, not ignore one.

But if the release occurs outside of the area, how can we detect it? It seems we would have to add a temporary listener/EventStream to the stage to detect any MouseRelease events, run the required code, and remove it once it has been detected.

TomasMikula:
> But if the release occurs outside of the area, how can we detect it?

That is to be researched ;) And what if it occurs outside of the stage?

JordanMartinez:
> And what if it occurs outside of the stage?

Then we throw an exception! Haha! In all seriousness, good point.

JordanMartinez:
@TomasMikula Did you find a way to detect changes once outside the Stage?

TomasMikula:
I haven't had time to investigate that yet :(

JordanMartinez:
I did some initial research and here's what I found:

- Of the issues on StackOverflow that somewhat addressed this, none of them were trying to do what we desire (detect a mouse-release outside of a JavaFX window). The best they came up with was to listen for mouse exit events on the Window/Stage itself.
- Since JavaFX succeeds Swing, I looked to see what solutions existed for Swing if the same issue arose. [This StackOverflow question](https://stackoverflow.com/questions/8586098/global-event-listeners-with-awteventlistener-how-to-pull-mouseevents-from-it) provides an example that uses `Toolkit` to add a listener that can track global mouse events. However, I don't yet know if such a thing exists for JavaFX. I'm also not sure if we can (or should) use the `Toolkit` for our purposes here...
- There's also [JNativeHook](https://github.com/kwhat/jnativehook), but I think it's license (GPLv3) is incompatible with this project's license.
- Lastly, I just came across [System-Hook[(https://github.com/kristian/system-hook) which is licensed under MIT. However, it seems that it only works on Windows.

JordanMartinez:
Actually, [this StackOverflow question](https://stackoverflow.com/questions/26610660/how-to-drag-a-javafx-node-and-detect-a-drop-event-outside-the-javafx-windows) demonstrates a way to do such a thing. We might need to modify their approach, but I think it would work.

JordanMartinez:
> The problem with #357 seems like the release outside the area is never detected. #371 ignores a mouse release. To fix #357, we need to detect a mouse release that we seem to miss right now, not ignore one.

Actually, it seems like there's no longer a bug. I added a `println` statement and the mouse release is detected when it occurs outside of the area, thus `autoscrollTo` is set back to null and terminates the emission.... huh, weird.

PR #375, opened 2016-10-04
Relayout popup when any of its settings are invalidated

### LINKED ISSUE #128: Two problems positioning popup relative to caret
1. The positioning does not work right after the editors text was changed by calling
StyledTextArea.replaceText(int start, int end, String text)
and the caret was moved by the change. (A typically use case that does so is code completion). 
After the next user edit positioning works alright again.

2.  With an initially opened and focused editor positioning the popup at the caret  bottom (PopupAlignment.CARET_BOTTOM) does not work. It shows up somewhere in the upper half of the caret.

Reproduction:
Both problems can be seen in the SQuirreL SQL FX snapshot 20150313_1200 which is based on RichTextFX 0.6.3, see
http://sourceforge.net/projects/squirrel-sql/files/sqfx/squirrelsqlfx-snapshot-20150313_1200.zip/download

In the SQL editor use the code completion short cut (ctrl+space) to display the code completion popup.

JordanMartinez:
This should fix #128's second problem (the previous PR had a weird commit history issue).


TomasMikula:
👍

PR #378, opened 2016-10-06
Request follow caret

### LINKED ISSUE #101: StyledTextArea scrolls to the caret every time span style is changed
I have a StyleClassedTextArea and load a String into it. I have a button event which calls setStyleClass a number of times using data external to the document. For some reason, when I call it, the text area always automatically scrolls to the bottom of the document. I have tried all kinds of variations, but it happens. Also, when I make the setStyleClass dependent on the document's content changing and I edit the text in the middle of the text field, it scroll-jumps into a position where the line I am editing becomes the bottom-most line in the view. What am I doing wrong?

Thanks,

Matthias

------

Minimal Code Example (just paste a lot of text into the field and click the button and provide some stylesheet in the css file defining a 'red' style):

------

    package application;

    //imports ...

    public class Main extends Application {
	
	private Scene scene;
	private StyleClassedTextArea textArea;
	
	@Override
	public void start(Stage stage) {
		
		stage.setTitle("ScrollTest);
        
		try {
			BorderPane root = new BorderPane();
			textArea = new StyleClassedTextArea();
			textArea.setWrapText(true);
		
			BorderPane border = new BorderPane();
			HBox topBox = new HBox();
		    
		    Button updateButton = new Button("foo");			
		    updateButton.setPrefSize(100, 20);
		    updateButton.setOnAction(new EventHandler<ActionEvent>() {
		    	@Override public void handle(ActionEvent e) {
		    		textArea.setStyleClass(5, 10, "red");
		        }
		    });
	
		    
		    topBox.getChildren().addAll(updateButton);
		    
			scene = new Scene(border, 600, 400);
			border.setCenter(textArea);
			
			border.setTop(topBox);
            scene.getStylesheets().add(getClass().getResource("application.css").toExternalForm());
			
			stage.setScene(scene);
			stage.show();
			
		} catch(Exception e) {
			e.printStackTrace();
		}
	}
	
		launch(args);
	}
    }

JordanMartinez:
Should fix #101, but I'm not sure if I implemented it correctly.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextAreaBehavior.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextAreaBehavior.java
@@ -169,13 +169,16 @@ class StyledTextAreaBehavior {
                 .orElse(otherNavigation).ifConsumed((b, e) -> b.view.clearTargetCaretOffset())
                 .orElse(verticalNavigation)
                 .orElse(copyAction)
+                .ifConsumed((b, e) -> b.view.requestFollowCaret())
+                // no need to add 'ifConsumed' after charPress since
+                // requestFollowCaret is called in keyTypedTemplate
                 .orElse(charPressConsumer);
 
         InputMapTemplate<StyledTextAreaBehavior, KeyEvent> keyTypedBase = consume(
                 // character input
                 EventPattern.keyTyped().onlyIf(noControlKeys.and(e -> isLegal(e.getCharacter()))),
                 StyledTextAreaBehavior::keyTyped
-        );
+        ).ifConsumed((b, e) -> b.view.requestFollowCaret());
         InputMapTemplate<StyledTextAreaBehavior, ? super KeyEvent> keyTypedTemplate = when(b -> b.view.isEditable(), keyTypedBase);
 
         InputMapTemplate<StyledTextAreaBehavior, ? super MouseEvent> mouseEventTemplate = sequence(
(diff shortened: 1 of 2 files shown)


richtextfx/src/main/java/org/fxmisc/richtext/StyledTextAreaBehavior.java
```diff
@@ -169,13 +169,14 @@
                 .orElse(otherNavigation).ifConsumed((b, e) -> b.view.clearTargetCaretOffset())
                 .orElse(verticalNavigation)
                 .orElse(copyAction)
-                .orElse(charPressConsumer);
+                .orElse(charPressConsumer)
+                .ifConsumed((b, e) -> b.view.requestFollowCaret());
```
TomasMikula: Maybe these last two lines should be swapped? Shouldn't make a change in observable behavior, but when `charPressConsumer` consumes an event, we expect a KEY_TYPED event, and we `requestFollowCaret` there.

JordanMartinez:
Updated in light of your comment.

TomasMikula:
👍

JordanMartinez:
Since that's now a public method... we should add some javadoc to explain
its usage in the method and class javadoc

JordanMartinez:
I amended the commit, but it seems to have shoved the two commits into one. Thus, `requestFollowCaret` isn't actually public in this PR. I'll need to open another one to correct that issue.

Issue #333, opened 2016-06-21
Move UndoManager creation code outside StyledTextArea


JordanMartinez:
Relates to #233. In order to have non-linear undo/redo capability, `StyledTextArea` and its model class will need to take an `UndoManager` argument and push the creation code outside of the area itself.

There's a few things to keep in mind:
1. Since the area (which IMO adheres more to the MVVM paradigm now) could be used in a larger view component, the definition of a valid change can now be different than the normal ones given. Opening up this argument would give more flexibility to the developer to determine what a "change" actually is.
2. With a non-linear undo/redo capabilities, the size of the queue can change between clones. One might have an unlimited queue, another could have a fixed size queue, and a third a zero-size queue. As such, an UndoManagerFactory can no longer be the main argument passed because it doesn't give the developer the freedom to choose which queue size to use. One exception to this is to pass in a `QuadFunction` that accepts the arguments now found in `StyledTextAreaModel#createRich/PlainUndoManager`. However, this might restrict the developer due to the first point made.
3. <del>Since an `EditableStyledDocument` exists outside of the area before being inserted into one then that document could, at least technically, be modified by a non-`StyledTextArea` class. If that occurs, the code should be robust enough to update all the areas displaying its content to account for that change (e.g. a valid undo might no longer be valid). This implies that the ``UndoManagerFactory` should somehow be tied to the `EditableStyledDocument` for the lifetime of that document.</del>
   - I haven't been able to think of a good way to tie the two items together. Perhaps some object that groups the two together could work, but how does the `UndoManagerFactory` know which items to ignore (an area changed the document) and which to accept (the document was modified directly)?
   - At the same time, this possible bug exists even right now: an area's `UndoManager`'s next undo/redo could be invalidated if a developer edited that document directly instead of through the area. Rather than accounting for that possibility (why would that happen in the first place?), it just shouldn't be done.
4. A new filtering mechanism is needed to filter which events an area did and which it did not do to insure that that area's `UndoManager`'s next undo/redo is only the one that the area did and not some other clone's change. UndoFX already uses something like this to ignore changes to the content that it performs:

``` java
private final EditableStyledDocument<?, ?> document = // creation code
private final SuspendableNo performingUpdate = new SuspendableNo();

public final EventStream<TextChange> textChanges() {
    return document.textChanges().conditionOn(performingUpdate);
}

performingUpdate.suspendWhile(() -> replaceText(0, 0, "Some text"));
```

[...]

PR #381, opened 2016-10-14
Include UndoManager in StyledTextArea's constructor's parameter list …

JordanMartinez:
…in preparation for NonlinearUndoManagers
- Source compatibility is maintained
- Since `setUndoManager(UndoManagerFactory)` was removed, the API necessary for adding an UndoManager with alternative history sizes (i.e. fixed size or zero size) was moved from STAModel to STA. Additionally, two convenience methods were added to make installing them just as convenient: `installNew[Plain/Rich]LinearUndoManager(UndoManagerFactory)`.

This addresses #333. Since my PR for nonlinear undos/redos on UndoFX is complicated, this commit would allow me to test it using RichTextFX and a local compiled copy of UndoFX with my commits.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextArea.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/StyledTextArea.java
@@ -8,6 +8,7 @@ import static org.reactfx.util.Tuples.*;
 import java.time.Duration;
 import java.util.Optional;
 import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.IntConsumer;
 import java.util.function.IntFunction;
@@ -233,9 +234,7 @@ public class StyledTextArea<PS, S> extends Region
 
     // undo manager
     @Override public UndoManager getUndoManager() { return model.getUndoManager(); }
-    @Override public void setUndoManager(UndoManagerFactory undoManagerFactory) {
-        model.setUndoManager(undoManagerFactory);
-    }
+    @Override public void setUndoManager(UndoManager undoManager) { model.setUndoManager(undoManager); }
 
     /**
      * Popup window that will be positioned by this text area relative to the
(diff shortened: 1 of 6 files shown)


JordanMartinez:
I'll update this later on to fix those build errors. However, I feel like this shouldn't be merged until the `custom object` PR gets merged as it will be easier for me to update his code to include mine rather than his to include mine.

Also, this PR doesn't address a possible NPE if a user passes in a null argument to `setUndoManager`. Should that be changed to an `Optional<UndoManager>`?

JordanMartinez:
> Also, this PR doesn't address a possible NPE if a user passes in a null argument to setUndoManager. Should that be changed to an Optional<UndoManager>?

I included a check to insure that the default UndoManager is created.

richtextfx/src/main/java/org/fxmisc/richtext/StyledTextArea.java
```diff
@@ -1163,30 +1162,30 @@ public void dispose() {
         virtualFlow.dispose();
     }
 
-    public UndoManager createPlainLinearUndoManager(UndoManagerFactory factory) {
-        return model.createPlainLinearUndoManager(factory);
+    public UndoManager createPlainUndoManager(UndoManagerFactory<PlainTextChange> factory, int capacity) {
```
JordanMartinez: `int capacity` could be changed into a wrapper class (much like your `CaretOffsetX` class to force users to examine the javadoc and see that a value of `-1` means unlimtied, `0` zero, and `n` is the size of the capacity:
````java
public ChangeQueueCapacity {
    public final static ChangeQueueCapacity UNLIMITED = new ChangeQueueCapcity(-1);
    public final static ChangeQueueCapacity ZERO = new ChangeQueueCapcity(0);

    private final int capacity;
    public final int getCapacity() { return capacity; }

    public ChangeQueueCapacity(int capacity) {
        this.capacity = capacity;
    }
}
````

JordanMartinez:
In the latest commit, I'm using a locally compiled version of my [nonlinear UndoFX PR](https://github.com/TomasMikula/UndoFX/pull/12).

JordanMartinez:
Closing due to [my summary of Tomas' comment and its implications](https://github.com/TomasMikula/RichTextFX/issues/333#issuecomment-288479708)

Issue #285, opened 2016-04-07
Inserting caret offset by css padding


terjedahl:
When applying style to StyledTextArea such as: "-fx-padding: 10;", and then clicking in the text to insert the caret, the caret gets inserted 10px to the right of where I clicked - in practice more than one character later than where I tried to insert it.

JordanMartinez:
I wasn't able to reproduce this issue via the example below

````java
import javafx.application.Application;
import javafx.geometry.Insets;
import javafx.scene.Scene;
import javafx.scene.control.ScrollPane;
import javafx.scene.control.TextArea;
import javafx.scene.layout.BorderPane;
import javafx.scene.layout.StackPane;
import javafx.stage.Stage;
import org.fxmisc.richtext.CharacterHit;
import org.fxmisc.richtext.InlineCssTextArea;

public class Moko extends Application {

    public static void main(String[] args) {
        launch(args);
    }

    @Override
    public void start(Stage primaryStage) {
        InlineCssTextArea area = new InlineCssTextArea("One.\nTwo.\nThree.");
        area.setStyle("-fx-padding: 20;");
        Scene scene = new Scene(new StackPane(area), 500, 500);

        primaryStage.setScene(scene);
        primaryStage.show();
    }
}
````

TomasMikula:
Fixed by #396

PR #396, opened 2016-11-17
Fix computing hit character if padding is set on the paragraph-text

JFormDesigner:
This PR fixes computing hit character (at mouse press) if padding is set on the paragraph-text.

To reproduce this, apply a large padding (as below), click at some text in the styled text area and see where the caret is positioned. Without this fix the caret is always some characters right to the click location.

```
.styled-text-area .paragraph-box .paragraph-text {
    -fx-padding: 0 10em 0 10em;
}
.styled-text-area .paragraph-box:first-paragraph .paragraph-text {
    -fx-padding: 10em 10em 0 10em;
}
.styled-text-area .paragraph-box:last-paragraph .paragraph-text {
    -fx-padding: 0 10em 10em 10em;
}
```

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/ParagraphBox.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/ParagraphBox.java
@@ -134,7 +134,8 @@ class ParagraphBox<PS, S> extends Region {
     public CharacterHit hit(double x, double y) {
         Point2D onScreen = this.localToScreen(x, y);
         Point2D inText = text.screenToLocal(onScreen);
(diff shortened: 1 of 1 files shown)


TomasMikula:
Sorry for late response. This looks good to me! 👍 Should also fix #285. I suggest we wait a bit for #356 to be merged, as this one should be easier to rebase.

afester:
@JFormDesigner very cool ! I came across the same issue some days ago, and now its already fixed 👍

TomasMikula:
No merge conflict after all :)

Issue #303, opened 2016-05-01
-fx-highlight-(text)-fill issues?


radicaled:
Sorry for yet another issue, but...

I'm on 1.0.0-SNAPSHOT and it seems that setting `-fx-highlight-fill` and `-fx-highlight-text-fill` aren't working as expected. I'm following the instructions here: https://github.com/TomasMikula/RichTextFX/wiki/Styling-RichTextFX-and-a-CSS-Cheat-Sheet#cheat-sheet

The bottom part of the attached image is from a StyleClassedTextArea. You can see that the text is being styled correctly (it's got a class of "header"), but the highlight CSS isn't.
![no-fill-no-chill](https://cloud.githubusercontent.com/assets/40635/14941744/00020fe6-0f75-11e6-8af5-76e3e95ae254.png)

JordanMartinez:
@radicaled Looking at the source code [here](https://github.com/TomasMikula/RichTextFX/blob/master/richtextfx/src/main/java/org/fxmisc/richtext/ParagraphText.java#L26-28), that's something else that needs to be fixed :-/

JordanMartinez:
I've updated the CSS Guide with a note that highlighting doesn't currently work.

radicaled:
Thanks for looking into it! I'll... figure something out until this gets fixed.

JordanMartinez:
The order of styles is already applied in this order:
1. selection
2. styled text
3. default (style-less) values

The "selection style," if you will, overrides only the background and foreground of the text; all other style-related items (font, font size, weight, etc.) are not overridden. So, here's one way we could implement this:
1. Determine the selection range
2. Determine where the `Paragraph`'s style segments overlap with that range
3. Construct the Text objects
 - if no overlap, just apply the style
 - if overlap covers all of the segment, apply style and selection style
 - if overlap covers a portion of the segment, create multiple Text objects: apply the regular style and selection style to the selected portion and only the regular style to the unselected one(s).

So, given the following example where angle brackets represent the selection:
**Some text here _that <was selected_ and> whatnot**

There would be 5 Text objects:
- **Some text here**
- **_that_**
- **_`was selected`_**
- **`and`**
- **whatnot**

However, you'll notice that in the source code, the Text objects are only created once, not every time `layoutChildren()` is called. So, I don't know if what issues might arise if this approach was taken.

JordanMartinez:
@radicaled Why not consider fixing the issue? It might save you more time overall anyways.

[...]

PR #398, opened 2016-11-17
fixed -fx-highlight-fill and -fx-caret-blink-rate (issue #303)

JFormDesigner:


JordanMartinez:
What is this supposed to do?

kuc:
@JordanMartinez Fixing issue #303 ?

JFormDesigner:
@kuc yes, this fixes #303

I'm also working on a fix for white selection text color (`-fx-highlight-text-fill`). Coming soon...

JordanMartinez:
@JFormDesigner & @kuc Thanks for the clarification. At the same time, how does this fix that issue? Because my original reply to that issue showed that it was a coding issue and I don't see how creating a list of CSS Metadata helps. (I'm not as familiar with CSS as I should be.)

JFormDesigner:
@JordanMartinez `getCssMetaData()` is invoked to ask a node what CSS properties it supports. It is the "connection" between CSS and node properties. IOW without `getCssMetaData()`, CSS attribute `-fx-highlight-fill` is simply ignored, because JavaFX does not know that `StyledTextArea` does support `-fx-highlight-fill`.

JFormDesigner:
BTW this PR does not fix `-fx-highlight-text-fill` (text color of selection),
only `-fx-highlight-fill` (background color of selection).

JordanMartinez:
@JFormDesigner Ah! Ok. I was confused because I felt like the code you implemented would only do something if `StyledTextArea` extended `Control`. That just shows a misunderstanding on my part about `getCssMetaData()` and how it works.

JFormDesigner:
updated (and tested) to current master branch

JordanMartinez:
@JFormDesigner Could you post the code you used to test this? I checked on my side and the background color of the selection did not change to `red`. If I am doing something wrong, let me know. Otherwise, it seems the changes you made, while still needed, don't fix the whole problem.

After updating `styled-text-area.css` to:
````css
.styled-text-area .virtual-flow {
    -fx-cursor: text;
    -fx-highlight-fill: red;
    -fx-font-size: 32pt;
}
````
and fixing the `getCssMetaData` method to what you have in this PR, 
and using this code...
````java
import javafx.application.Application;
import javafx.scene.Scene;
import javafx.stage.Stage;
import org.fxmisc.flowless.VirtualizedScrollPane;
import org.fxmisc.richtext.StyleClassedTextArea;
import org.fxmisc.richtext.StyledTextArea;

import java.util.Collection;
import java.util.Collections;

public class CssTest extends Application {

        launch(args);
    }

    @Override
    public void start(Stage primaryStage) {
        StyledTextArea<String, Collection<String>> area = new StyledTextArea<String, Collection<String>>("", (a, b) -> {},
                Collections.emptyList(), (a, b) -> a.getStyleClass().addAll(b));
        area.replaceText("Some text in the area");
        VirtualizedScrollPane<StyledTextArea<String, Collection<String>>> vsPane = new VirtualizedScrollPane<>(area);

        Scene scene = new Scene(vsPane, 500, 500);
        primaryStage.setScene(scene);
        primaryStage.show();
    }
}
````

System:
````
Linux Mint 18.1

java version "1.8.0_111"
Java(TM) SE Runtime Environment (build 1.8.0_111-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.111-b14, mixed mode)
````

**Comment by JFormDesigner** [comment]
@JordanMartinez it works without `.virtual-flow` in the CSS selector:
```css
.styled-text-area {
    -fx-highlight-fill: red;
    -fx-caret-blink-rate: 100ms;
}
```

BTW I've rebased the PR to current master; this PR also fixes #426

JFormDesigner:
In [Markdown Writer FX](https://github.com/JFormDesigner/markdown-writer-fx) I use a brighter color, which makes selected text better readable but still identifiable as selection:
```css
.styled-text-area {
    -fx-highlight-fill: derive(#1E90FF, 50%);
}
```
Comparison:
image

Maybe you could consider adding this to `styled-text-area.css`?

The optimal solution would be changing selected text color (`-fx-highlight-text-fill`) to `white`, but this does not work because of JDK bug https://bugs.openjdk.java.net/browse/JDK-8149134. See PR #428

JordanMartinez:
Ah, ok! Could you also add a commit that updates the `styled-text-area.css` file to remove the `.virtual-flow` from it?

I'm not sure if there would need to be another area later in that file that writes something like...
````css
.virtual-flow {
    -fx-cursor: text;
}
````

JFormDesigner:
> Could you also add a commit that updates the styled-text-area.css file to remove the .virtual-flow from it?

Sure. Done.

JordanMartinez:
Great! Thanks for your work and patience.

JordanMartinez:
@JFormDesigner I didn't say beforehand, but I like the default color you use when selecting a text as you mentioned [in your above comment](https://github.com/TomasMikula/RichTextFX/pull/398#issuecomment-273098491). 

What if you submitted a PR that updates that in the css file and then update #428 to revert that fix, so that when the JDK bug is fixed, it'll be back to what it was before?

Issue #407, opened 2016-12-20
Underline is hidden by background shape


afester:
With the following CSS

```
-rtfx-background-color: #dcdcdc;
-rtfx-underline-color: blue;
-rtfx-underline-width: 1px;
```

the blue underline is hidden by the gray background shape.
The underline should be in front of the background shape instead. The following hack in ParagraphText.java fixes this:

```
getChildren().add(/*0,*/ underlineShape);
```

by adding the underline shapes to the end of the ParagraphText's child list. Needs some additional verification before submitting a pull request.

TomasMikula:
Fixed by #408

PR #408, opened 2016-12-21
Added background color to spell checker demo. Fixed order of shapes s…

afester:
Fix for issue #407: 
Added background color to spell checker demo.
Fixed order of shapes so that underlines are shown even if a background shape is used.

### DIFF
--- a/richtextfx-demos/src/main/resources/org/fxmisc/richtext/demo/spellchecking.css
+++ b/richtextfx-demos/src/main/resources/org/fxmisc/richtext/demo/spellchecking.css
@@ -1,4 +1,5 @@
 .underlined {
+    -rtfx-background-color: #f0f0f0;
     -rtfx-underline-color: red;
     -rtfx-underline-dash-array: 2 2;
     -rtfx-underline-width: 1;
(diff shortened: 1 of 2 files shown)


JFormDesigner:
@afester the fix has the side effect that the underlines are now painted over the text (e.g. if text contains "gyqp" characters)

I think it would better to use:
~~~java
getChildren().add(1, underlineShape);
~~~

Then the underlines are painted over the background, but under the text.

afester:
@JFormDesigner I know - but is that not the expected behavior? For example, Eclipse paints the spell checking line over the text:

image

and MS Word shows the same behavior.

Anyway, `getChildren().add(1, underlineShape);` would work - the background nodes and the underline nodes for each segment would then be inserted alternately at the beginning of the ParagraphBox (followed by the selection and the caret shape, followed by the segment nodes). (Note: another effect is then that the underline is hidden by a text selection).

JFormDesigner:
@afester OK, sorry, you're right. It's better to have the underlines paint over all other, especially the text selection.

TomasMikula:
👍

PR #415, opened 2017-01-10
Make `positionCaret` package-private

JordanMartinez:
Since this may have value beyond what I know for code editors and the like, I have to ask: why did you originally implement this method? Is making this method package private desirable? AFAIK, we're only doing that because many confused `moveTo`'s functionality for `positionCaret`, since the latter implies that more so than the former.
In my opinion, it would be better to rename `moveTo` to include "caret" in it (`moveCaretTo`? `moveCaret`? `caretTo`?) and rename `positionCaret` to include more explanatory details (`positionCaretDisconnectedly`? `positionCaretSeparately`?).


JordanMartinez:
For now, I'm going to merge this. We can always expose it later on as a better-named method.

Issue #386, opened 2016-10-23
Feature request: Caret line position vs. paragraph position


DaveJarvis:
The methods [lineStart](http://fxmisc.github.io/richtext/javadoc/0.7-M2/org/fxmisc/richtext/model/NavigationActions.html#lineStart-org.fxmisc.richtext.model.NavigationActions.SelectionPolicy-) and [lineEnd](http://fxmisc.github.io/richtext/javadoc/0.7-M2/org/fxmisc/richtext/model/NavigationActions.html#lineEnd-org.fxmisc.richtext.model.NavigationActions.SelectionPolicy-) move the cursor to the start and end of the _paragraph_, not the line. Their names should be changed to `paragraphStart` and `paragraphEnd` to avoid confusion, as well as renaming [selectLine](http://fxmisc.github.io/richtext/javadoc/0.7-M2/org/fxmisc/richtext/model/NavigationActions.html#selectLine--) to `selectParagraph`.

When pressing either `Home` or `End`, the caret position should change relative to the current line, not the current paragraph, which is what users have come to expect (e.g., the text area used to write this message has typical home/end key behaviour). There are a few related issues: #384 and #360.

It would be convenient if there was a `getLineNumber()` method that indicated the line number within the current paragraph, such as:

```
StyleClassedTextArea editor = // ...
Paragraph p = editor.getCurrentParagraph();
int n = p.getLineNumber();
```

Or:

```
StyleClassedTextArea editor = // ...
int n = editor.getCurrentLineNumber(); // delegates to paragraph
```

Once `lineStart` and `lineEnd` provide line-level values (instead of paragraph), it implies that writing a key interceptor for cutting a line (e.g., typing `Ctrl+x` as per IDE behaviour) should be as simple as:

```
editor.lineStart(); // should clear by default
int start = editor.getCaretPosition();
editor.lineEnd(); // should clear by default
int end = textArea.getCaretPosition();
editor.deleteText( start, end );
```

Or, pie-in-the-sky, to avoid extraneous UI updates:

```
int start = editor.getCaretLineStartColumn(); 
int end = editor.getCaretLineEndColumn();
editor.deleteText( start, end );
```

This would also make it trivial to override the default Home/End behaviour to move the cursor to the start and end of lines, rather than paragraphs.

JordanMartinez:
[...]

Issue #222, opened 2015-12-05
Add support for multiple carets and selection ranges


ghost:
Hi Tomas, is it on the radar to support multiple caret positions and selection ranges(with the ability to insert,paste text into those rsnges)? Just noticing that Flowless, and RichTextFX are going through major refactoring, so thought of asking :)

JordanMartinez:
I wonder if it might be wise to implement #152 before doing this...

melkhaldi:
You guys know better!

melkhaldi:
This can enable many things like refactoring code variables, (members?), paste into multiple ranges , etc

JordanMartinez:
Let me explain my thinking:
When removing the skin, `StyledTextArea`'s dispose method was no longer needed, but it will be needed once 152 is implemented. So, since we've already done a lot of refactoring as whole, why not continue it with that part as well? Then, any new changes would be built on top of this refined implementation of the MVC paradigm.

However, I might be saying this because I'm viewing the master branch in terms of a "story" rather than "how it was developed" (I recently read an article that explained people usually view a git history in terms of one of those two ideas, but I can't remember where that article is).

TomasMikula:
I agree with Jordan that continuing the refactoring is more a natural immediate course, because, let's clean up the basics first. That doesn't mean that multiple carets/selections can't be implemented first. But unless someone steps up and does the work (:wink:), it will not happen before.

JordanMartinez:
True... except I'm currently focused on one of the last problems in my program (and then I can release it.... at least, once I figure out how to do that :-D)

At the same time, 152 isn't a high priority right now, just a nice thing to have in my specific use case.

Would you be up for implementing 152 @melkhaldi ? I don't have a need to implement multiple carets or selections as of right now, and Tomas seems to be working on other things. Yes, there would be a higher chance of getting this feature implemented if you did it, but my guess is that this feature relates enough to 152 that there would be reason for you to do that as well. At the same time, you would probably understand the library well enough to implement 152 if you implemented this feature, or vice versa, depending on the order taken.

melkhaldi:
Oh no! You really don't want me near your code. At my best I may be the
hacker type of user who fiddles with methods with lots of patience and
trial and error to find his way.  I once hoped to try making a CodeField
(syntax highlighted text field) ...but that didn't really move... :/ I mean
I tried that a long time ago, but now I feel I got better at this gig! Will
try to go for a deep dive into the docs and see how the library is made
[...]

PR #417, opened 2017-01-10
Add API: get the line index of the caret in a multi-line paragraph/viewport

JordanMartinez:
Implements feature requested in #386:

> It would be convenient if there was a getLineNumber() method that indicated the line number within the current paragraph, such as:

````java
StyleClassedTextArea editor = // ...
Paragraph p = editor.getCurrentParagraph();
int n = p.getLineNumber();
````

> Asking the editor's line number should give the line number from the top of the text area (starting from line 1), not the current paragraph. Also, using natural numbers seems like a good fit, as opposed to whole numbers, for line numbers.

### DIFF
--- a/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/CaretLineIndexDemo.java
+++ b/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/CaretLineIndexDemo.java
@@ -1,0 +1,60 @@
+package org.fxmisc.richtext.demo;
+
+import javafx.application.Application;
+import javafx.scene.Scene;
+import javafx.scene.control.Label;
+import javafx.scene.layout.BorderPane;
+import javafx.scene.layout.VBox;
+import javafx.stage.Stage;
+import org.fxmisc.flowless.VirtualizedScrollPane;
+import org.fxmisc.richtext.InlineCssTextArea;
+import org.fxmisc.richtext.LineNumberFactory;
+import org.reactfx.value.Val;
+
+import java.util.Objects;
+
+public class CaretLineIndexDemo extends Application {
+    @Override
+    public void start(Stage primaryStage) throws Exception {
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < 15; i++) {  // lines
+            for (int j = 0; j < 100; j++) { // content on each line
+                sb.append(j).append(" ");
+            }
+            sb.append("\n");
+        }
+        InlineCssTextArea area = new InlineCssTextArea(sb.toString());
+        area.setWrapText(true);
+        area.setParagraphGraphicFactory(LineNumberFactory.get(area));
+        VirtualizedScrollPane<InlineCssTextArea> vspane = new VirtualizedScrollPane<>(area);
+
+        Label inParagraphLabel = new Label("");
+        Val.create(area::getCaretLineIndexInParagraph, area.caretPositionProperty())
+                .values()
+                .filter(Objects::nonNull)
+                .map(v -> "Line Index in Paragraph is: " + String.valueOf(v))
+                .feedTo(inParagraphLabel.textProperty());
+
+        Label inViewportLabel = new Label("");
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by JordanMartinez** [comment]
This works in all cases but when the caret is on the last line in a multi-line paragraph. When calling `lineEnd`, it will move the caret to the second-to-last position.

When I accounted for this, it then screwed up caret navigation via the arrows keys in rare cases.

**Comment by JordanMartinez** [comment]
@DaveJarvis In this PR, the method `getCaretLineIndexInParagraph` makes sense to me, but what about the second one? Looking at previous comments in other unrelated issues, it seems you wanted the second method so that you can have more accurate scrolling capabilities. However, I don't know how much such a method will help in that case. What is your use case?

**Comment by DaveJarvis** [comment]
I've changed the way scrolling is implemented.

The problem I had was this: Given a plain Markdown, R Markdown, XML, or R XML file to edit, ensure that the HTML preview of the edited content is [scrolled](https://github.com/DaveJarvis/scrivenvar/blob/master/src/main/java/com/scrivenvar/preview/HTMLPreviewPane.java#L130) such that the cursor position is reflected. The [editor](https://github.com/DaveJarvis/scrivenvar/blob/master/src/main/java/com/scrivenvar/editors/EditorPane.java#L55) is now [completely decoupled](https://github.com/DaveJarvis/scrivenvar/blob/master/src/main/java/com/scrivenvar/processors/ProcessorFactory.java#L72) from the preview pane so that processors can [inject](https://github.com/DaveJarvis/scrivenvar/blob/master/src/main/java/com/scrivenvar/processors/CaretInsertionProcessor.java#L65) a hidden element within the text at the caret offset during processing. In other words, a token is passed along from editor, through various links of a processor chain (chain-of-responsibility pattern), and written as a `<span>` element in the preview, which is used to synchronize the preview with the editor's caret offset.

The only issue of note now is #386, which will allow efficient editing by enabling the use of Home/End keys to perform the expected behaviour.

The name `getCaretLineIndexInParagraph()` seems a bit clunky and counter-intuitive. Here's what seems natural to me:

```
// Get a handle to the paragraph that contains the caret.
Paragraph p = editor.getCaretParagraph();

// Throws InvalidCaretParagraphException if the paragraph doesn't contain the caret.
int line = p.getCaretLineNumber();

// Returns the number of lines in the paragraph (unused in this example, but useful).
int tally = p.getLineCount();

// Move the caret to the start of the line; i.e., Home key
editor.moveTo( p, line, 0 );

// Find the number of columns for the caret's line.
int column = p.getColumnCount( line ):

// Move the caret to the end of the line; i.e., the End key
editor.moveTo( p, line, column );
```

I don't know if this is possible, but the code could be simplified to:

```
editor.moveTo( CaretPosition.END_OF_LINE );
editor.moveTo( CaretPosition.START_OF_LINE );
```

Or simpler still:

```
// End of the caret's current line
editor.moveToEnd();

// Start of the caret's current line
editor.moveToStart();
```

Or a bit more generically, re-using the aforementioned `column` variable:

```
// Jump to the end of the current line of the caret's paragraph.
editor.moveTo( line, column );

// Jump to the beginning of the current line of the caret's paragraph.
editor.moveTo( line, 0 );
```

JordanMartinez:
You mean... [`lineStart` and `lineEnd`](https://github.com/TomasMikula/RichTextFX/commit/1554a514e2e8592293c42cb20e753877fdeaab29), which is now [the default behavior](https://github.com/TomasMikula/RichTextFX/commit/1855384e5ddf1e6afa912ce66f2b761dcb7a131f)? :smile:

DaveJarvis:
That's fantastic! Am greatly anticipating the `0.7-M4`.

I still think a `Caret` object is a good idea, which could eliminate a decent amount of complexity in client code.

JordanMartinez:
To be honest, a `caret` object would bring more OOP to that area in the code.... It might also help implement #222.... Hmm... I'm going to open a new issue so we can discuss that further.

Issue #384, opened 2016-10-18
Support for  numbered and unnumbered Lists


afester:
It should be possible to define numbered and unnumbered, nested lists. Each paragraph in the list would serve as a list item, the paragraph would be indented according to its level inside the nested list, and optionally prepended either with a level specific bullet (ideally defineable by CSS) or a numeric value, either as single numbers or as combined numbers (like 1., 1.1, 1.2, 1.2,1 etc).

This requires enhancements in the data model (optionally allowing to add each paragraph to a List) and in the rendering layer (indenting ParagraphText and adding the bullet node).
It also requires some adjustments in the behaviour of the component: When pressing enter inside a paragraph which is part of a List, the new Paragraph will become a list item of the same List.

JordanMartinez:
How should this be implemented?

Would we be generifying the Paragraph similarly to how we generified the Segment? In which case, rather than a list of `Paragraph`s, we would have a list of generic lines. A line could be a `List` that could contain another `List` (for nested lists) or a `Paragraph` (for the actual item itself). For example
````java
public class StyledTextArea<Line, LS, Seg, S> {
    // rather than "getParagraphs()" this would now be
    public List<Line> getLines();
}

// thus generifying the Paragraph to be
StyledTextArea<
    // paragraph is now lines
    Either<NestedList, Paragraph<String>>, String, 
    // your custom object support
    Either<StyledText<String>, Image>, String>
> area = //;
````

afester:
> Would we be generifying the Paragraph similarly to how we generified the Segment?

Ah no, I am still busy with unscrambling the knots in my head, caused by all those generic types from the `Segment` implementation :joy:

My current thinking is that a Paragraph must not be as generic as a Segment - and a Paragraph's list structure is more tied to the Paragraph's style than to its type. The current approach is to add an `Optional<ListItem>` to a `Paragraph`, which can be passed around when creating new derived Paragraphs (when splitting, joining etc.). The `ListItem` can then further reference to some List object which further defines the structure of the List (this is especially required for hierarchically numbered Lists). But, let's discuss any other reasonable approach ...

My playground for this enhancement is https://github.com/afester/RichTextFX/tree/paragraphBullets. It already implements indentation of single paragraphs and rendering for level dependent bullets:

![image](https://cloud.githubusercontent.com/assets/2681864/19642715/d2aad9c2-99e6-11e6-83aa-f2afce7264ab.png)

[...]

PR #419, opened 2017-01-10
Update method & behavior: lineStart/End, not paragraphStart/End

JordanMartinez:
> The methods lineStart and lineEnd move the cursor to the start and end of the paragraph, not the line. Their names should be changed to paragraphStart and paragraphEnd to avoid confusion, as well as renaming selectLine to selectParagraph.

> When pressing either Home or End, the caret position should change relative to the current line, not the current paragraph, which is what users have come to expect (e.g., the text area used to write this message has typical home/end key behaviour). There are a few related issues: #384 and #360.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
@@ -1041,6 +1041,36 @@ public class GenericStyledArea<PS, SEG, S> extends Region
     }
 
     /**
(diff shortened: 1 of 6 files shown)


JordanMartinez:
Apparently, I wrote this in the wrong PR:
> This works in all cases except when the caret is on the last line in a multi-line paragraph. When calling `lineEnd`, it will move the caret to the second-to-last position.

> When I accounted for this, it then screwed up caret navigation via the arrows keys in rare cases.

However, it turns out this was an unrelated bug. See #423.

PR #402, opened 2016-12-06
fixed IllegalArgumentException on undo

grigoriliev:
I changed the TextChange.mergeWith method so that the undo merging mimics the behavior of a regular text editor. This also fixes #322 (java.lang.IllegalArgumentException: Unexpected change received).

PR #424, opened 2017-01-17
Add dependency for TestFX tests

JordanMartinez:


JordanMartinez:
I'm merging this now so that others who have implemented code (namely, #402) can rebase onto the current master and starting adding TestFX tests to their PRs.

PR #428, opened 2017-01-17
Fix -fx-highlight-text-fill (selected text color) (issue #303)

JFormDesigner:
**Note**: this PR should be **not yet merged into master** because it does not work correctly as long as the JDK bug (see below) is not fixed. Created the PR to avoid that the fix is lost...

This PR attempts to fix text color of selection (white selected text), BUT:
Because of JDK bug https://bugs.openjdk.java.net/browse/JDK-8149134
this does not work correctly if a paragraph contains more than one segment
and the selection is (also) in the second or later segments.
Visually the text color of the selection may be mix black & white.
Sometimes text becomes invisible (white text on white background).

Mixed text selection color:

image

Sometimes makes text invisible:

image


JordanMartinez:
Thanks for filing this @JFormDesigner

PR #434, opened 2017-01-22
Line to Paragraph

DaveJarvis:
Public API documentation was missed.


JordanMartinez:
@DaveJarvis Thanks for catching that.
@afester Thanks for merging this as I have been busy these past few days.

Issue #433, opened 2017-01-21
Caret Position = InlineCssTextArea .getText().length() but it doesn't scroll vertically to the bottom


goxr3plus:
-  Java Version: **jdk1.8.0_112**  

- RichTextFX Version: **richtextfx-fat-0.6.10.jar**

I am using :         
 ```
InlineCssTextArea textArea = new InlineCssTextArea();
textArea.setEditable(false);
textArea.setFocusTraversable(false);
```

and i am trying to scroll programmatically at the bottom  :
 ```
//It returns the maximum length of the InlineCssTextArea  `text`
 System.out.println(textArea.getText().length());
 System.out.println(textArea.getCaretPosition());
			    
 //It doesn't scroll to the end
 textArea.positionCaret(textArea.getText().length());
```

The problem is that the length of `InlineCssTextArea `text is `830`  even if the caret is positioned at `830` it doesn't scroll to the bottom...

![screenshot49984](https://cloud.githubusercontent.com/assets/20374208/22170011/b528c4d2-df81-11e6-8dad-0f4b58f60a1a.png)

Is it a bug?

JordanMartinez:
Are you using `positionCaret` intentionally? The right way to move the caret is via `moveTo` (I don't know if that will fix your problem).

goxr3plus:
Hello Jordan Martinez , thanks for fast reply .

I have tried your recommendation with different ways but again that Vertical scroll bar position doesn't seem to move down . Is there any other solution?

Best Regards.

JordanMartinez:
@goxr3plus Can you provide a simple Application that reproduces the bug? I could probably help troubleshoot what's going on later on today as I'm not as familiar with the `0.6.10` release since it was released so long ago.

goxr3plus:
@JordanMartinez 

Below is a simple working example showing the problem . I had run it on Window 10:

I will be very glad to download the newer version of RichTextFX but the links seem to be dead , or those who working are giving as option the version `0.6.10 ` . The link for the newer repositories will be gold :) .
```
import java.util.Arrays;

import org.fxmisc.richtext.InlineCssTextArea;

import javafx.application.Application;
import javafx.scene.Scene;
import javafx.scene.layout.GridPane;
import javafx.scene.layout.Priority;
import javafx.stage.Stage;

/**
 * @author Alexander
 *
 */
public class Tester extends Application {

    @Override
    public void start(Stage primaryStage) throws Exception {

[...]

PR #436, opened 2017-01-25
Fix dead link to manual download of 0.7-M3 release

JordanMartinez:
As discovered in #433.


JordanMartinez:
Note: the build failed on this because I merged the code before the build finished. The next commit #437 shows that this wasn't an issue.

Issue #363, opened 2016-09-19
[0.7] Contextmenu not supported by : org.fxmisc.richtext.CodeArea


pierrecochet:
Hi,
Since i use the 0.7 the property ContextMenu isn't recognized, whereas it works on 0.6. Do i have any way to fix it, or any altrenative ?

JordanMartinez:
Just a guess, but I think this is because `StyledTextArea` does not extend `Control` but `Region` in the new release....

rivagti:
@JordanMartinez yeah, you're right. but any idea how to implement it?

nazmuddin:
@JordanMartinez  , I did it like this in `GenericStyledArea`

`private Popup popup;
    private Button cut;
    private Button copy;
    private Button paste;
    private Button select;

    private EventHandler contextMenu = new EventHandler<MouseEvent>() {
        @Override
        public void handle(MouseEvent event) {
            if (event.getButton() == event.getButton().SECONDARY) {
                popup.setAutoFix(false);
                popup.show(area, event.getScreenX(), event.getScreenY());
            } else if (popup.isShowing() && event.getClickCount() == 1) {
                popup.hide();
            }
        }
    };
    
    public RichTextFx() {
        StageFactory fact = StageFactory.INSTANCE;
        mainStage = fact.createStage();

        popup = new Popup();
        cut = new Button("Cut");
        cut.setOnAction(e ->{
            area.cut();
            popup.hide();
        });
        copy = new Button("Copy");
        copy.setOnAction(e -> {
            area.copy();
            popup.hide();
        });
        paste = new Button("Paste");
        paste.setOnAction(e ->{
            area.paste();
            popup.hide();
        });
        select = new Button("Select All");
        select.setOnAction(e->{
            area.selectAll();
        });
        VBox box = new VBox();
        box.setId("popup"); 
        box.getStylesheets().add(getClass().getResource("context-menu.css").toExternalForm());
        box.setPrefSize(80, 50);
        box.getChildren().addAll(cut, copy, paste, select);

        popup.getContent().add(box);
        area.setOnMouseClicked(contextMenu);
   }
`
area is instance of GenericStyledArea. Hope it helps.

JordanMartinez:
@rivagti See @nazmuddin's example in combination with mine below. He uses a `Popup` whereas I use a `ContextMenu`. He uses `setOn[EventType]` whereas I use `InputMap`. The two should give you a good enough idea for how to resolve this issue:
````java
[...]

PR #439, opened 2017-01-26
Add context menu

JordanMartinez:
This addresses #363.

There is no `ContextMenu` by default: this matches the same behavior as all other nodes.

### Questions
- Should the default offset values be `5` or some positive non-zero value? In the TestFX test, having it as zero will fail the tests because the mouse clicks on the menu rather than the area once it is shown. Developer who set the menu in their code will also need to override the default values every time, which is unnecessary boilerplate.
    - **Edit:** I've updated the PR to set the default offset values to `2`. There's no specific reason for this choice, so it's up for modification.

### DIFF
--- a/richtextfx/build.gradle
+++ b/richtextfx/build.gradle
@@ -18,6 +18,7 @@ dependencies {
     testCompile ("org.testfx:testfx-junit:4.0.5-alpha") {
         exclude(group: "junit", module: "junit")
     }
+    testCompile group: 'com.nitorcreations', name: 'junit-runners', version: '1.2'
 }
 
 javadoc {
--- a/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
@@ -41,6 +41,7 @@ import javafx.geometry.Bounds;
 import javafx.geometry.Insets;
 import javafx.geometry.Point2D;
 import javafx.scene.Node;
+import javafx.scene.control.ContextMenu;
 import javafx.scene.control.IndexRange;
 import javafx.scene.layout.Background;
 import javafx.scene.layout.BackgroundFill;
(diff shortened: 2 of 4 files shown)


JFormDesigner:
IMHO it would be better to use `ContextMenuEvent.CONTEXT_MENU_REQUESTED`, instead of `mouseClicked(MouseButton.SECONDARY)`, to show the context menu. Then the user can also use the keyboard to show the context menu.

`javafx.scene.control.Control` also uses `ContextMenuEvent.CONTEXT_MENU_REQUESTED`.

JordanMartinez:
Ah! Good point! I assumed it was just a right-click.

JordanMartinez:
I've included @JFormDesigner's point about the `ContextMenuEvent`.

JordanMartinez:
Since I haven't heard anything else about the default offset values, I'm going to merge this. They can always be modified later on if need be.

Groostav:
If I wanted a context menu containing the usual `copy`, `paste`, `undo`, `redo`, etc, whats the easiest way to get it? 

RichTextFX exposes the functionality of a text-field with methods simply named `copy()`, `paste()`, etc, which are pretty clearly meant for this purpose.

But:

1. theres no `delete()` method
2. the `undo` and `redo` _disable if no previous state_ and _disable if no next state_ bindings are non-trivial
3. the `cut` and `copy` _disable if no text selected_ is non-trivial,
4. anything else im missing? icons?

I'm wondering if you would be kind enough to write one, and then let us do something like create a static method `createDefaultContextMenu`, such that I could write

```java
styleClassedTextArea.contextMenu = StyleClassTextArea.makeDefaultContextMenu()
```

or even

```xml
<StyleClassedTextArea>
  <contextMenu>
    <StyleClassedTextArea fx:factory="makeDefaultContextMenu"/>
  </contextMenu>
</StyleClassedTextArea>
```

Jugen:
@Groostav 
> I'm wondering if you would be kind enough to write one ....

Here you go, modify and style as needed, to use do: `area.setContextMenu(new DefaultContextMenu())`

```java
private class DefaultContextMenu extends ContextMenu
{
    private MenuItem cut, copy, delete, paste, undo, redo;
    private GenericStyledArea area;
    private IndexRange range;
    
    public DefaultContextMenu()
    {
        showingProperty().addListener( (ob,ov,showing) -> checkMenuItems( showing ) );

        cut = new MenuItem( "Cut" );
        cut.setOnAction( AE -> { hide(); area.cut(); } );

        copy = new MenuItem( "Copy" );
        copy.setOnAction( AE -> { hide(); area.copy(); } );

        delete = new MenuItem( "Delete" );
        delete.setOnAction( AE -> { hide(); area.deleteText( range ); } );

        paste = new MenuItem( "Paste" );
        paste.setOnAction( AE -> { hide(); area.paste(); } );

        redo = new MenuItem( "Redo" );
        redo.setOnAction( AE -> { hide(); area.redo(); } );

        undo = new MenuItem( "Undo" );
        undo.setOnAction( AE -> { hide(); area.undo(); } );

        getItems().addAll( copy, cut, delete, paste, redo, undo );
    }
    
    private void checkMenuItems( boolean showing )
    {
        if ( ! showing ) return;

        area = (GenericStyledArea) getOwnerNode();
        UndoManager history = area.getUndoManager();
        undo.setDisable( ! history.isUndoAvailable() );
        redo.setDisable( ! history.isRedoAvailable() );

        range = area.getSelection();
        boolean noSelection = range.getLength() == 0;
        delete.setDisable( noSelection );
        copy.setDisable( noSelection );
        cut.setDisable( noSelection );
    }
}
```

PR #445, opened 2017-03-03
Update flowless; Exclude transitive ReactFX dependency explicitly

JordanMartinez:


TomasMikula:
Did the transitive dependency cause problems?

JordanMartinez:
I'm not sure. It did when we used a newer version of ReactFX before in RichTextFX but not in Flowless or UndoFX. It might not now, but I haven't tested it without it.

JordanMartinez:
Regardless, I thought it would be better to exclude ReactFX specifically rather than using the global "exclude all transitive dependencies" option in case one of those dependencies should add another dependency.

TomasMikula:
Isn't it a feature to get an error when there are incompatible versions of a dependency?

JordanMartinez:
uh..... I don't know....?

TomasMikula:
I find it better than at runtime. And when that happens, you can _then_ choose to manually override the version of the dependency, on a case by case basis.

JordanMartinez:
> I find it better than at runtime. And when that happens, you can then choose to manually override the version of the dependency, on a case by case basis.

Is there an article that explains this more? I'd hate to waste your time by asking you to explain it further.

Also see #446

TomasMikula:
My reasoning is that most of the time the build with transitive dependencies will be fine. In the rare case of conflicting dependency versions, I find it better to know about it. In case the transitive dependency is newer than the direct dependency, we would probably want RichTextFX to move to the newer version as well, rather than silently override the dependency version to the older one (which might cause problems for the library that asks for the newer version).

JordanMartinez:
Ah ok. Now I see what you mean. In my own words, you want to know when a transitive dependency has a version conflict so we know how to resolve that conflict: if the transitive dependency is an older version, then we should update it asap (since we only use your projects as dependencies), and if it's a newer one, it simply informs us to also update RichTextFX's version. 

ReactFX was excluded at first because we were using a snapshot version (or something like that). I think we (or I) forgot to update it to a stable release once that was resolved.

TomasMikula:
Yes, and if we had a dependency that we don't control that requires an older version of a transitive dependency, then we could choose to _temporarily_ exclude that transitive dependency, but update the build as soon as the versions are in sync again.

JordanMartinez:
Shouldn't this guideline be added to `Contributing.md`?

TomasMikula:
Naah, let's not overload contributors with too much bureaucracy :)

JordanMartinez:
I see your point. On the other hand, you could just assign this responsibility to me. That wouldn't overload other (or future) contributors and would still insure it gets done. If I should stop contributing to this project, you'd be in the same place as we are now if you decide not to do so. The pros outweigh the cons in my mind, but I'll leave that up to you to decide.

Issue #150, opened 2015-06-10
Get BoundingBox of letters in StyledTextArea


kushal256:
Is it possible to get BoundingBox of letters within a StyledTextArea?  Looks like this would require access to package private related classes like StyledTextAreaView and VirtualFlow?  The implementation might be very similar to StyledTextAreaView::getSelectionBoundsOnScreen(), but not really sure?

My end goal was to build graphical connections (like arrows) between two words that each live in a different StyledTextArea's.  Is something like this possible?

TomasMikula:
You are correct, it would require some reflection hacks and the implementation would be similar to `getSelectionBoundsOnScreen()`.

This is another example of a feature that is hard to support in the current unfortunate Control/Skin architecture, where the Control (i.e. the text area) is supposed to be the _model_ in the MVC terminology and Skin is the _view_. Only Control (i.e. model) API is exposed to the user, so there is no direct access to the Skin. I started a [discussion](http://mail.openjdk.java.net/pipermail/openjfx-dev/2015-March/016827.html) about this on the JavaFX mailing list, but nothing really has come out of it.

My plan is to switch to not using skins in RichTextFX at all whenever I find the time to do the refactoring.

Even then, though, it seems to me that what you would want is more like getting an `ObservableValue<BoundingBox>`, i.e. something that updates as the text area is edited or scrolled. That would be a whole other story (though still having the prerequisite of getting rid of the unfortunate skin architecture).

JordanMartinez:
This is another issue that you should be able to implement now that the skin has been removed in 3f00de3c9cc1e85b3660d8f846b26aa85153eb4f.

kushal256:
Awesome, thanks!

kushal256:
This definitely helps!  

As Tomas noted, there are still parts that are a whole other story to get this done.  I hacked away and made some changes and it partially works.  If there is any interest, I can submit via a pull request (would love to get feedback and suggestions for improvement!), but here's the gist of what I hacked away.

Added to my subclass of CodeArea:
 private void setupCoordMapping() {
        // //////////////
        // 1. Collect GUI components we need to listen to changes for
        // Worked in 0.6.10, need to adjust
        // Tied to VirtualFlow ctor last line: getChildren().addAll(content, hbar, vbar);
        // final ScrollBar hbar = (ScrollBar) getVirtualFlow().getChildrenUnmodifiable().get(1);
[...]

PR #455, opened 2017-03-17
Add feature: get character bounds on screen

JordanMartinez:
A few questions I have:

- Should `selectionShape` be used to get the bounds? Will that switch-out approach cause any issues later on? I didn't want to create a new `Path` object each time the method is called so as to reduce memory usage.
- I used the `FORWARD` bias in `endPosition = offsetToPosition(to, Bias.FORWARD)`. Is that correct? Or should it be `BACKWARD`? I still don't understand how that part of the code works.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
@@ -1392,6 +1392,72 @@ public class GenericStyledArea<PS, SEG, S> extends Region
         });
     }
 
+    /**
+     * Gets the character bounds on screen
+     *
+     * @param from the start position
+     * @param to the end position
+     * @return the bounds or {@link Optional#empty()} if line is not visible
+     */
+    public Optional<Bounds> getCharacterBoundsOnScreen(int from, int to) {
+        if (from < 0) {
+            throw new IllegalArgumentException("From is negative: " + from);
+        }
+        if (from > to) {
+            throw new IllegalArgumentException(String.format("From is greater than to. from=%s to=%s", from, to));
+        }
+        if (to > getLength()) {
+            throw new IllegalArgumentException(String.format("To is greater than area's length. length=%s, to=%s", getLength(), to));
+        }
+
+        // no bounds exist if range is just a newline character
+        if (getText(from, to).equals("\n")) {
+            return Optional.empty();
+        }
+
+        // if 'from' is the newline character at the end of a multi-line paragraph, it returns a Bounds that whose
+        //  minX & minY are the minX and minY of the paragraph itself, not the newline character. So, ignore it.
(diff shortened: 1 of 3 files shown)


JordanMartinez:
Addresses #150.

JordanMartinez:
Manually tested this code (well, how would you test it via TestFX...?) with:
````java
import javafx.application.Application;
import javafx.geometry.Bounds;
import javafx.scene.Scene;
import javafx.scene.control.Button;
import javafx.scene.control.TextField;
import javafx.scene.layout.BorderPane;
import javafx.scene.layout.HBox;
import javafx.scene.layout.Region;
import javafx.stage.Popup;
import javafx.stage.Stage;
import org.fxmisc.flowless.VirtualizedScrollPane;
import org.fxmisc.richtext.InlineCssTextArea;

import java.util.Optional;

public class CharacterBoundsDemo extends Application {

    @Override
    public void start(Stage primaryStage) throws Exception {
        InlineCssTextArea area = new InlineCssTextArea("0123456789\n123456789\n123456789");
        area.setWrapText(true);
        area.setStyle("-fx-font-size: 32pt;");
        VirtualizedScrollPane<InlineCssTextArea> vsPane = new VirtualizedScrollPane<>(area);

        Region r = new Region();
        r.setStyle("-fx-background-color: red;");
        Popup p = new Popup();
        p.getContent().add(r);

        TextField fromField = new TextField("0");
        TextField toField = new TextField("10");

        Button b = new Button("Show");
        b.setOnAction(ae -> {
            int from = Integer.parseInt(fromField.getText());
            int to = Integer.parseInt(toField.getText());
            Optional<Bounds> optB = area.getCharacterBoundsOnScreen(from, to);
            optB.ifPresent(bounds -> {
                r.setMinSize(bounds.getWidth(), bounds.getHeight());
                p.show(primaryStage, bounds.getMinX(), bounds.getMinY());
            });
        });
        BorderPane root = new BorderPane();
        root.setCenter(vsPane);
        root.setBottom(new HBox(fromField, toField, b));

        primaryStage.setScene(new Scene(root, 200, 400));
        primaryStage.show();
    }
}
````

JordanMartinez:
> Should `selectionShape` be used to get the bounds? Will that switch-out approach cause any issues later on? I didn't want to create a new `Path` object each time the method is called so as to reduce memory usage.

I don't see this as a problem because it'll only run on the JavaFX Application Thread, so there shouldn't be any issue with concurrency-related things. Additionally, `selectionShape` is already bound to `leftInsets` and `topInsets` which does affect the accuracy of the returned `Bounds` object.

> I used the FORWARD bias in endPosition = offsetToPosition(to, Bias.FORWARD). Is that correct? Or should it be BACKWARD? I still don't understand how that part of the code works.

After testing this out, there wasn't any difference between `FORWARD` and `BACKWARD` even across multi-line paragraphs or multiple paragraphs. I think that only matters for some other thing `Position`-related thing, though Tomas could better explain that.

PR #463, opened 2017-03-23
Remove middleman: ESD is the model now, not STAModel

JordanMartinez:
When the skin was removed, STAModel was created to separate model concerns from view concerns. However, this was before EditableStyledDocument became a public interface. Now that ESD is a public interface, STAModel does nothing more than delegate method calls to that object and handle positions of things within the document (caretPosition, caretColumn, selectedText, etc.)

As of now, this build currently fails because STAModelTest can not create a STAModel. This can be resolved by using TestFX to create an area and then run the tests.
Even so, since #333 should be done, these tests shouldn't matter.

However, I believe more of the model-related things should still be stored in EditableStyledDocument (e.g., the `SuspendableVal`s for various things [text, paragraphs, etc.]) as these values only need to be stored in one object, not stored in each view. I still need to do this.


JordanMartinez:
The two issues above have now been addressed

PR #464, opened 2017-03-23
Cleanup interfaces

JordanMartinez:
I'm not sure if every method is where it should be, but this would better clarify various things. Additionally, since `TextEditingArea` wouldn't refer to just the model component anymore, it could account for view-specific API.


JordanMartinez:
I'd appreciate feedback on this before I merge this.

JordanMartinez:
I'm not sure if this is the best approach to take / best way to modularize the API now that more view-related API is being exposed. Additionally, some of the "model" API (like `selectLine()` relies upon "view" API to work. This needs further discussion.

PR #470, opened 2017-03-25
Fix move selected text bug

### LINKED ISSUE #469: `MoveSelectedText` throw IllegalArgumentException due to type in TextChange
Due to the `TextChange` type introduced in #458, `moveSelectedText` now throws an exception when the area's selection is empty.

JordanMartinez:
Fixes #469.


JordanMartinez:
That last commit statement should be "if nothing is selected" not "is nothing is selected." Oh well.

Issue #472, opened 2017-03-28
[0.7-M4] Inserting text via user keyboard input inserts text in reverse order because caret position doesn't advance forward after input


DaveJarvis:
**Replicate**
1. Run the Java Keywords demo
1. Start typing

**Expected Results**
The caret advances with each character typed.

**Actual Results**
The caret retains its position, which causes the typed text to be entered backwards.

![java-keywords-demo](https://cloud.githubusercontent.com/assets/2131950/24387596/d8c890d4-132a-11e7-8c7d-22645f98c2fd.png)

JordanMartinez:
I'm guessing you're using the new 0.7-M4 release? That's probably a bug from #468. :-/

JordanMartinez:
Nope. I was wrong. The bug start showing up in #463. After I moved the model-API to ESD, every time one types a letter, it moves the caret to the right twice so that the caret is one character farther than it should be. After I moved the view API to the area, the behavior reverted to what @DaveJarvis just now described in this issue.

DaveJarvis:
Correct, this is in the 0.7-M4 release. 0.7-M3 worked as expected.

JordanMartinez:
I've updated the release notes to note this issue.

JordanMartinez:
@TomasMikula Would it be possible to re-release the 0.7-M4 release using the updated master branch that accounts for this issue? Or would this require another release?

JordanMartinez:
Just fixed it. Dave, could you try it out to insure there are no other
regressions?

Although I already tagged you in the issue, is it possible, Tomas, to
re-release the 0.7-M4 release with the current master branch? Or does this
require a new release?

Blessings,
Jordan

DaveJarvis:
@JordanMartinez, once the new release is available, I'll try it out.

I'd suggest making a new release, 0.7-M5. (Mostly to avoid caching issues.)

TomasMikula:
I pushed a new SNAPSHOT release, can you try that out?

I think it used to be possible to override a release, but yeah, there can be caching issues.

JordanMartinez:
:-/ Dang...

DaveJarvis:
SNAPSHOT release has resolved this issue.

JordanMartinez:
Can a new milestone release be made then?

@TomasMikula Besides the bugs and dependency bugs tagged in this project's list of issues, is there anything else that needs to be done before a stable `0.7` release can be made?

[...]

PR #473, opened 2017-03-28
Fix regression: Update caret position after replace call

JordanMartinez:
Turns out I didn't copy this code over into the view correctly because it mixed model and view API calls together


JordanMartinez:
Fixes #472

JordanMartinez:
This fix also insures a `ESD#replace` call doesn't occur unless an actual change will occur by it. Otherwise, it would throw an `IllegalStateException` when trying to construct a `TextChange` that doesn't insert/delete/replace anything.

PR #482, opened 2017-04-06
Added some documentation to some small classes

twistedsquare:
I tried to document everything where I was confident about how it worked, so a few methods in these classes are still uncommented.  But thought it might prove useful in improving the Javadoc a bit, even if some of the content is already on the wiki.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/TextExt.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/TextExt.java
@@ -18,9 +18,16 @@ import javafx.scene.shape.StrokeLineCap;
 import javafx.scene.text.Text;
 
 /**
- * A class which adds some more styleable properties to JavaFX's Text class.
+ * A class which adds some more styleable properties to JavaFX's {@link Text} class.
  *
- * They can be styled using the properties (and accessors/mutators) or via CSS.
+ * The extra items can be styled using the properties (and accessors/mutators) or via CSS.
+ * Each property is documented with its CSS property.  Each CSS property begins with the "-rtfx"
+ * prefix.
+ *
+ * <p>Note that the underline properties specified here are orthogonal to the {@link #underlineProperty()} inherited
+ * from {@link Text}.  The underline properties defined here in {@link TextExt} will cause an underline to be
+ * drawn if {@link #underlineWidthProperty()} is non-null and greater than zero, regardless of
+ * the value of {@link #underlineProperty()}.</p>
  */
 public class TextExt extends Text {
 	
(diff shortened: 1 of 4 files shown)


JordanMartinez:
Thank you for the PR!
I'll take a closer look at this later on in the week as I don't have time right now.

richtextfx/src/main/java/org/fxmisc/richtext/model/Paragraph.java
```diff
@@ -13,6 +13,28 @@
...
+ * {@code StyledText<S>} for the text part and add another segment type for images).
+ * <b>Note that Paragraph is an immutable class</b> - to modify a Paragraph, a new
+ * Paragraph object must be created. Paragraph itself contains some methods which
+ * take care of this, such as concat(), which appends some Paragraph to the current
+ * one and returns a new Paragraph.</p>
+ *
+ * @param <PS> The type of the paragraph style.
+ * @param <SEG> The type of the content segments in the paragraph (e.g. {@code StyledText<S>})
+ * @param <S> The type of the style of individual segments.
+ */
```
JordanMartinez: This looks good. Perhaps there should also be a note that explains that constructing a `Paragraph` must always have at least one SEG object, even if it is an empty one, to insure the code works properly.

richtextfx/src/main/java/org/fxmisc/richtext/TextExt.java
```diff
@@ -17,6 +17,11 @@
 import javafx.scene.shape.StrokeLineCap;
 import javafx.scene.text.Text;
 
+/**
+ * A class which adds some more styleable properties to JavaFX's Text class.
+ *
+ * They can be styled using the properties (and accessors/mutators) or via CSS.
+ */
```
JordanMartinez: Perhaps there should be a note here that explains that the added-on CSS properties all start with the "-rtfx" prefix.

JordanMartinez: Moreover, I believe the added-on underline CSS properties will take effect even if `Text#isUnderline` is false. This should probably be stated at the class level and for each underline-related property

richtextfx/src/main/java/org/fxmisc/richtext/model/StyledText.java
```diff
@@ -94,15 +114,30 @@ public void encode(DataOutputStream os, StyledText<S> t) throws IOException {
...
+
+    /**
+     * The style of this piece of styled text.
+     */
     public S getStyle() { return style; }
 
+    /**
+     * Creates a new StyledText with the same content but the given style.
+     *
+     * <p>This class is unmodified.</p>
```
JordanMartinez: Does "This class is unmodified" need to be there since you already specify that the method creates a new object and the class-level javadoc says the object is immutable?

twistedsquare:
I believe I've addressed all your comments in the newest commits. Thanks for taking the time to consider them.

JordanMartinez:
Thanks!

PR #483, opened 2017-04-07
Fix regression: changing empty paragraph's style is a valid change

### LINKED ISSUE #481: setParagraphStyle throws exception
With version 0.7M5, if I call setParagraphStyle, I get an exception saying that it's invalid to have a (rich text) change which doesn't add or remove text.  This seems like a bug given that setParagraphStyle does exactly that, and in general it should be valid to have a rich text change which only changes styles.  Stack trace:

```
java.lang.IllegalStateException: Cannot get the type of a change that neither inserts nor deletes anything.
	at org.fxmisc.richtext.model.TextChange.<init>(TextChange.java:31)
	at org.fxmisc.richtext.model.RichTextChange.<init>(RichTextChange.java:6)
	at org.fxmisc.richtext.model.ReadOnlyStyledDocument.lambda$null$7(ReadOnlyStyledDocument.java:274)
	at org.reactfx.util.Tuple2.map(Tuple2.java:31)
	at org.fxmisc.richtext.model.ReadOnlyStyledDocument.lambda$replace$8(ReadOnlyStyledDocument.java:271)
	at org.reactfx.util.Tuple2.map(Tuple2.java:31)
	at org.fxmisc.richtext.model.ReadOnlyStyledDocument.replace(ReadOnlyStyledDocument.java:270)
	at org.fxmisc.richtext.model.ReadOnlyStyledDocument.replaceParagraph(ReadOnlyStyledDocument.java:285)
	at org.fxmisc.richtext.model.GenericEditableStyledDocumentBase.setParagraphStyle(GenericEditableStyledDocumentBase.java:179)
	at org.fxmisc.richtext.model.SimpleEditableStyledDocument.setParagraphStyle(SimpleEditableStyledDocument.java:7)
```

JordanMartinez:
Fixes #481

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/model/TextChange.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/model/TextChange.java
@@ -10,12 +10,16 @@ public abstract class TextChange<S, Self extends TextChange<S, Self>> {
         INSERTION,
         /** Indicates that the change will delete something but not insert anything */
         DELETION,
-        /** Indicates that the change will remove something and insert something as its replacement */
-        REPLACEMENT
+        /** Indicates that the change will remove something and insert something as its replacement.
+         * For {@link PlainTextChange}s, the replacement will only be content.
+         * For {@link RichTextChange}s, the replacement may only be the style, only the content, or a mix of both.
+         * Style changes may change the style of the paragraph's content or a non-empty {@link Paragraph}
+         * ({@code p.length() > 0}) */
+        REPLACEMENT,
+        /** Indicates that the paragraph style of an empty {@link Paragraph} ({@code p.length() == 0}) was changed */
+        RESTYLED_PARAGRAPH
     }
 
-    private ChangeType type;
-    public final ChangeType getType() { return type; }
 
     protected final int position;
     protected final S removed;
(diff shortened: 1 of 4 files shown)


richtextfx/src/main/java/org/fxmisc/richtext/model/TextChange.java
```diff
@@ -10,12 +10,16 @@
...
-        /** Indicates that the change will remove something and insert something as its replacement */
-        REPLACEMENT
+        /** Indicates that the change will remove something and insert something as its replacement.
+         * For {@link PlainTextChange}s, the replacement will only be content.
+         * For {@link RichTextChange}s, the replacement may only be the style, only the content, or a mix of both.
+         * Style changes may change the style of the paragraph's content or a non-empty {@link Paragraph}
+         * ({@code p.length() > 0}) */
+        REPLACEMENT,
+        /** Indicates that the paragraph style of an empty {@link Paragraph} ({@code p.length() == 0}) was changed */
+        RESTYLED_PARAGRAPH
```
JordanMartinez: Thinking about this now, I think this should be named "RESTYLED_EMPTY_PARAGRAPH" to be clearer and prevent confusion (e.g. a user thinks it occurs when a non-empty paragraph is restyled

Issue #486, opened 2017-04-10
Refactor: ChangeType in TextChange


JordanMartinez:
To address #322, I reimplemented the solution in #402 by using more readable code via `ChangeType`, which was through lazy evaluation initially. This caused regressions, as discovered in #472, and was later evaluated immediately in #470 to help find these regressions. Besides simplifying the logic, I used `ChangeType` because I thought developers would be able to use it in their code to quickly determine what kind of change the change was. By doing this, I realize now that I was solving the first problem and indirectly trying to solve another problem (adding convenience methods to `TextChange` that allows one to quickly determine what kind of change it was). 

And that is the issue: my solution to the second problem was not fully thought out because it was not the issue I was trying to solve. I didn't realize it then (but do realize it now because of the regressions, both known and yet unknown, that have appeared since `0.7-M3`) that a `RichTextChange` can be a number of different changes (insertion, deletion, replacement, text style change, or a paragraph style change)--change tyes which `ChangeType` does not cover. 

So I see two options for addressing this bug:

1. Refactor the code to always create a `TextChange` that can get exactly what kind of change type it is
2. Change the way the code knows that a `TextChange` is an insertion or deletion type of change and thus whether to merge it with another `TextChange`.

I'm against the former because it will be more invasive to the current API and using a `TextChange` does not always require knowing what kind of change it is. For example, sometimes one just needs to know how much to adjust the caret based on how much content was removed and inserted.
I think the latter is best because it will immediately address all possible (known and currently unknown) regressions related to #458.

To still maintain code readability, I don't think we should just implement methods like `TextChange#isInsertion` and `TextChange#isDeletion`. Rather, I think `ChangeType` (which I think needs to be better named since it doesn't address all the changes a `TextChange` could be and since it pollutes the namespace that developers might want to use later in their own code) should be as follows:
````java
// Enum needs a better name!
public class ChangeType {
    INSERTION,
    DELETION,
    NEITHER // indicates it is neither an insertion/deletion
````
And then the `TextChange#mergeWith` code would still read:
````java
// this is either a insertion/deletion
if (this.getType() != ChangeType.NEITHER 
    // the two types match: both are insertion or both are deletion
    && this.getType() == that.getType()) { 
    // merge code
}
````
Unfortunately, I can't think of a better name to address the namespace pollution issue I raised above. Does anyone else have a better name?

[...]

PR #487, opened 2017-04-12
Limit TextChange's type's possibilities by only using it for merging

JordanMartinez:
Fixes regressions introduced by #458 by implementing the changes proposed in #486.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/model/TextChange.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/model/TextChange.java
@@ -5,30 +5,39 @@ import java.util.Optional;
 
 public abstract class TextChange<S, Self extends TextChange<S, Self>> {
 
-    public static enum ChangeType {
+    /**
+     * Indicates whether this change can be merged with another change if they are both {@link #INSERTION} or
+     * both {@link #DELETION}.
+     */
+    public static enum MergeType {
         /** Indicates that the change will insert something but not remove anything */
         INSERTION,
         /** Indicates that the change will delete something but not insert anything */
         DELETION,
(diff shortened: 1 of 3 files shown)


richtextfx/src/main/java/org/fxmisc/richtext/model/TextChange.java
```diff
@@ -5,30 +5,39 @@
...
-        /** Indicates that the change will remove something and insert something as its replacement.
-         * For {@link PlainTextChange}s, the replacement will only be content.
-         * For {@link RichTextChange}s, the replacement may only be the style, only the content, or a mix of both.
-         * Style changes may change the style of the paragraph's content or a non-empty {@link Paragraph}
-         * ({@code p.length() > 0}) */
-        REPLACEMENT,
-        /** Indicates that the paragraph style of an empty {@link Paragraph} ({@code p.length() == 0}) was changed */
-        RESTYLED_PARAGRAPH
+        /** Indicates that the change is a style change, a replacement or something other than the other two types */
+        NEITHER,
```
kuc: I really don't like that name. When someone would be reading code, "neither of WHAT?" will be constant question.
Because "style change, a replacement or something other" implies, in code below, that no merge will be done, I propose to rename this to `MergeType.NONE`

JordanMartinez: I like `NONE` better, but it still leaves the similar question of "if it isn't none, then what is it?" :smile:

I think we'll just take the minor performance hit and make the code more readable by writing it like so:
````java
if (this.getType == INSERTION || this.getType() == DELETION)
    && this.getType() == that.getType() // and so forth
````

Issue #488, opened 2017-04-12
WrapText + LineNumber, wrong line numbers position


AlexP11223:
When text wrapping and line numbers are enabled, line numbers are displayed incorrectly (which is very confusing and makes them unusable):

![](http://i.imgur.com/HghKg1d.png?1)

They displayed in the center instead of the beginning of the line, like in other text editors. For example Notepad++:

![](http://i.imgur.com/MSgxlGJ.png)

Tried in current master spellchecking demo, as well as in my project with stable and milestone RichTextFX releases.

        StyleClassedTextArea textArea = new StyleClassedTextArea();
        textArea.setWrapText(true);
        textArea.setParagraphGraphicFactory(LineNumberFactory.get(textArea));

JordanMartinez:
Since you can implement your own `ParagraphGraphicFactory`, use `LineNumberFactory` as a base upon which you implement the alignment you want.

AlexP11223:
Isn't it a bug? All editors I know (JetBrains IDEs, Github, ...) use top alignment for line numbers.

JordanMartinez:
Right, but AFAIK, all code editors don't wrap their lines like you are doing.

AlexP11223:
They all have option to wrap, and it can be useful for non-code text files, some config files.

JordanMartinez:
Mm.. makes sense. If we just make LNF layout its children at the top, won't that properly lay out the number regardless of wrapping?

AlexP11223:
Yeah, it should. If I add `lineNo.setAlignment(Pos.TOP_CENTER);` to LNF `apply` it works as expected.

btw I thought that it would be possible to use right alignment and remove zeros added for padding, but looks like in that case width would not be the same :(

![](http://i.imgur.com/jwpwaQ9.png)

JordanMartinez:
Couldn't you use `Pos.TOP_RIGHT`?

JordanMartinez:
Or is that what you meant by "I thought that it would be possible to use right alignment?"

AlexP11223:
Yeah, Currently there is no difference between RIGHT and CENTER because it is padded with 0s.

JordanMartinez:
[...]

PR #489, opened 2017-04-14
Fix line numbers alignment and change default padding to spaces. #488

AlexP11223:
Fixed line numbers alignment and changed default padding to spaces as discussed in #488

Before:

After:


JordanMartinez:
Thanks!

PR #495, opened 2017-04-28
Update undo fx

JordanMartinez:
I took a short break from homework.

### DIFF
--- a/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/richtext/RichText.java
+++ b/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/richtext/RichText.java
@@ -130,8 +130,8 @@ public class RichText extends Application {
         textColorPicker.valueProperty().addListener((o, old, color) -> updateTextColor(color));
         backgroundColorPicker.valueProperty().addListener((o, old, color) -> updateBackgroundColor(color));
 
-        undoBtn.disableProperty().bind(Bindings.not(area.undoAvailableProperty()));
-        redoBtn.disableProperty().bind(Bindings.not(area.redoAvailableProperty()));
+        undoBtn.disableProperty().bind(area.undoAvailableProperty().map(x -> !x));
+        redoBtn.disableProperty().bind(area.redoAvailableProperty().map(x -> !x));
 
         BooleanBinding selectionEmpty = new BooleanBinding() {
             { bind(area.selectionProperty()); }
--- a/richtextfx/src/main/java/org/fxmisc/richtext/model/UndoActions.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/model/UndoActions.java
@@ -1,9 +1,8 @@
 package org.fxmisc.richtext.model;
 
-import javafx.beans.value.ObservableBooleanValue;
-
 import org.fxmisc.undo.UndoManager;
 import org.fxmisc.undo.UndoManagerFactory;
+import org.reactfx.value.Val;
 
 /**
  * Undo/redo actions for {@link TextEditingArea}.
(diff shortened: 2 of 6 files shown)


JordanMartinez:
I could also ignore any `replace` calls that will result in an identity change by calling:
````java
// check the start and end indexes
if (getDocument(start, end).equals(replacement)) {
    return;
}
// the actual replacement code
````

**Comment by TomasMikula** [comment]
If your only motivation is performance, I say don't do it. Extracting a subsequence from a document has about the same cost as replacing it.

**Comment by JordanMartinez** [comment]
Ok. That's what I was thinking, but I wasn't sure how much of a performance hit that would be.

File: richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/richtext/RichText.java [hunk 1]
```diff
@@ -130,8 +130,8 @@ public void start(Stage primaryStage) {
         textColorPicker.valueProperty().addListener((o, old, color) -> updateTextColor(color));
         backgroundColorPicker.valueProperty().addListener((o, old, color) -> updateBackgroundColor(color));
 
-        undoBtn.disableProperty().bind(Bindings.not(area.undoAvailableProperty()));
-        redoBtn.disableProperty().bind(Bindings.not(area.redoAvailableProperty()));
+        area.undoAvailableProperty().values().feedTo(undoBtn.disableProperty());
+        area.redoAvailableProperty().values().feedTo(redoBtn.disableProperty());
```
JordanMartinez: I'm not modeling good coding by storing the resulting `Subscription`, but... is that really necessary?

TomasMikula: That would be necessary when either

 - you wanted to reuse the button later for a different area (obviously); or
 - (less obviously, and more importantly) if the life time of the area was longer than that of the button, so that the button could be garbage collected when no longer used.

Btw, you can also do

```java
undoBtn.disableProperty().bind(area.undoAvailableProperty().map(x -> !x));
```

JordanMartinez: > That would be necessary when either
>
>   - you wanted to reuse the button later for a different area (obviously); or
>   - (less obviously, and more importantly) if the life time of the area was longer than that of the button, so that the button could be garbage collected when no longer used.

Right, but my point is that since neither of the above occur, is it ok if I just do what I did since `Subscription#unsubscribe` will never be called?

> undoBtn.disableProperty().bind(area.undoAvailableProperty().map(x -> !x));

Ah, I like that approach better.

TomasMikula:
Yes, it would be OK.

JordanMartinez:
I forgot that I could merge this... lol, I was expecting you to.

richtextfx/src/main/java/org/fxmisc/richtext/model/UndoActions.java
```diff
@@ -21,8 +20,8 @@
     default void redo() { getUndoManager().redo(); }
 
     default boolean isUndoAvailable() { return getUndoManager().isUndoAvailable(); }
-    default ObservableBooleanValue undoAvailableProperty() { return getUndoManager().undoAvailableProperty(); }
+    default Val<Boolean> undoAvailableProperty() { return getUndoManager().undoAvailableProperty(); }
 
     default boolean isRedoAvailable() { return getUndoManager().isRedoAvailable(); }
-    default ObservableBooleanValue redoAvailableProperty() { return getUndoManager().redoAvailableProperty(); }
+    default Val<Boolean> redoAvailableProperty() { return getUndoManager().redoAvailableProperty(); }
```
JordanMartinez: Just FYI. I get a compiler warning on this and the corresponding undo line: "Unchecked assignment `Val` to `Val<Boolean>`. Reason `getUndoManager()` has raw type so result of `redoAvailableProperty()` is erased"

If I add the type parameter (`UndoActions<C>` and `UndoManager<C> getUndoManager()`), they go away. However, then I can't specify the type in `GenericStyledArea` since the option to create both a plain and rich undo manager exists.

TomasMikula: `getUndoManager()` should return `UndoManager<*>`.

TomasMikula: And by `UndoManager<*>` I mean `UndoManager<?>`. Apparently, the process of garbage collecting my Java knowledge has been initiated 😆

JordanMartinez: You should have used a different garbage collector ;-) (and is that some reference to Scala's way of generics?)

> `getUndoManager()` should return `UndoManager<?>`

Yeah, I tried that and... wait, no I didn't try that :sweat_smile:

JordanMartinez: I think I misled myself by using `C` as the type parameter instead of `?`. Hmm... Should the class-to-interface conversion for `TextChange` still be accepted though? The refactoring isn't necessary, but it might be forward-thinking.

PR #496, opened 2017-04-28
Revert to previous merge method

JordanMartinez:
Reverts back to the previous merge method


TomasMikula:
The type param on `UndoActions` should not be necessary.

JordanMartinez:
As said in my other comment, should the `TextChange` class-to-interface conversion be done at all? If it should, I feel like it should be in its own PR that does a lot of other refactoring.

TomasMikula:
Yeah, not in this PR. Not sure what it gives us anyway.

JordanMartinez:
Not really anything. It was only necessary because of my error in adding a type param to `UndoActions`

I'll fix this tomorrow.

JordanMartinez:
Ok. Only the revert is in this PR now.

Issue #484, opened 2017-04-07
Optimize ParagraphText's background and underline shapes


JordanMartinez:
Coming from [Tomas' comment in 411](https://github.com/TomasMikula/RichTextFX/issues/411#issuecomment-290228778), optimize `ParagraphText` to only instantiate its `backgroundShape` and `underlineShape` objects when needed.

JordanMartinez:
This was more properly closed by #558. The shapes were further optimized by only using one shape per shared value between multiple consecutive TextExt nodes rather than one per segment.

Issue #346, opened 2016-08-05
Add support for a border around some text


NavalOrange:
Hey there, I was wondering if there was a nice way to get a border for a highlight area in a StyleClassedTextArea. And if not I would like to make a feature request : D.

Something similar to what I'm looking for is the name change indicator in IntelliJ.
![intellijhighlightborder](https://cloud.githubusercontent.com/assets/1515071/17426660/d526949e-5a8e-11e6-84cf-99811b9398eb.png)

Specifically what I was looking to do is highlight locations in a text file based on user inputted row, column and lengths. For most of the use cases just having a colored background is fine. However, in the event that the user inputs a zero length location, there is nothing to highlight. With a border around a zero width selection, there would be feedback for the user.

In IntelliJ, it looks like this:
![intellijhighlightborder2](https://cloud.githubusercontent.com/assets/1515071/17426754/b9805bfc-5a8f-11e6-8ce5-95b75411a43c.png)

Thanks for the help!

JordanMartinez:
> I was wondering if there was a nice way to get a border for a highlight area in a StyleClassedTextArea. And if not I would like to make a feature request : D.

Currently, there is not a way to do that. Something similar was already asked in #337 (see the comments after the initial issue).

NavalOrange:
Aaah sorry about that, thanks for the info!

JordanMartinez:
@NavalOrange This is possible with #455, but it still wouldn't be the best way to do it.

JordanMartinez:
This is slightly outside the scope of this issue, but I've been wondering whether the entire style handling for rich text should be completely refactored. As pointed out [here](https://github.com/TomasMikula/RichTextFX/pull/497#issuecomment-298988944) and originally [here](https://github.com/TomasMikula/RichTextFX/pull/166#issuecomment-131957896), there is a white (or same color as background) line that appears when two texts share the same background color but not the same style, resulting in 2+ background shapes being used for that background color.
By following the same approach we are currently using (each Text segment gets its own background shape), this break continues to occur now and will show up when the border text feature is implemented.

So, I'm wondering whether we could apply styles in a different ways. For example, style that apply only to the text (its size, its fill color, etc.) should be extracted from the style object and applied to that text. However, styles that apply to things behind/around the text (background color, underline, border around the text) should be handled separately by `ParagraphText` (or perhaps `TextFlowExt`).

[...]

PR #497, opened 2017-05-03
Support for setting stroke colour and width of background shapes in TextExt

twistedsquare:
I wonder if you're interested in adding this feature.  It allows you to use stroke as a highlight for some text, instead of or in addition to background colour.  The main limitation is noted in the docs: adjacent segments with different styles will get their strokes painted separately.  That will make it less useful for some use cases, but if you just have a fairly plain text editor and want to add support for highlighting e.g. find results by drawing around them instead of using a background colour, it's useful.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/ParagraphText.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/ParagraphText.java
@@ -18,6 +18,7 @@ import javafx.scene.shape.Path;
 import javafx.scene.shape.PathElement;
 import javafx.scene.shape.StrokeLineCap;
 
+import javafx.scene.shape.StrokeType;
 import org.fxmisc.richtext.model.Paragraph;
 import org.reactfx.value.Val;
 import org.reactfx.value.Var;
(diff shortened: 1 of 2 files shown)


JordanMartinez:
What problem is this PR trying to solve? "support for highlighting" ?

twistedsquare:
The problem was "the editor doesn't support drawing a border around particular text segments".  I was porting an interface which used background colour and a stroked border around the region to highlight search results in the editor.  TextExt lacked support for the stroke aspect and I couldn't see a work-around from outside the library, so I added it.  I know it's not a desperately lacking feature but I thought I would offer the code, to see if you felt enough other people would have similar use cases.

JordanMartinez:
I think the feature is worth adding, but the main limitation "adjacent segments with different styles will get their strokes painted separately" presents an issue. I think this should be further discussed in #346 before attempting a PR. Additionally, #484 might play a role in it, too.

twistedsquare:
Ah, sorry -- didn't spot there was already a request for it.  Not sure which thread to point this out, but there is already an issue with adjacent segments with different styles and solid backgrounds: a faint sliver of the main editor background colour is visible between adjacent segments with different styles but the same background colour (example below).  So I figured it wasn't adding a new restriction, it's just more prominent when you're drawing a stroke.

image

JordanMartinez:
> Ah, sorry -- didn't spot there was already a request for it.

No worries! It happens.

> There is already an issue with adjacent segments with different styles and solid backgrounds: a faint sliver of the main editor background colour is visible between adjacent segments with different styles but the same background colour (example below)

Yeah. I think I noted the same thing when the feature was first introduced. However, because it was not super noticeable, nothing else was done.

Let's move further discussion to the issue, not this PR.

JordanMartinez:
@twistedsquare Would you like to re-implement this again after I merge #558? The "adjacent segments" issue raised earlier is addressed in that PR.

JordanMartinez:
Closing since #560 added this support.

Issue #499, opened 2017-05-05
Calling getCharacterBoundsOnScreen on a paragraph with selection stops the selection being displayed


twistedsquare:
If you have a paragraph that features part of the selection, and you call getCharacterBoundsOnScreen, then the selection will no longer be displayed on that paragraph until you change the selection extent, or scroll the paragraph in and out of view (which resets the display).  Affects 0.7-M5.

I have the fix for this (pull request coming in a sec) but thought I'd record the issue first.

JordanMartinez:
Good catch! (and thanks for opening an issue for it before submitting a PR)

JordanMartinez:
Closed via #500

PR #500, opened 2017-05-05
Fixes issue #499: calling getCharacterBoundsOnScreen on a paragraph w…

twistedsquare:
…ith selection stops the selection being displayed.

The getCharacterBoundsOnScreen method borrows the selectionShape to do the calculation, but it was only taking a reference to the list beforehand, so the saved "copy" was getting wiped out and thus not restoring the selection properly afterwards.  Simple fix: take copy of the list, not a reference.


JordanMartinez:
Thanks!

PR #501, opened 2017-05-11
Expose visible paragraphs

JordanMartinez:
I feel like this should be suspended somewhere...


JordanMartinez:
Ok, I've updated it to suspend whenever virtual flow might change the displayed cells. I'm not sure what kinds of tests should be run on it.

PR #504, opened 2017-05-11
Write more TestFX tests

JordanMartinez:
I had to shorten some of the test names on the last commit (cut, copy, paste tests) because I was getting build failures due to Gradle being unable to write the test report. I guess it's due to too long of a path name. [This issue](https://discuss.gradle.org/t/gradle-is-generating-files-with-their-names-too-big-in-cucumber-tests/8537/3) may be what's causing it, but I was running it on Gradle 3.5.

In addition, the `delete[Next/Previous]Word` deletes the word and the following space (or whatever else is picked up by `wordBreaks[Forwards/Backwords](2)`. Is this a bug? Or is the deleted space intentional?


JordanMartinez:
I'm also wondering whether `StyledTextAreaBehaviorTest` should be moved to its own package in the test packages, and move each test suite into its own test class. This would minimize the long path name issue I encountered above and make the code more readable.

JordanMartinez:
Aye, that approach is definitely cleaner.

JordanMartinez:
Hmm... it built fine on mine multiple times, though there were some flaky tests from time to time...

I'll have to look at this again later in the week.

JordanMartinez:
Well that doesn't make any sense. Without the force-display stage code, the tests do not pass on my machine but it does on Travis, despite both of us using the Linux OS...?

JordanMartinez:
I'm guessing this has to do with running the code in headless mode, but I'm not sure how to do that on my end. At the same time, I don't understand why headless mode would work properly whereas head (?) mode wouldn't.

JordanMartinez:
After upgrading to the latest TestFX release, the tests passed on my machine.... very strange indeed.

JordanMartinez:
Ok, looks like everything works, except for Page Down for very strange reasons. I'm not sure why, but at least I can move on to the next tests.

JordanMartinez:
@TomasMikula @twistedsquare since you two use Mac, could one of you test out this PR? I'm going to clean this PR up a bit and then merge it. However, I want to make sure it builds on a Mac, so we don't get numerous issues later about it not building properly on other people's systems.

JordanMartinez:
Actually, hold that thought. There's one more test I was going to write, but I'll have to do that later in the week.

JordanMartinez:
I'm not sure how to properly set up the Travis files to get these to work properly, so I've filed TestFX/TestFX#396

JordanMartinez:
The Linux builds have been failing because when I copied TestFX's Travis files over, they were building on the Trusty distro whereas we were building on the Precise distro. However, I'm not sure why that would make a difference. Both the OSX and Trusty builds fail on the same test in ContextMenuTests.

JordanMartinez:
> The Linux builds have been failing because when I copied TestFX's Travis files over, they were building on the Trusty distro whereas we were building on the Precise distro. However, I'm not sure why that would make a difference.

I removed the OSX environment and the two Linus builds passed. After comparing the two, it seems the Linux builds fail when `install: true` is used. Everything else appears to be the same.

JordanMartinez:
@TomasMikula Can you sign up this project for Appveyor, so we can add CI for Windows as well?

JordanMartinez:
I cleaned up the commit history. Everything is the same except for `ContextMenuTests`, which now shows the context menu through means other than a simulated user right click. I'm not sure if that will cause the two tests that were otherwise failing to pass in Travis on OSX, but it's a start.

As for the other tests that are failing, they all relate to using home/end, up/down on a multi-line paragraph when no modifiers are down and when only shift is down. Strangely, when shift and shortcut are down, they work correctly.
I need to look at the `SelectionPolicy.EXTEND` that's sometimes used for OSX environments to insure that my tests aren't failing because I'm testing for behavior that isn't normal on that OS.

JordanMartinez:
I looked at the `SelectionPolicy.EXTEND`, and it's only used in shift + left mouse presses. I'm not sure why the multi-line navigation tests would fail. As for the context menu tests, not the opposite ones fail. Very strange...

I think for now, I'll add an `assumeTrue` statement, so that the tests that otherwise fail on OSX only get run on Linux. This would allow me to continue to work on other parts of this project while still insuring most of the tests pass.

JordanMartinez:
Ok. I finally got the multi-line paragraph navigation tests to pass. Now I just need to get these remaining 2 ContextMenu tests to pass (or go back to my idea of skipping them if run on a mac).

JordanMartinez:
Before merging this, I think I should clean up the tests once more and rename some tests to explain what they do (in other words, `homeMovesCaretToStartOfArea()` instead of `home()`).

PR #511, opened 2017-05-28
Only fire MouseOverText END event when a BEGIN event has already fired

### LINKED ISSUE #506: Setting MouseOverTextDelay while mouse is over caret immediately emits the END event when no BEGIN event was ever fired
Using the following test:
````java
public class ViewTests extends InlineCssTextAreaAppTest {

    @Test
    public void testMouseOverTextDelay() {
        // setup
        SimpleBooleanProperty beginFired = new SimpleBooleanProperty(false);
        SimpleBooleanProperty endFired = new SimpleBooleanProperty(false);
        area.addEventHandler(MouseOverTextEvent.MOUSE_OVER_TEXT_BEGIN, e -> beginFired.set(true));
        area.addEventHandler(MouseOverTextEvent.MOUSE_OVER_TEXT_END,   e -> endFired.set(true));

        interact(() -> {
            area.replaceText("a long line of some example text");
            area.setMouseOverTextDelay(null);   // insure nothing is set
        });

        // no delay set -> nothing fires
        moveTo(firstLineOfArea()).sleep(300);
        assertFalse(beginFired.get());
        assertFalse(endFired.get());

        // delay -> begin event fires
        area.setMouseOverTextDelay(Duration.ofMillis(300));
        moveBy(20, 0);
        assertFalse(beginFired.get());

        assertFalse(endFired.get()); // <--- FAILS HERE

        sleep(400);
        assertTrue(beginFired.get());
        assertFalse(endFired.get());

        beginFired.set(false);

        // end event fires
        moveBy(0, 30);
        assertFalse(beginFired.get());
        assertTrue(endFired.get());
    }

}
````

One would expect the end event to fire only after a begin event has been fired.

### PR DESCRIPTION
**Pull request by JordanMartinez**
Fixes #506.

If a delay is set, a BEGIN event is fired, and the delay is set to `null`, the corresponding END event does not currently fire when another MouseEvent is fired. Is that desired? This behavior does not occur (i.e. the END event does fire when another MouseEvent is fired) when the delay is set to a non-null delay value.

**Edit:** The above "bug" might be something we want to change in the future, but for right now, it's not a high priority. Thus, I'm going to leave this as the default behavior. I've updated the javadoc to note this quirk.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/ParagraphBox.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/ParagraphBox.java
@@ -114,7 +114,7 @@ class ParagraphBox<PS, SEG, S> extends Region {
         return text.getParagraph();
     }
 
-    public EventStream<Either<Tuple2<Point2D, Integer>, Object>> stationaryIndices(Duration delay) {
+    public EventStream<Either<Tuple2<Point2D, Integer>, Object>> stationaryIndices(Duration delay, BooleanProperty beginEventFired) {
         EventStream<Either<Point2D, Void>> stationaryEvents = new MouseStationaryHelper(this).events(delay);
         EventStream<Tuple2<Point2D, Integer>> hits = stationaryEvents.filterMap(Either::asLeft)
                 .filterMap(p -> {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: richtextfx/src/main/java/org/fxmisc/richtext/ParagraphBox.java [hunk 1]
```diff
@@ -125,7 +125,9 @@ public String toString() {
                         return Optional.empty();
                     }
                 });
-        EventStream<?> stops = stationaryEvents.filter(Either::isRight).map(Either::getRight);
+        EventStream<?> stops = stationaryEvents.filter(Either::isRight).map(Either::getRight)
+                // only fire END event when a BEGIN event has already been fired
+                .conditionOn(beginEventFired);
```
JordanMartinez: This line should occur immediately after stationaryEvents to speed up the process

Issue #513, opened 2017-06-01
Build fails on Linux with: "error: unmappable character for encoding ASCII"


cbvms123:
Compiling and building on Linux fails because of an UTF-8 encoded source file:
richtextfx/src/main/java/org/fxmisc/richtext/model/TextEditingArea.java

Adding the encoding options
build.gradle: compileJava.options.encoding = 'UTF-8'

richtextfx/build.gradle(javadoc section): options.addStringOption('encoding', 'UTF-8')

Resolved the issue in my environment.

JordanMartinez:
I use Linux and it builds fine on mine. Why does it fail on yours but not mine? Do our locations have anything to do with it?

cbvms123:
Indeed, I checked my locale related settings and found when unsetting LC_CTYPE  the build works also on my box. Maybe with LC_CTYPE=C you may be able to reproduce the error on your side as well.

JordanMartinez:
I'd guess that's the issue. I've contributed to this project for a while and never had that arise.

Via the command `env`, I don't have an environment variable for `LC_CTYPE`, but I do for LANG: `LANG=en_US.UTF-8`

PR #514, opened 2017-06-01
Added options to specify UTF-8 encoding, so that build on Linux works

cbvms123:
Fix for UTF-8 compile issue:#513


JordanMartinez:
Closing since we know this was caused by setting `LANG_CTYPE` on Linux

PR #520, opened 2017-06-11
C# Async Example

yemikudaisi:
Added C# keyword example with VS C# Highlighting based style.

### DIFF
--- a/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/CSharpKeywordAsync.java
+++ b/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/CSharpKeywordAsync.java
@@ -1,0 +1,164 @@
+package org.fxmisc.richtext.demo;
+
+import java.time.Duration;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Optional;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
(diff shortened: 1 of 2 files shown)


JordanMartinez:
Hmm....

How is this different (besides demonstrating a different language) from the Java Async demo?

yemikudaisi:
it isn't actually, I'm just try to make someone else's work easier with the C# keywords

JordanMartinez:
Hmm....

If I were to merge this, it sets a precedent for every person who creates a `[Language]Async` demo that is functionally-equivalent to the `JavaAsync` demo. I think for that reason, this PR as it is now shouldn't be merged.

However, it brings up a different question. What if, instead of having multiple Async demos for different languages, we provide a `LanguageAsync` demo (feel free to come up with a better name), that demonstrates different syntax highlighting for various programming languages? Then, one could add another pattern for the language's keywords, an example source text in that language, and the stylesheet needed to highlight its syntax correctly. By doing that, we wouldn't spam the demos folder with multiple `LanguageAsync` demos that do the exact same thing.

If so, we'd need to add a combo box that one could use to select a new language. When selected, the area's text would replaced with that language's text, the stylesheets would be switched to that language's stylesheet, and the compute highlighting code would be called again.

Your thoughts?

yemikudaisi:
It seems more appropriate, I will look into it and come up with a viable solution for review.

JordanMartinez:
Are you still interested in doing this? Or should I close this PR?

JordanMartinez:
Closed due to lack of response.

PR #523, opened 2017-06-17
Cleanup some aspects of the code

JordanMartinez:


JordanMartinez:
When building using Gradle 4.0, it fails. When using 3.5, it succeeds. Looks like we need to update the `build.gradle` file...?

Issue #362, opened 2016-09-19
Undo merges with subsequent changes despite a reasonable period of user inactivity when it shouldn't.


eliottizi:
Hi,
I am currently building an IDE using RichTextFX and when i use the method undo , redo it didnt work properly.
If you undo what you write it delete all the text you write normaly it only delete the last word or the last line.
If you don't do any action like copy/paste/cut on the code the undo method will clear all the code area even if u tiped 1000 lines. 
Hope u will understand !
Windows 10 x64, coding with NetBeans.
I'm using the RichTextFX FatJar version 0.7

JordanMartinez:
@eliottizi I'm going to try clarifying what you've written. Is this correct?

1. Add text to the area by manually typing content into it. In other words, don't copy/paste something.
2. Call `undo` via `CTRL + Z` and undo your work
 - Expectation: delete the last word on the last line

My guess is that every time you input something via the keyboard, the previous undo is merged with the new change you have done (e.g. added a character). i think if you pause for a bit, it stops this, but I've never fully understood how this part of the code works.

TomasMikula:
Yes, if subsequent edits "touch" or overlap each other, they are merged. As a consequence (that I haven't realized until now), if you start with an empty document, every edit will be merged into the previous one. Pausing will not help. We might want to provide a more meaningful default behavior. In the meantime, you can invoke `area.getUndoManager().preventMerge()` manually. A reasonable time to invoke it might be after the user paused typing for a second or two. See e.g. [Trigger processing after a period of inactivity](http://tomasmikula.github.io/blog/2014/03/22/trigger-processing-after-a-period-of-inactivity.html) for how to do it.

eliottizi:
Yes that's it @JordanMartinez  !
I will try it @TomasMikula .
Thanks for your answers guys ;)

JordanMartinez:
@TomasMikula We could quickly implement such an idea via a `StateMachine` that stores a boolean (false = do not merge changes; true = merge changes when possible) for determining whether to `preventMerge` or not:
- initial state = false
- user starts typing: state = true; a timer starts
 - if user types before timer ends, timer restarts
- if timer finishes, state = false
- if any other text manipulation occurs (e.g. cut, paste, developer-made function), state = false & timer is cancelled.

However, I see one type of bug that could arise from this: 
- Let's say the user is typing some long text.
- Before the timer runs out, the user calls some shortcut that will execute a custom developer-made text-manipulating function. 
[...]

PR #530, opened 2017-06-22
Stop next change from merging with previous one after inactive period.

JordanMartinez:
Addresses #362.

This should really be changed in UndoFX. But since that won't have a new release in the foreseeable future, this can bring the same functionality until such a release exists.


JordanMartinez:
This PR needs a test to prove that the merge prevention actually occurs.

JordanMartinez:
I also realized that the test in UndoManagerTest called "deletingTextThatWasJustInsertedShouldNotMergeTheTwoChanges()" is no longer needed due to reverting back to the original merge behavior in #496. I'll remove that in a separate PR.

PR #550, opened 2017-07-31
Add Hyperlink demo - minimum needed for custom object integration

JordanMartinez:
Shows custom object support in a less-cluttered way than the RichTextDemo.

### DIFF
--- a/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/hyperlink/HyperlinkOps.java
+++ b/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/hyperlink/HyperlinkOps.java
@@ -1,0 +1,97 @@
+package org.fxmisc.richtext.demo.hyperlink;
+
+import org.fxmisc.richtext.model.SegmentOps;
+
+import java.util.Optional;
+
+public class HyperlinkOps<S> implements SegmentOps<Hyperlink<S>, S> {
+
+    @Override
+    public int length(Hyperlink<S> hyperlink) {
+        return hyperlink.length();
+    }
+
+    @Override
+    public char charAt(Hyperlink<S> hyperlink, int index) {
+        return hyperlink.charAt(index);
+    }
+
+    @Override
+    public String getText(Hyperlink<S> hyperlink) {
+        return hyperlink.getDisplayedText();
+    }
+
+    @Override
+    public Hyperlink<S> subSequence(Hyperlink<S> hyperlink, int start, int end) {
+        return hyperlink.subSequence(start, end);
+    }
+
+    @Override
+    public Hyperlink<S> subSequence(Hyperlink<S> hyperlink, int start) {
+        return hyperlink.subSequence(start);
+    }
+
+    @Override
+    public S getStyle(Hyperlink<S> hyperlink) {
+        return hyperlink.getStyle();
+    }
+
+    @Override
+    public Hyperlink<S> setStyle(Hyperlink<S> hyperlink, S style) {
+        return hyperlink.setStyle(style);
+    }
+
+    @Override
+    public Optional<Hyperlink<S>> join(Hyperlink<S> currentSeg, Hyperlink<S> nextSeg) {
+        if (currentSeg.isEmpty()) {
+            if (nextSeg.isEmpty()) {
+                return Optional.empty();
+            } else {
+                return Optional.of(nextSeg);
+            }
+        } else {
+            if (nextSeg.isEmpty()) {
+                return Optional.of(currentSeg);
+            } else {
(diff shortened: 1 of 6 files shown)


richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/hyperlink/TextHyperlinkArea.java
```diff
@@ -0,0 +1,69 @@
...
+import org.fxmisc.richtext.GenericStyledArea;
+import org.fxmisc.richtext.TextExt;
+import org.fxmisc.richtext.model.ReadOnlyStyledDocument;
+import org.fxmisc.richtext.model.StyledText;
+import org.fxmisc.richtext.model.TextOps;
+import org.reactfx.util.Either;
+
+import java.util.function.Consumer;
+
+public class TextHyperlinkArea extends GenericStyledArea<ParStyle, Either<StyledText<TextStyle>, Hyperlink<TextStyle>>, TextStyle> {
```
stevedefazio: This demo seems like a lot of code just to demo hyperlinks. And it isn't clear to me at a quick glance if something like ParStyle is part of the minimal code necessary to create a hyperlink or if it's part of the bells and whistles of the demo.

JordanMartinez: > Not to disparage your work, it's great that you put this together.

None taken. I appreciate your feedback!

You are correct. `ParStyle` isn't needed at all.

If Java supported algebraic data types, writing the `SegmentOps` and real/empty hyperlink data types would be a lot easier to code... but it doesn't :-/

richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/hyperlink/HyperlinkOps.java
```diff
@@ -0,0 +1,41 @@
+package org.fxmisc.richtext.demo.hyperlink;
+
+import org.fxmisc.richtext.model.SegmentOpsBase;
+
+public class HyperlinkOps<S> extends SegmentOpsBase<Hyperlink<S>, S> {
```
stevedefazio: Where does SegmentOpsBase come from? I don't see it in here, and my compiler isn't finding it.

JordanMartinez: I added that in #553 since I realized that it's easier just to tell developers to extend a base class that already properly overrides everything than to tell them to copy/paste the same pattern in the wiki page.

No release currently has that, but you can get it if you follow the Jitpack instructions on the ReadMe.

richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/hyperlink/TextHyperlinkArea.java
JordanMartinez: I also reworked `StyledText`'s `equals` method to return true when both text objects are empty since `createText` will sometimes create an empty StyledText object but not use the one we want it to use (emptySeg).

Also, I'm not sure if our current pattern/approach is correct in one aspect: the empty segment's style. Technically, the segment is only empty in text, but not in style. The style is used when inserting text without specifying the style. The code will look at the style at the position of insertion and use that style for the text being inserted.

I think we need to update the code to use this approach:
````java
public Style getStyleForInsertion(int position) {
 if (useInitialTextstyle) {
  return initialTextStyle;
 } else {
  return getStyleAtPosition(position);
 }
}

public Style getStyleAtPosition(int position) {
 Optional<Style> s = getStyleAtPosition0(position);
 if (s.isPresent()) {
  return s.get();
 } else {
  return getStyleAtPosition(position - 1);
 }
}

public Optional<Style> getStyleAtPosition0(int position) {
 Style style = content.getStyleAtPosition(position);
 return Optional.ofNullable(style);
}
````

JordanMartinez:
This version follows `StyledText` in that it doesn't have a `real` and `empty` version of the same object.

Also, when concatting two hyperlinks together that share the same ancestor, it will form one hyperlink.

richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/hyperlink/HyperlinkOps.java
JordanMartinez: Just FYI. The updated version doesn't use `SegmentsOpsBase` It seems that even the empty segment object needs to be able to update its style because of what I mention in [my comment to your previous one](https://github.com/TomasMikula/RichTextFX/pull/550#discussion_r132587029). This is why I'm proposing #567.

PR #564, opened 2017-08-08
Enhancement: Selecting a newline char selects the rest of the line

### LINKED ISSUE #549: Improve selection display
When there is a selection with empty lines, the result may confuse the user :
img1

Some text editors use half-characters (github web editor) :
img2

And others fill the entire line (intellj) :

intellj

JordanMartinez:
Resolves #549 according to the IntelliJ style

Note: this PR should be refactored because `getParagraphSelection` now returns `IndexRange(start, paragraph.length() + 1)` in some cases.

I'd like to say that there should be two separate methods: one for getting just the range inside the paragraph and another that adds one if it includes the newline character.


JordanMartinez:
I decided to refactor `getRangeShape` related methods to now return a shape/bounds object that can include the newline character.

PR #565, opened 2017-08-08
Fix either ops bug

### LINKED ISSUE #554: Undoing on a Right of Either<SegmentOps, SegmentOps> throws an "Unexpected change received" exception
This is caused by `EitherOps` defaulting to the left ops when creating an empty segment.

JordanMartinez:
Resolves #554


JordanMartinez:
Odd.... Travis says the build passed, but GitHub doesn't seem to have registered the update...

PR #575, opened 2017-09-06
fix selection painting of empty lines if text area displays line numbers

JFormDesigner:
It the text area displays line numbers, then the selection of empty lines is not displayed correctly.

image

This PR fixes this.


JordanMartinez:
Not sure how long the Mac build will take, but I can't fathom why you're change would make it fail, so I'll merge this.

JordanMartinez:
Thanks again!

PR #577, opened 2017-09-07
adjust wrapped lines selection to the rest of the line

JFormDesigner:
Adjust right edge of wrapped lines selection to the rest of the line (similar to #549 / #564). IOW when selection the space character at the end of a wrapped line, then the selection is expanded to the right edge of the text area.

Old:
("01234 678 01" is a single line)

image image

New:

image image

Also closed rectangle path in `ParagraphText.createRectangle()`. Not sure whether this makes any difference, but `com.sun.javafx.text.PrismTextLayout.getRange()` does it too. So there might be a reason...


JordanMartinez:
Thanks!

Issue #573, opened 2017-09-06
NPE in clipboard copy


JFormDesigner:
I get an NPE when pressing `Ctrl+C`, but only if the selection includes a line separator and ends at the beginning of a line.

This is a regression. Used 0.7-M3 before where it worked. So I checked out various commits to find out the commit that broke it. It is commit a2aa8813e48e91e9eef461c20d28c5eac8724592. It works when I checkout the previous commit.

To reproduce, run this little app and press `Ctrl+C`:
~~~java
import javafx.application.Application;
import javafx.scene.Scene;
import javafx.stage.Stage;
import org.fxmisc.flowless.VirtualizedScrollPane;
import org.fxmisc.richtext.CodeArea;

public class ClipboardCopyNPE extends Application {
    @Override
    public void start(Stage primaryStage) throws Exception {
        CodeArea area = new CodeArea("abc\ndef\nghi");
        VirtualizedScrollPane<CodeArea> vsPane = new VirtualizedScrollPane<>(area);

        area.selectRange(2, 4);
        
        Scene scene = new Scene(vsPane, 400, 400);
        primaryStage.setScene(scene);
        primaryStage.show();
    }
}
~~~

The NPE is:
~~~
Exception in thread "JavaFX Application Thread" java.lang.NullPointerException
	at org.fxmisc.richtext.model.SuperCodec$1.encodeSuper(SuperCodec.java:33)
	at org.fxmisc.richtext.model.SuperCodec$1.encodeSuper(SuperCodec.java:1)
	at org.fxmisc.richtext.model.SuperCodec.encode(SuperCodec.java:15)
	at org.fxmisc.richtext.model.StyledText$2.encode(StyledText.java:94)
	at org.fxmisc.richtext.model.StyledText$2.encode(StyledText.java:1)
	at org.fxmisc.richtext.model.SuperCodec$1.encodeSuper(SuperCodec.java:35)
	at org.fxmisc.richtext.model.SuperCodec$1.encodeSuper(SuperCodec.java:1)
	at org.fxmisc.richtext.model.SuperCodec.encode(SuperCodec.java:15)
	at org.fxmisc.richtext.model.ReadOnlyStyledDocument$3.encode(ReadOnlyStyledDocument.java:139)
	at org.fxmisc.richtext.model.ReadOnlyStyledDocument$3.encode(ReadOnlyStyledDocument.java:1)
[...]

PR #578, opened 2017-09-07
fix issue #573: NPE in clipboard copy

JFormDesigner:


JordanMartinez:
I can't recall why I added the second part (maybe I was tired and incorrectly thought the extra condition was needed?)

Thanks for the fix.

Issue #584, opened 2017-09-17
Regression: Image custom object is not properly handled in RichText demo


afester:
* Launch the RichTextDemo
* Insert an image
* Place cursor to start of paragraph
* Press Enter
=> Image is shown twice, once in the upper (new) paragraph and once in the lower paragraph.
When using an older variant of the LinkedImageOps, it works well.

afester:
Modifying realSubSequence in NodeSegmentOpsBase so that it creates an empty segment when the sub sequence length is 0 fixes it:

```
    @Override
    public SEG realSubSequence(SEG seg, int start, int end) {
        if (start == end) {
            return createEmpty();
        }
        return seg;
    }
```
Needs some additional testing before submitting a PR, but I think this should be a safe assumption in any case ...

JordanMartinez:
This issue no longer arises in #590

JordanMartinez:
Closed by #590

Issue #567, opened 2017-08-11
Refactor: Decouple style from segment


JordanMartinez:
The base area's type signature is:
````java
public GenericStyledArea<ParagraphStyle, SegmentType, SegmentStyle>
````
When writing custom objects, we have to write the following code:
````java
public CustomArea extends 
    GenericStyledArea<
        ParagraphStyle, 
        Either<
            TextObject<TextStyle>, 
            CustomObject<TextStyle>
        >, 
        TextStyle
    >
````
You'll notice that the styles are embedded into the custom objects. this doesn't make sense for two reasons: 

1. What if the custom object doesn't need a style (e.g. a hyperlink should always be styled as a hyperlink)

````java
public CustomArea extends 
    GenericStyledArea<
        ParagraphStyle, 
        Either<
            TextObject<TextStyle>, 
            CustomObject<
                TextStyle   // This is useless for the custom object
            > 
        >, 
        TextStyle
    >
````

2. What if one wanted to use a style objects for each object? In other words, `Text` could have `TextStyle` and `CustomObject` could have `CustomObjectStyle`? If so, we'd have to write this hairy code:

````java
public CustomArea extends 
    GenericStyledArea<
        ParagraphStyle, 
        Either<
            TextObject<
                Either<
                    TextStyle, 
                    CustomObjectStyle
                >
            >, 
            CustomObject<
                Either<
                    TextStyle, 
                    CustomObjectStyle
                >
            >, 
        Either<
            TextStyle, 
            CustomObjectStyle
        >
    >
{}
````

I propose we separate the two, so that an segment type no longer knows its own style. Thus, the type signature would become:
````java
public Case1Area extends 
    GenericStyledArea<
        ParagraphStyle, 
        Either<
            // stores the text, nothing more
            Text,
            // stores the displayed text and its actual link, nothing more
            Hyperlink 
        >, 
        // Text will render node using TextStyle. 
        // Hyperlink will ignore its associated TextStyle object 
        //    and set its node's style class to "hyperlink"
        TextStyle 
    >

public Case2Area extends 
    GenericStyledArea<
        ParagraphStyle, 
        Either<
            TextObject, 
            CustomObject
        >, 
        Either<
            TextStyle, 
            CustomStyle
        >
    >

// maybe a class like this that adds helper methods to work with the Eithers?
public GenericStyledEitherArea<PS, Seg1, Seg2, Style1, Style2> extends
[...]

PR #585, opened 2017-09-17
Issue-584: Fixed sub sequence when the resulting sub sequence will be…

afester:
… empty

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/model/SegmentOpsBase.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/model/SegmentOpsBase.java
@@ -46,7 +46,11 @@ public abstract class SegmentOpsBase<SEG, S> implements SegmentOps<SEG, S> {
                     String.format("End cannot be greater than segment's length. End=%s Length=%s", end, length(seg))
             );
         }
-        return seg == empty ? empty : realSubSequence(seg, start, end);
+
+        if (seg == empty || start == end) {
+            return empty;
+        }
+        return realSubSequence(seg, start, end);
     }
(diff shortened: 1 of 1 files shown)


JFormDesigner:
I think this needs more investigation because this PR would undo PR #578 which fixes #573

afester:
Thanks for the link to #573 - will check that ...

JordanMartinez:
The issue that arises both in #573 and #584 is the fact that a style is still associated with an empty segment. Thus, one cannot use a singleton empty segment throughout the area when referring to an empty one because of the style. If you do, it messes up the undo/redo system (e.g. `Unexpected change received: Expected StyledText[text="" style=null] Actual StyledText[text="" style=]`. See also my comments from the Hyperlink demo PR [here](https://github.com/TomasMikula/RichTextFX/pull/550#discussion_r132587029) and [here](https://github.com/TomasMikula/RichTextFX/pull/550#discussion_r132762944)

Again, this is why I'm proposing #567 as then we could have algerbraic data types: an empty and non-empty segment that can be created, subsequenced, etc. without needing the know anything about its style.
The real issue is that `Paragraph` doesn't have a way to know what style it should use when creating an empty segment. If it's the first paragraph in the area, it should default to `initialTextStyle`. If it's in the middle, it should default to the style stored at the last part in the preceding paragraph.

afester:
This requires some more investigation and also integration tests should be provided for #573 and #584 .

JordanMartinez:
@afester Can you write a test that shows the error arising? If you're not familiar with TestFX, you can write a sample JavaFX App and then list comments as to what is supposed to happen once the stage is shown. Then, I can convert it into a TestFX test.

PR #587, opened 2017-09-20
Add API to get paragraph bounds if visible

### LINKED ISSUE #537: Feature: Add API for Paragraph's Bounds
To help implement #77 and #405, there should be some API to calculate a given paragraph's bounds by its index (the index in the list of visible paragraphs and all paragraphs). For example
````java
public final Optional<Bounds> paragraphBounds(int parIndex);
public final Optional<Bounds> visibleParagraphBounds(int parIndex);
````

### PR DESCRIPTION
**Pull request by JordanMartinez**
Resolves #537

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
@@ -1266,6 +1266,58 @@ public class GenericStyledArea<PS, SEG, S> extends Region
         return position(parIdx, 0).toOffset();
     }
 
+    /**
+     * Returns the bounds of the paragraph if it is visible or {@link Optional#empty()} if it's not.
+     *
+     * The returned bounds object will always be within the bounds of the area. In other words, it takes
+     * scrolling into account. Note: the bound's width will always equal the area's width, not necessarily
+     * the paragraph's real width (if it's short and doesn't take up all of the area's provided horizontal space
+     * or if it's long and spans outside of the area's width).
+     *
+     * @param visibleParagraphIndex the index in area's list of visible paragraphs.
+     */
+    public Bounds getVisibleParagraphBoundsOnScreen(int visibleParagraphIndex) {
+        return getParagraphBoundsOnScreen(virtualFlow.visibleCells().get(visibleParagraphIndex));
+    }
+
+    /**
+     * Returns the bounds of the paragraph if it is visible or {@link Optional#empty()} if it's not.
+     *
+     * The returned bounds object will always be within the bounds of the area. In other words, it takes
+     * scrolling into account. Note: the bound's width will always equal the area's width, not necessarily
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by JordanMartinez** [comment]
For now, this will do. If we want to change how the bounds works on short/long paragraphs, we can update it in a later PR.

**Comment by garybentley** [comment]
While useful this only works if the paragraph is already fully visible in the view.  In my case I am trying to write ParagraphGraphicFactory and would like to display an image at an offset from the top of the paragraph.  However the following code:

```
area.setParagraphGraphicFactory (new IntFunction<Node> ()
{

   @Override
   public Node apply (int idx)
   {

       Optional<Bounds> pb  = area.getParagraphBoundsOnScreen (idx);

   }

});
```

will either throw an exception:

> Caused by: java.lang.NullPointerException
>         at org.fxmisc.richtext.GenericStyledArea.getParagraphBoundsOnScreen(GenericStyledArea.java:1372)
>         at java.util.Optional.map(Optional.java:215)
>         at org.fxmisc.richtext.GenericStyledArea.getParagraphBoundsOnScreen(GenericStyledArea.java:1362)

or return `Optional.empty` if the paragraph is only partially visible.

Is there any way to get the paragraph bounds if it is only partially visible?

My use case here is that I have an offset (int) into the text string.  I would like to display an icon to the left of the start of the line where the offset is, thus if I have the following text (a single paragraph):

> To Sherlock Holmes she is always the woman. 
> [*] I have seldom heard him mention her under any other name.
> In his eyes she eclipses and predominates the whole of her sex. It was not that he felt any emotion akin to love for Irene Adler.

I would like to display an icon (for example) at the start of the line "I have seldom..."

If getting the paragraph bounds is not possible, is it possible to get the text offset of the paragraph within the overall text and then the bounds of a particular character(s) within the paragraph?

JordanMartinez:
Could you open a new issue for this? I'd rather continue the conversation there. Once opened, I'll copy/paste my response below:

I'm pretty sure the code does return the bounds of paragraph, even if it's only partially visible. Since we don't have a test verifying that, I could be wrong. 
I'm not sure why it's throwing an NPE. However, it might be due to the area going through an update. If so, that might be the culprit.

> If getting the paragraph bounds is not possible, is it possible to get the text offset of the paragraph within the overall text and then the bounds of a particular character(s) within the paragraph?

You can use `area.getAbsolutePosition(int paragraphIndex, int columnIndex)` and `area.getCharacterBoundsOnScreen(int startPosition, int endPosition)`:
````java
int pos = area.getAbsolutePosition(paragraphIndex, columnIndex)
Bounds b = area.getCharacterBoundsOnScreen(pos, pos + 1);
````
However, since this goes through the same method, I'm not sure if your "partially visible" issue will arise again.

Unfortunately, there's no `area.getAbsolutePosition(int paragraphIndex, int lineIndex, int columnIndex)` method yet.

PR #593, opened 2017-09-26
Allow RichtTextFX to access Java9 hidden methods

imario42:
This pull request let you compile and use RichtTextFX with Java 9.
Methods, where possible, are changed to their public Java 9 counterparts.
Methods still private (only in TextFlowExt) are still used using reflection but in a way which does not even require the hidden classes to be visible.
Thus, you can use RichtTextFx without runtime switches.

### DIFF
--- a/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/hyperlink/HyperlinkDemo.java
+++ b/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/hyperlink/HyperlinkDemo.java
@@ -1,11 +1,14 @@
 package org.fxmisc.richtext.demo.hyperlink;
 
-import com.sun.deploy.uitoolkit.impl.fx.HostServicesFactory;
 import javafx.application.Application;
 import javafx.scene.Scene;
 import javafx.stage.Stage;
 import org.fxmisc.flowless.VirtualizedScrollPane;
 
+import java.awt.*;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.function.Consumer;
 
 /**
@@ -21,7 +24,16 @@ public class HyperlinkDemo extends Application {
 
     @Override
     public void start(Stage primaryStage) {
-        Consumer<String> showLink = HostServicesFactory.getInstance(this)::showDocument;
+        Consumer<String> showLink = (string) -> {
+            try
+            {
+                Desktop.getDesktop().browse(new URI(string));
+            }
+            catch (IOException | URISyntaxException e)
+            {
+                throw new RuntimeException(e);
+            }
+        };
         TextHyperlinkArea area = new TextHyperlinkArea(showLink);
 
         area.appendText("Some text in the area\n");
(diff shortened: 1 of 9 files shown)


JordanMartinez:
Thanks for the PR! 

We'll need to update the build-related files to generate a separate build for Java 9 on Linux and Mac.

However, since everyone is not yet using Java 9, we want to allow RTFX to build on Java 9 and 8 machines alike. This code seems like it will only work on Java 9 if merged.

TestFX encountered the same problem: how can one build the project on both Java versions while maintaining the same code base as much as possible in cases where the API does not change across the Java versions? 

When I wrote TestFX/TestFX#429, I created two packages with the same exact name and type signature. However, their implementations were specific to the Java version. Thus, Gradle would configure the core project in TestFX to use the correct internal package as a dependency based on which Java version was being used in Gradle. Using this API-adapter approach, any future changes to the core project will work in both versions of Java.

I don't know if we can take a similar approach here. It depends on what we need to access and how the type signatures change between the two Java versions. However, if we can, then let's do it that way because everyone is not yet using Java 9.

imario42:
Hi!

Yes, I am aware of that Problem and the PR was merely meant as document if someone else need a Java 9 version of RichtTextFX right now.

I am also playing around with SceneBuilder and would like to also keep the codebase the same for Java 8 + Java 9.

I too would go the adapter way. Having a JavaFxAdapter interface and a JavaFx8AdapterImpl and JavaFx9AdapterImpl in separate projects. Gradle then can choose which to bundle depending on the target Java version.
Using the manifest a factory can then lookup which of those two class to load.

Because I am on vacation for a couple of days now I can not start with that right now.

Best regards,
Mario


Am 26.09.2017 um 20:53 schrieb JordanMartinez <notifications@github.com<mailto:notifications@github.com>>:


Thanks for the PR!

We'll need to update the build-related files to generate a separate build for Java 9 on Linux and Mac.

However, since everyone is not yet using Java 9, we want to allow RTFX to build on Java 9 and 8 machines alike. This code seems like it will only work on Java 9 if merged.

TestFX encountered the same problem: how can one build the project on both Java versions while maintaining the same code base as much as possible in cases where the API does not change across the Java versions?

When I wrote TestFX/TestFX#429<https://github.com/TestFX/TestFX/pull/429>, I created two packages with the same exact name and type signature. However, their implementations were specific to the Java version. Thus, Gradle would configure the core project in TestFX to use the correct internal package as a dependency based on which Java version was being used in Gradle. Using this API-adapter approach, any future changes to the core project will work in both versions of Java.

I don't know if we can take a similar approach here. It depends on what we need to access and how the type signatures change between the two Java versions. However, if we can, then let's do it that way because everyone is not yet using Java 9.

JFormDesigner:
@imario42 I've rebased your PR to master and made it compatible to Java 8 in PR #614 

This works good so far, but I've a question regarding class `GenericIceBreaker`:
What's the reason for checking `delegateMethod.canAccess(delegate)` before invoking `delegateMethod.setAccessible(true)`?

Are there any downsides if I remove this check and always invoke `setAccessible(true)`?

`canAccess()` is not available in Java 8 and I'd like to avoid using reflection for Java 9 if this check is not necessary.

imario42:
Hi!

No need to check for canAccess. I just always do … but in this case it is not required.

Regards,
Mario


Am 15.10.2017 um 13:24 schrieb JFormDesigner <notifications@github.com<mailto:notifications@github.com>>:


@imario42<https://github.com/imario42> I've rebased your PR to master and made it compatible to Java 8 in PR #614<https://github.com/FXMisc/RichTextFX/pull/614>

This works good so far, but I've a question regarding class GenericIceBreaker:
What's the reason for checking delegateMethod.canAccess(delegate) before invoking delegateMethod.setAccessible(true)?

Are there any downsides if I remove this check and always invoke setAccessible(true)?

canAccess() is not available in Java 8 and I'd like to avoid using reflection for Java 9 if this check is not necessary.

JordanMartinez:
Closing since this has been implemented in #614

Issue #594, opened 2017-09-28
Background and underline shapes not properly calculated


afester:
Background and underline shapes are combined even if they are not consecutive.
The SpellChecking demo shows this: not only the invalid words are underlined and highlighted, but all words inbetween also.

JordanMartinez:
Resolved by #595

PR #595, opened 2017-09-28
Issue #594: Properly calculate the background and underline shapes fo…

afester:
…r non-consecutive ranges

* Added an integration test which does some simple styling tests
* The underline test fails without the code modifications in ParagraphText.java
* Fixed the calculation of shapes in ParagraphText.java: shapes must only be combined when the ranges are consecutive

### DIFF
--- a/richtextfx/src/integrationTest/java/org/fxmisc/richtext/style/StylingTests.java
+++ b/richtextfx/src/integrationTest/java/org/fxmisc/richtext/style/StylingTests.java
@@ -1,0 +1,100 @@
+package org.fxmisc.richtext.style;
+
+import static org.junit.Assert.assertEquals;
+
+import java.util.List;
+
+import org.fxmisc.richtext.SceneGraphTests;
+import org.junit.Test;
+
+import javafx.scene.shape.Path;
+import javafx.scene.text.Text;
+
+
+public class StylingTests extends SceneGraphTests {
+
+    private final static String HELLO = "Hello ";
+    private final static String WORLD = "World";
+    private final static String AND_ALSO_THE = " and also the ";
+    private final static String SUN = "Sun";
+    private final static String AND_MOON = " and Moon";
+
+    @Test
+    public void simpleStyling() {
+        // setup
+        interact(() -> {
+            area.replaceText(HELLO + WORLD + AND_MOON);
+        });
+
+        // expected: one text node which contains the complete text
+        List<Text> textNodes = getTextNodes(0);
+        assertEquals(1, textNodes.size());
+
+        interact(() -> {
+            area.setStyle(HELLO.length(), HELLO.length() + WORLD.length(), "-fx-font-weight: bold;");
+        });
+
+        // expected: three text nodes
+        textNodes = getTextNodes(0);
+        assertEquals(3, textNodes.size());
+
+        Text first = textNodes.get(0);
+        assertEquals("Hello ", first.getText());
+        assertEquals("Regular", first.getFont().getStyle());
+
+        Text second = textNodes.get(1);
+        assertEquals("World", second.getText());
+        assertEquals("Bold", second.getFont().getStyle());
+
+        Text third = textNodes.get(2);
+        assertEquals(" and Moon", third.getText());
+        assertEquals("Regular", third.getFont().getStyle());
+    }
+
+
+    @Test
+    public void underlineStyling() {
+
+        final String underlineStyle = "-rtfx-underline-color: red; -rtfx-underline-dash-array: 2 2; -rtfx-underline-width: 1; -rtfx-underline-cap: butt;";
+        
+        // setup
+        interact(() -> {
+            area.replaceText(HELLO + WORLD + AND_ALSO_THE + SUN + AND_MOON);
+        });
+
+        // expected: one text node which contains the complete text
+        List<Text> textNodes = getTextNodes(0);
+        assertEquals(1, textNodes.size());
+        assertEquals(HELLO + WORLD + AND_ALSO_THE + SUN + AND_MOON, 
+                     textNodes.get(0).getText());
+
+        interact(() -> {
+            final int start1 = HELLO.length();
+            final int end1 = start1 + WORLD.length();
+            area.setStyle(start1, end1, underlineStyle);
+
+            final int start2 = end1 + AND_ALSO_THE.length();
+            final int end2 = start2 + SUN.length();
+            area.setStyle(start2, end2, underlineStyle);
+        });
+
+        // expected: five text nodes
+        textNodes = getTextNodes(0);
+        assertEquals(5, textNodes.size());
+
+        Text first = textNodes.get(0);
+        assertEquals(HELLO, first.getText());
+        Text second = textNodes.get(1);
+        assertEquals(WORLD, second.getText());
+        Text third = textNodes.get(2);
+        assertEquals(AND_ALSO_THE, third.getText());
+        Text fourth = textNodes.get(3);
+        assertEquals(SUN, fourth.getText());
+        Text fifth = textNodes.get(4);
+        assertEquals(AND_MOON, fifth.getText());
+
+        // determine the underline paths - need to be two of them!
+        List<Path> underlineNodes = getUnderlinePaths(0);
+        assertEquals(2, underlineNodes.size());
+    }
+}
--- a/richtextfx/src/main/java/org/fxmisc/richtext/CaretPath.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/CaretPath.java
@@ -1,0 +1,10 @@
+package org.fxmisc.richtext;
+
+import javafx.scene.shape.Path;
+
+/**
+ * A path which describes a caret shape in the Scene graph.
+ * 
+ */
+public class CaretPath extends Path {
+}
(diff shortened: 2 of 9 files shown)


richtextfx/src/integrationTest/java/org/fxmisc/richtext/style/StylingTests.java
```diff
@@ -0,0 +1,160 @@
...
+
+import com.nitorcreations.junit.runners.NestedRunner;
+
+import javafx.scene.Node;
+import javafx.scene.layout.Region;
+import javafx.scene.shape.Path;
+import javafx.scene.text.Text;
+import javafx.scene.text.TextFlow;
+
+@RunWith(NestedRunner.class)
```
JordanMartinez: NestedRunner is only necessary if you want to use a class to set up common things among multiple tests. Since you're not doing that here, you can remove it.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/style/StylingTests.java
```diff
@@ -0,0 +1,160 @@
...
+
+    /**
+     * @return A list of text nodes which render the current text.
+     */
+    private List<Text> getTextNodes() {
+        // get the ParagraphBox (protected subclass of Region) 
+        Region paragraphBox = getParagraphBox(0);
+
+        // get the ParagraphText (protected subclass of TextFlow) 
+        TextFlow tf = (TextFlow) paragraphBox.getChildrenUnmodifiable().get(0);
```
JordanMartinez: Could this be changed to check for the instance first, only cast if it's valid and throw an exception if it's not? Otherwise, I am not sure it will be the first one if a selection is made.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/style/StylingTests.java
```diff
@@ -0,0 +1,160 @@
...
+    @Test
+    public void simpleStyling() {
+        // setup
+        interact(() -> {
+            area.replaceText(firstWord + styledWord1 + remainingLine);
+        });
+
+        // expected: one text node which contains the complete text
+        List<Text> textNodes = getTextNodes();
+        assertEquals(1, textNodes.size());
```
JordanMartinez: Perhaps this should be its own test?

richtextfx/src/integrationTest/java/org/fxmisc/richtext/style/StylingTests.java
```diff
@@ -0,0 +1,160 @@
...
+
+        // Set word "World" to bold
```
JordanMartinez: The comment forces me to look up what String object is equated with "World".

richtextfx/src/integrationTest/java/org/fxmisc/richtext/style/StylingTests.java
```diff
@@ -0,0 +1,160 @@
...
+import javafx.scene.text.TextFlow;
+
+@RunWith(NestedRunner.class)
+public class StylingTests extends InlineCssTextAreaAppTest {
+
+    private final static String firstWord = "Hello ";
+    private final static String styledWord1 = "World";
+    private final static String moreText = " and also the ";
+    private final static String styledWord2 = "Sun";
+    private final static String remainingLine = " and Moon";
```
JordanMartinez: Relating to line 89's comment, since the string itself does not contain the style, it should equal to what the text it is assigned is. Otherwise, `styledWord1` is unstyled in one moment and styled in a later moment.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/style/StylingTests.java
```diff
@@ -0,0 +1,160 @@
...
+        assertEquals("Regular", third.getFont().getStyle());
+    }
+
+
+    @Test
+    public void underlineStyling() {
+        // setup
+        interact(() -> {
+            area.replaceText(firstWord + styledWord1 + moreText + styledWord2 + remainingLine);
+            //               "Hello World and also the Sun and Moon"
```
JordanMartinez: the comment isn't necessary if we rename the strings to the text they represent

richtextfx/src/integrationTest/java/org/fxmisc/richtext/style/StylingTests.java
```diff
@@ -0,0 +1,160 @@
...
+        interact(() -> {
+            final int start1 = firstWord.length();
+            final int end1 = start1 + styledWord1.length();
+            area.setStyle(start1, end1,
+                          "-rtfx-underline-color: red; -rtfx-underline-dash-array: 2 2; -rtfx-underline-width: 1; -rtfx-underline-cap: butt;");
+
+            final int start2 = end1 + moreText.length();
+            final int end2 = start2 + styledWord2.length();
+            area.setStyle(start2, end2,
+                          "-rtfx-underline-color: red; -rtfx-underline-dash-array: 2 2; -rtfx-underline-width: 1; -rtfx-underline-cap: butt;");
```
JordanMartinez: Could this be declared once in a variable and then just pass that variable in to both `area.setStyle()` calls?

richtextfx/src/integrationTest/java/org/fxmisc/richtext/style/StylingTests.java
afester: True. It was essentially a preparation for later enhancements, but it is indeed better to add it once it is really required.

afester:
> Won't this return the background shapes and border shapes as well if we add more tests that test for those things, too.

Yes, that is true. In general those methods rely on a particular order how the nodes are inserted into the text flow. It would be better if the node objects themselves know "what" they are. The straight forward solution would be to subclass Path as CaretPath, SelectionPath, BackgroundPath and UnderlinePath - and then use the sub classes instead of plain Path. This would make debugging and inspection of the scene graph (and writing scene related tests) MUCH simpler, since it is immediately visible what kind of path that particular node represents. Alternatively, we could set the id with a specific prefix for each path, but that would require unique id generation (like "bg1,", "bg2" etc.) and it would refer to the instance, not type. Finally, we could also subclass Path once and user a getter to retrieve the type attribute, something like

```
Path underlinePath = new TypedPath(PATHTYPE.UNDERLINE);
...
if (underlinePath.getType() == PATHTYPE.UNDERLINE) { ...}
```

richtextfx/src/integrationTest/java/org/fxmisc/richtext/style/StylingTests.java
afester: True. get(0) is not always correct - I need to iterate over the children and look up the TextFlow (for which there should be only one, though),

richtextfx/src/integrationTest/java/org/fxmisc/richtext/style/StylingTests.java
JordanMartinez: I like NestedRunner because it reduces code duplication when setting up similar tests. However, it doesn't allow one to run tests within the nested classes in my IDE. I'm not sure how to fix that, but even if we did, I would think that such a thing should not need to be explained for the tests to work on someone else's computer.

JordanMartinez:
Very good points. I definitely like the subclass or path type approach better than setting the ID. However, which would be easier to use when writing tests and would scale well if we need to add more components to it?

JordanMartinez:
> However, which would be easier to use when writing tests and would scale well if we need to add more components to it?

Actually, since we may never add more rtfx-specific components (YAGNI principle), which approach (and its API) will make reading through the tests easy?

JordanMartinez:
Looks like #596 is going to work soon. Once it gets merged, can you rebase this (with the changes mentioned) on top of that PR's work?

JordanMartinez:
Alright #596 has been merged. Looking forward to your update.

afester:
Updated PR primarily for review of the path subclass approach.
The one failed Travis test seems to be 
org.fxmisc.richtext.api.CaretTests > caret_bounds_are_absent_after_moving_caret_without_following_it FAILED
java.util.concurrent.TimeoutException

JordanMartinez:
I still see your original commits at the start of this PR and a whole lot of "commit noise" from my other PR. Did you rebase your commits or is this a GitHub issue?

Yeah... that test apparently failed in my PR, too. However, Travis had some backlog and it didn't report the issue correctly or something, so I thought it passed. They emailed me the failure later.

The issue is that the first test fails, irregardless of which one is run first in that class. I'm not sure where it's timing out thought. Might need to add `TestLoggingEvent.STANDARD_ERROR` to the build.gradle's integrationTest.testLogging closure.

I'll look over the code later on this week (got other things to do today unfortunately...). However, could you move the test itself to `integrationTest`?

afester:
> I still see your original commits at the start of this PR and a whole lot of "commit noise" from my other PR. Did you rebase your commits or is this a GitHub issue?

Yes, I rebased but forgot that the rebased local branch needs to be force-pushed :1st_place_medal:
I just re-did the rebase, looks much better now...

> However, could you move the test itself to `integrationTest`?
The tests are in the `src/integrationTest` directory - I suppose that this is the right location?

richtextfx/src/main/java/org/fxmisc/richtext/CaretPath.java
```diff
@@ -0,0 +1,10 @@
+package org.fxmisc.richtext;
+
+import javafx.scene.shape.Path;
+
+/**
+ * A path which describes a selection shape in the Scene graph.
+ *
```
JordanMartinez: "selection" is written instead of what you actually meant: "caret."

Otherwise, everything looks good!

JordanMartinez:
Ok. The Mac build issue has been resolved now.

JordanMartinez:
@afester Can you fix the javadoc issue and rebase on top of master? Then the mac build should pass.

JordanMartinez:
Thanks for your work!

PR #600, opened 2017-09-30
Feature: map visible paragraph index to all paragraph index & vice versa

### LINKED ISSUE #599: Feature: Provide mapping between all-paragraph and visible-paragraph index systems
Given an index in `area.getParagraphs()`, one should be able to map it to its respective `area.getVisibleParagraphs()` index. Likewise, the index of the latter should be mappable to the index of the former. Thus, we should add an API like
````java
public final Optional<Integer> allParToVisibleParIndex(int allParIndex);
public final int visibleParToAllParIndex(int visiblePar);
````

JordanMartinez:
Resolves #599


JordanMartinez:
AppVeyor isn't set up yet, but Travis built fine.

PR #601, opened 2017-09-30
Cleanup integration test

JordanMartinez:
To summarize:

- Splits `NavigationTests` into their own separate files for easier management
- Creation of new test base (`RichTextFXTestBase`) in case one needs to use a different area flavor (e.g. `GenericStyledArea`) while still having access to OS-related things and position helper methods
- Prevent possible interference with other tests (Only set up context menu in `ContextMenuTests`, clear clipboard before tests run)
- Optimize tests' string creation via `StringBuilder` to only do so once
- Use more accurate way of determining caret position via `getCharacterBoundsOnScreen`
- Rename nested classes and test methods to use "_" as a space-like character for easier readability
- No longer use glass robot by default; CI builds will determine this now
- Moved some tests to more appropriate files with related tests
(see commit messages for full list of changes)


JordanMartinez:
I'm out of time for today, but I am going to see if the Mac build will pass if I run the HitTests tests that currently fail on Mac now that we have a more accurate way of determining the caret's paragraph

JordanMartinez:
@JFormDesigner Per your comments [here](https://github.com/FXMisc/RichTextFX/pull/576#issuecomment-328888327), I've rewritten some tests to use the `getCharacterBoundsOnScreen` API to more accurately test the caret's position.

JordanMartinez:
AppVeyor is still not set up yet, but the tests so far pass on Travis and Mac (aside from the two bugs discovered: #602 and #603)

JordanMartinez:
@afester I appreciate that you submitted the PR, but I decided to revert what you did in #589 (the glass adapter) in this PR. We may need to revert my change, so that it's back to yours (at least when Gradle detects that Windows is running it). I hope to know more once I get the Windows CI set up properly.

Issue #606, opened 2017-10-06
How best to construct a ParagraphGraphicFactory that allows for dynamic positioning of Nodes


garybentley:
This is a continuation and broadening of the question in: https://github.com/FXMisc/RichTextFX/pull/587

What is the best way to create a ParagraphGraphicFactory that allows for Nodes to be dynamically positioned depending upon the vertical position/offset of a piece of text at a particular index/offset into a paragraph.

For example in the following block of text:

> To Sherlock Holmes she is always the woman. I have seldom heard him mention her under any other name. In his eyes she eclipses and predominates the whole of her sex. It was not that he felt any emotion akin to love for Irene Adler. All emotions, and that one particularly, were abhorrent to his cold, precise but admirably balanced mind.

If I wish to add a Node (some kind of visual indicator) against the text "It was not that he felt" in the paragraph graphic to the left I could do:

```
       area.setParagraphGraphicFactory(new IntFunction<Node> ()
       {

            @Override
            public Node apply (int idx)
            {

                Pane pane = new Pane ();
                pane.setPrefWidth (50);
                pane.setStyle ("-fx-background-color: yellow; -fx-border-width: 1; -fx-border-color: green;");

                try
                {

                    int ppos = area.getAbsolutePosition (idx, 0);
                    Optional<Bounds> posb = area.getCharacterBoundsOnScreen (ppos, ppos);
                    Optional<Bounds> charb = area.getCharacterBoundsOnScreen (ppos + 150, ppos + 151);

                    if (charb.isPresent ())
                    {

                        Label l = new Label ("=>");
                        pane.getChildren ().addAll (l);
                        l.relocate(10, charb.get ().getMinY () - posb.get ().getMinY ());

                    }

                } catch (Exception e) {
                    // Ignore.
                }

                return pane;

            }

        });

```

This works relatively well for determining the y offset for positioning, the 150 is a nominal value used as an example, however there are a few issues:

 1. On first adding the quoted text via pasting into a GenericStyledArea wrapped in a VirtualizedScrollPane (I just modified the RichText.java example), I get the image below:

![example1](https://user-images.githubusercontent.com/921991/31261194-0bc074d6-aa96-11e7-81ea-c4ea8e9267bd.png)

It's only when I move paragraph 0 to paragraph 1 by adding a new line above do I get 
![example2](https://user-images.githubusercontent.com/921991/31261230-387ccc04-aa96-11e7-881b-b0424ba18f7b.png)

[...]

PR #607, opened 2017-10-06
Expose viewport dirty as public API

JordanMartinez:
Addresses an issue raised in #606


JordanMartinez:
Huh... I'm surprised the Mac build didn't fail with the first `CaretTest` as reported in #608...

PR #609, opened 2017-10-06
Fix Mac build error

### LINKED ISSUE #608: First test run on Mac fails with TimeoutException
This first arose in #596, but was not reported as such due to a backlog in Travis CI at the time. It reappeared in #595

JordanMartinez:
Resolves #608


JordanMartinez:
Reducing the number of lines fixed the timeout exception (not sure why that would cause a timeout in the first place, but oh well...)

PR #610, opened 2017-10-12
Text position tracking demo

garybentley:
This isn't a full solution for tracking changes.  Ideally the positions would be tracked within the StyledDocument itself and a reference given out to the caller.  However I consider that to be beyond the scope of what I'm trying to achieve so I didn't go to that level.  This example is more designed to provoke discussion on approaches.

Only a single class is used with sub-classes for the various parts, such as TextPosition, Indicator and Margin.  This is to allow for easy navigation and understanding of the example.  I can decompose the example into full classes if necessary.

I've added a method "modelToView" which is the method that was used in the Swing JTextComponent.  This translates a text offset position (model) to a bounds within the area (view).  This does require a different handling of new lines to how getCharacterBoundsOnScreen handles it.  For my purposes a user clicking on an empty paragraph or end of a line is a valid location.

The "Margin" class is there as a visual example of displaying the tracked locations.  The tracking of changes to the viewport, bounds and richChanges stream can be moved around as necessary.  There isn't a "good" place for these, the main aim to is prevent a flood of changes requesting layouts.


JordanMartinez:
Thanks for the PR. I think this shows that we should add a few more guidelines to submitting PRs to the demo package. Could you update your code to do the following:

- Put this demo into its own package
- Add the suffix "Demo" to your demo, so that people know which class is the demo and which classes are things used eventually in the demo
- Modulate the code a bit more by converting some things into their own classes.

Also, I'm actually finding it rather difficult to read your code because of all the additional white space before/after curly braces, before a method's enclosing `()`, etc. I get that that is a valid style for some, but would you mind adjusting the code to be more like the other demos? None of the other demos follow that style, and I'd like to keep things that way for now.

Lastly, I'm making this comment just after glancing through your code. I haven't read through it carefully yet mainly because the extra whitespace was making it harder for me to read :sweat_smile:

JordanMartinez:
I don't fully understand how your code works yet, nor why it works _that_ way as opposed to another way. My hunch says the code could be better organized or structured, but I don't know why yet.

There are a number of things to address here:
- the demo should explain briefly what the user is supposed to do when they first start it up. It took me a while of reading through the code before I realized that I was supposed to click on the area somewhere in the text and then the indicator would appear.
- you should be using `InlineCssTextArea` because it eliminates the generics "noise" in `GenericStyledArea`.

As for the code itself, I noticed a few things you might want to improve:
- Margin checks three conditions that could be moved into one filter stream when subscribing to the area's rich changes event stream. For example, you basically write 
````java
stream
    .filter(condition && condition)
    .subscribe (i -> {
        if (condition) { // why isn't this in the filter check?
            doStuff()
        }
    );
````
- the aforementioned checks above are `event position >= i position && event position <= i position && event.getNetLength() < 0`. The first two lead to the same result as `event position == i position`.

garybentley:
The demo does do that.  In the class comment of TextPositionTrackingDemo it states that the user should click to add a position and right click to remove it.

It works the way it does because it is the fastest way I could do it.  It is not designed to be a complete solution or the "best" way as I describe in the comments in Margin.  It is supposed to be illustrative.  The best method, as I see it, would be to add a tracked position to the area itself and it keeps track of the position movements.  The demo is there to show a minimum change solution with next to no changes required to the core RichTextFX code.

I missed the extra filter check, my oversight.  The demo is a long way behind my actual code.

I'll change it to InlineCssTextArea however please bear in mind that this demo is now taking up considerably more time than I want to give it or have (which is why I was wanting to add it as an issue with suggested code rather than a PR).

garybentley:
I've added an initial comment to the text area to give the user an idea of what they should do to interact with the demo.

JordanMartinez:
> The demo does do that. In the class comment of TextPositionTrackingDemo it states that the user should click to add a position and right click to remove it.

My apologies. I could have been much clearer about what I meant. I meant it would be much clearer for the user if such a person simply started up the demo and there was a label of text above/in the area that told the user what to do, so that the user could see what the code does immediately via the demo, and then look at the code later to see how it specifically works. I see that you have now done something like that in the last commit you pushed.

> It works the way it does because it is the fastest way I could do it. It is not designed to be a complete solution or the "best" way as I describe in the comments in Margin. It is supposed to be illustrative. 

My apologies again. I read through the comments in an earlier sitting. In a later sitting, I read through just the code and then made my above comments. So, that point wasn't as fresh in my memory anymore, and I believe I had different expectations as a result.

> please bear in mind that this demo is now taking up considerably more time than I want to give it or have (which is why I was wanting to add it as an issue with suggested code rather than a PR).

Yes, and I thank you for your time. Looking at the code, I need to think on whether `getRangeBoundsOnScreen` should be made public or not. If it shouldn't be, either because it shouldn't be or because the quirk in the code that you're trying workaround should be fixed in a better long-term way, then that might require more time and changes than you would be willing to contribute (and I don't blame you if you refused). At the same time, one needs to ask how clear the code is in terms of demonstrating what you are trying to do. Although "it works," does it demonstrate how one might/should implement the code so that the desired result is achieved?

At the very least, even if this PR is not merged, its existence does what you are trying to achieve: illustrate a way to achieve the desired result. 
In the middle, it could be merged because it does work and does demonstrate a way to do something. By proving that such a thing can be done, one can use it as a starting point and tweak it to their liking later.  
At the very most, I could continue working on it past this point (whether that be before it gets merged or afterwards), so that the code is cleaned up a bit more without taking up more of your time and that you still receive credit for your work.

garybentley:
Thanks, to be clear, I DON'T think that position tracking should be handled in this way.  Personally I'd like to see it tracked within the area or document.  I'm just interested here in whether I can solve the problem and get the feature I want.  As a developer of an application myself I understand you have different priorities to me so I'm not going to just ask for the feature(s).  I'm also well aware that you are looking for a new maintainer of RichTextFX so you have (probably) little interest in adding new features, especially ones you can't see a benefit of.

In my ideal world I would ask for a position from the area given a text offset, i.e.

`public TextPosition trackPosition(int)`

the implementation detail of how it is updated is moot to me.  However I don't want to be rummaging around in the internals of RichTextFX since that's not my area.  If I'm pushed, I'd say that a `TextPosition `object should subscribe to changes to the parent document/area and then keep itself updated.  

To give some context to this, I'm coming here from the Swing world (pity me), so I'm essentially looking for specific features in RichTextFX that are available within `JTextComponent `and `Document `(or more specifically how I can implement them myself).  The one great advantage `JTextComponent `has is that it allows for easy translation between the model and the view, hence my naming of the `getCharacterBoundsOnScreen `clone `modelToView`.  `JTextComponent `also has a `viewToModel `method that let's you move from x,y coordinates to an offset (think the `area.hit()` method). 
 Similarly the `Document `class has the method `createPosition(int)` which returns a Position object that tracks the offset within the document.  The details of how the offset is tracked are hidden.

Also, I have no interest in taking any credit :)  You guys have done all the hard work, my crappy demo is not worth any credit, it is merely a proof-of-principle example, i.e. that it can be done, not how it should be done.

JordanMartinez:
> Also, I have no interest in taking any credit :) You guys have done all the hard work, my crappy demo is not worth any credit, it is merely a proof-of-principle example, i.e. that it can be done, not how it should be done.

It's still farther than any of us have gone in trying to solve this issue :wink: 

> To give some context to this, I'm coming here from the Swing world (pity me), so I'm essentially looking for specific features in RichTextFX that are available within JTextComponentand Document(or more specifically how I can implement them myself). The one great advantage JTextComponenthas is that it allows for easy translation between the model and the view, hence my naming of the getCharacterBoundsOnScreenclone modelToView. JTextComponentalso has a viewToModelmethod that let's you move from x,y coordinates to an offset (think the area.hit() method).

Perhaps we should include a table in the wiki that shows new users of this project what the JavaFX counterpart in RichTextFX is for a Swing API? It may also help us in knowing what other features could be implemented in this project.

>  I'm also well aware that you are looking for a new maintainer of RichTextFX so you have (probably) little interest in adding new features, especially ones you can't see a benefit of.

I can see the benefit of it. However, RTFX has become more of something that distracts me from continuing the project that initially led me to contribute to this project. I like working on RTFX, but if I keep doing so, I'll never have time to work on my own stuff.

garybentley:
A table is an excellent idea.

Something like:

| Feature | Swing | RichTextFX |
| ------- | ----- | ---------- |
| Offset tracking | JTextComponent().getDocument().createPosition(int) | GenericStyledArea.trackPosition(int) |
| Translate model offset to view bounds | JTextComponent().modelToView(int) | GenericStyledArea.getCharacterBoundsOnScreen(int) |

JordanMartinez:
Closing due to inactivity.

PR #617, opened 2017-10-15
moved ParagraphText specific code from createStyledTextNode() methods to class ParagraphText

JFormDesigner:
The moved code is IMHO an implementation detail that belongs to class `ParagraphText` and not to demos.

As stated in [this comment](https://github.com/FXMisc/RichTextFX/pull/614#issuecomment-336666986) I'm still not sure whether binding `TextExt.selectionFill` to `TextExt.fill` makes any sense. I can't see any difference without this binding. In PR #428 I've also removed this binding and replace with binding `TextExt.selectionFill` to `ParagraphText.highlightTextFill`, but as long as the JDK is not fixed, styling text color of selection does not work at all...


JordanMartinez:
Thanks for the PR.

JordanMartinez:
I forgot to check whether you fix this in the demos as well. I don't think you do in this PR. Sorry about that. That's on me, not you.

JordanMartinez:
Nevermind the earlier comment. By putting the code in `ParagraphText`, this affects every area. Sorry once again!

PR #618, opened 2017-10-16
First iteration to have code compile on both Jdk8 and 9

afester:
This is work in progress.
Opened for further discussion
**Do not merge.**

This approach runs through the Java8 and Java9 compiler using the grade build. It uses two separate hierarchies for Java8 and Java9 specific files. The gradle build automatically detects the used JDK and selects the appropriate source folder.

Pending:
* Integration tests fail with Java9
  * Need to replace testfx-internal-java8 with testfx-internal-java9 - after doing this manually, only 11 tests still fail (108 Runs/10 Skipped/11 Failures)
* Demos not yet tested
* Source files are duplicated between Java8 and Java9. Need to pull out the common code.
* Eclipse integration ("gradle eclipse") does not seem to select the additional source folder yet

Later, this approach could be extended to support the Java9 multi release jar approach for packaging by moving the created class files into the appropriate folders below META-INF.


JordanMartinez:
This relates to #270 (adding this comment so it shows up there, too).

afester:
Already done with #614

PR #639, opened 2017-11-03
Add JabRef to the projects using RichTextFX

koppor:
JabRef is a bibliography management suite currently migrating to JavaFX. Our last published version uses RichTextFX.


koppor:
Refs #640

JordanMartinez:
Thanks for the PR.

PR #645, opened 2017-11-17
Allow initializing an [Editabled/ReadOnly]StyledDocument with content

### LINKED ISSUE #644: Allow ReadOnlyStyledDocument to be created from list of paragraphs
ROSD assumes that it will be created only via another document (subdocument, append/prepend, etc.) or if it only contains one segment.

Add support for creating a ROSD with a list of paragraphs of content.

### LINKED ISSUE #646: Allow easier construction of a ReadOnlyStyledDocument with content
In tandem with #644, we should be able to easily construct a list of `Paragraph`s that can be passed into a ROSD. Something like...
````java
ParagraphListBuilder<PS, SEG, S> builder = new ParagraphListBuilder<>(segmentOps, defaultParagraphStyle);
List<Paragraph<PS, SEG, S> list = builder
   // each add paragraph method is a wrapper around the `Paragraph` constructor
   // but ignores the segmentOps part since this should be the same object 
   // for the entire StyledDocument
   .addParagraph(SEG seg, S style)
   // to avoid passing in a paragraph style each time, we can ignore it
   // and use the default paragraph style
   .addParagraph(List<SEG> segs, StyleSpans<S> spans)
   // in cases where it's desired, it can still be added
   .addParagraph(overridingParagraphStyle, List<StyledSegmetn<SEG, S>> list)

   // when finished, we can create it
   // only one builder can be used per list
   .build();
````

JordanMartinez:
Resolves #646 

No longer adddresses #644

Before this, the initial document must always be empty

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/model/GenericEditableStyledDocument.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/model/GenericEditableStyledDocument.java
@@ -13,6 +13,23 @@ package org.fxmisc.richtext.model;
 public final class GenericEditableStyledDocument<PS, SEG, S> extends GenericEditableStyledDocumentBase<PS, SEG, S>
         implements EditableStyledDocument<PS, SEG, S> {
 
+    /**
+     * Creates an {@link EditableStyledDocument} with the given document as its initial content
+     */
+    public GenericEditableStyledDocument(ReadOnlyStyledDocument<PS, SEG, S> initialContent) {
+        super(initialContent);
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


JordanMartinez:
By following the idea I listed in #644 by exposing ROSD's main constructor (it takes a List of paragraphs as its argument), it allows a developer to pass in paragraphs that don't all use the same `SegmentOps` object, which could lead to bugs later on. By adding a builder that handles that more appropriately (and resolves #646), this isn't an issue.

JordanMartinez:
I've updated the snapshot version to include this PR.

Issue #650, opened 2017-11-20
Read only pseudo class is misspelled in GenericStyledArea


Jugen:
The read only pseudo class parameter in [GenericStyledArea](https://github.com/FXMisc/RichTextFX/blob/ed58d3181898a54682da3151b07eb451578d52aa/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java#L297) is currently `read-only` but to be consistent with CSS convention should be `readonly` (i.e. without the hyphen).

The misspelling causes the RichTextFX area to not respond to the CSS `:readonly` pseudo‑class.

JordanMartinez:
That's my fault. It originally did not have the hyphen, but I added it because I thought it was breaking convention. Mind submitting a PR for that?

PR #651, opened 2017-11-20
Corrected misspelled read-only pseudo class

Jugen:
Fix for #650 (Read only pseudo class is misspelled in GenericStyledArea)


JordanMartinez:
I'll update the snapshot in just a few minutes

Issue #653, opened 2017-11-21
Regression of properties that could be set in FXML, fail in 0.8.1


Jugen:
Before RichTextFX 0.8.1 properties like wrapText and editable could be set in FXML, but they now fail :-(

The cause of this is that these properties were made **default** methods in [ViewActions.java](https://github.com/FXMisc/RichTextFX/blob/8619544311d39ea108fcfc6a5e1b07a6ba9027e3/richtextfx/src/main/java/org/fxmisc/richtext/ViewActions.java#L42), whereas previously they were implemented in [GenericStyledArea.java](https://github.com/FXMisc/RichTextFX/blob/v0.8.0/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java#L308)

JordanMartinez:
Dang... I did that so the `GenericStyledArea` file was cleaner and wasn't as long. I didn't know it would produce that affect. sigh...

Are there other properties besides these two that should also be moved back into GSA?

Jugen:
Here is the full list of properties affected:

> editable
> wrapText
> autoScrollOnDragDesired
> 
> contextMenuObject
> contextMenuXOffset
> contextMenuYOffset

BTW if the following methods had `EventHandler<MouseEvent>` as a parameter instead of `Consumer<MouseEvent>` then they would also be accessible from FXML.

> onOutsideSelectionMousePress
> onInsideSelectionMousePress
> onNewSelectionDragEnd
> onSelectionDrop

JordanMartinez:
Mind submitting a PR for that? And let's include the change from `Consumer` to `EventHandler`

JordanMartinez:
Closed by #654. I've also opened #655 to address the second part of this issue.

PR #654, opened 2017-11-22
Add support for FXML (again)

Jugen:
Fixed regression part of #653


JordanMartinez:
[Looks like you'll need to revert the default methods back to regular interface method declarations](https://ci.appveyor.com/project/JordanMartinez/richtextfx/build/master%20109/job/sgoq984fdhy6jttr#L57)

Could you add a test in this PR, too, that initializes an area with all the FXML configurations? That should insure this doesn't get broken in the future. I'm not sure whether such a test should go in the regular `test` package or in the `integrationTest` package.

Jugen:
I've never submitted a test before, so need some guidance:
1. Does the test need to be registered somewhere or are all class files that have a main method in `org.fxmisc.richtext` automatically executed ?
2. Do you have to have a method annotated with @Test ?

JordanMartinez:
Mm... yeah I guess this should go in the `integrationTest` package. I suppose the test class should also go in the `api` package within `integrationTest` since this doesn't fall into behavior-related tests, nor is it expansive enough to need its own package.

So, create a new class that extends `RichTextFXTestBase` (or whatever it's called).
Add a start method that is similar to `InlineCssTextAreaAppTest`'s start method, but without the area or its helper methods.
Add a method that has a name format like "tests_whether_fxml_construction_of_area_works()" that is annotated with `@Test`. In its body, you could use the FXML loader to create an area, which is then assigned to the stage's root. Then, you could assert that certain properties are true (e.g. it has a context menu, it's text is wrapped, etc.)

In case you're wondering, some test classes are annotated with `@RunWith(NestedRunner.class)` because it allows one to set up contexts that get shared between multiple tests. Since this is a single test, it's not needed.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FxmlTester.java
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    @Test
+    public void test_fxml_construction_of_area()
+    {
+        Object obj = null;
+        // FxmlTest.fxml is located in resources folder:
+        // src/integrationTest/resources/org/fxmisc/richtext/api/
+        try	{ obj = FXMLLoader.load( getClass().getResource( "FxmlTest.fxml" ) ); }
+        catch ( Exception EX ) { EX.printStackTrace(); }
```
JordanMartinez: There shouldn't be a catch here. Rather, the test method should propagate the thrown exception:
````java
@Test
public void testSomething() throws IOException {
// no longer need the try or catch statements
// try { 
    FXMLLoader.load();
// } catch (IOException EX) {}
````
Also, it should throw the exception that FXMLLoader.load will throw, not the base Exception class upon which all other exceptions are based.

File: richtextfx/src/integrationTest/resources/org/fxmisc/richtext/api/FxmlTest.fxml [hunk 2]
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<?import javafx.scene.layout.VBox?>
+<?import org.fxmisc.richtext.StyleClassedTextArea?>
+<?import javafx.scene.control.ContextMenu?>
+<?import javafx.scene.control.MenuItem?>
+
+<VBox xmlns:fx="http://javafx.com/fxml/1">
```
JordanMartinez: To prevent noise, it should return a single area, not a VBox of two areas.

richtextfx/src/integrationTest/resources/org/fxmisc/richtext/api/FxmlTest.fxml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<?import javafx.scene.layout.VBox?>
+<?import org.fxmisc.richtext.StyleClassedTextArea?>
+<?import javafx.scene.control.ContextMenu?>
+<?import javafx.scene.control.MenuItem?>
+
+<VBox xmlns:fx="http://javafx.com/fxml/1">
+    <StyleClassedTextArea editable="true" wrapText="true" autoScrollOnDragDesired="true" />
+    <StyleClassedTextArea contextMenuXOffset="12.0" contextMenuYOffset="34.0" >
```
JordanMartinez: These two sets of properties should be combined into the same area

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FxmlTester.java
```diff
@@ -0,0 +1,26 @@
...
+        // Nothing needed here
+    }
+
+    @Test
+    public void test_fxml_construction_of_area()
+    {
+        Object obj = null;
+        // FxmlTest.fxml is located in resources folder:
+        // src/integrationTest/resources/org/fxmisc/richtext/api/
+        try	{ obj = FXMLLoader.load( getClass().getResource( "FxmlTest.fxml" ) ); }
```
JordanMartinez: This should return an area (you'll need to cast the returned object into what it actually is) so we can assert that its properties are what they were specified to be in the FXML document.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FxmlTester.java
```diff
@@ -0,0 +1,26 @@
...
+        catch ( Exception EX ) { EX.printStackTrace(); }
+        org.junit.Assert.assertNotNull( obj );
```
JordanMartinez: We should assert that the properties values are what we specified them to be.

Jugen: If FXMLLoader cannot set a property it will throw an exception, so if no exception is thrown then it's **guaranteed** to have been set :-)

richtextfx/src/integrationTest/resources/org/fxmisc/richtext/api/FxmlTest.fxml
Jugen: I can do that, I did it this way because I thought its easier to read than having one long line of properties :-)

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FxmlTester.java
JordanMartinez: Good to know. Mind writing that in as a comment for those of us who don't use it that often?

richtextfx/src/integrationTest/resources/org/fxmisc/richtext/api/FxmlTest.fxml
JordanMartinez: Does it need to be a one-liner? Could you do something like
````xml
<StyleClassedTextArea
    editable="true"
    wrapText=true"
    ... 
    <contextMenu>...</contextMenu>
</StyleClassedTextArea>

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FxmlTester.java
Jugen: Will do ....

Jugen:
> Were you planning to change the Consumer objects to EventHandler objects, too?

Yes, I've done that in a separate branch. I was going to do a PR once this one is done ?

JordanMartinez:
> Yes, I've done that in a separate branch. I was going to do a PR once this one is done ?

Ok. Yeah, I guess that falls into a slightly different scope that this one despite still being FXML related. We should open a separate issue for that, too, so that it shows up in the generated changelog.

JordanMartinez:
Is there a reason for wrapping the area in a VBox?

Jugen:
It could have been any `Pane` subclass (think Swing layout manager). Originally I had it as a `BorderPane`, until I wanted to add multiple text areas and made it a `VBox` for simplicity.  Technically it's possible for it to be a `Control`, although it doesn't usually make sense.

Bear in mind that the FXML file is still going to change in order to accommodate testing the `EventHandlers`, so if you don't mind I'd prefer to keep it as is for now. We can look at changing it then, if you still think it would be better.

JordanMartinez:
Looks good! Thank you for your work.

JordanMartinez:
I just realized that I forgot to ask you to squash your commits into as few as possible before merging... My bad.

PR #663, opened 2017-12-30
Replaced Panel with Toolbar.

scientificware:
Some small changes in RichText.java to test a pull request.

screenshot_richtextfx


JordanMartinez:
Looks like the build.gradle file needs to be updated since its not properly detecting the Java 9 version

scientificware:
It seems that we have to update Gradle to 4.2.1 or higher if the problem persists :
[Discussion about this bug report ](https://discuss.gradle.org/t/could-not-determine-java-version-from-9-0-1/24457)
[Gradle's blog about Java 9 support](https://blog.gradle.org/java-9-support-update)

They suggest to modify the line :
distributionUrl=https\://services.gradle.org/distributions/gradle-4.2.1.zip
in  RichTextFX/gradle/wrapper/gradle-wrapper.properties

JordanMartinez:
Looks like it. Would you mind changing that and re-generating the wrapper files?

JordanMartinez:
Two things:
- please include the additional change you made in the button's width in the original toolbar commit
- please change the commit message for the second commit to explain what you are doing in that commit (e.g. upgrade Gradle to 4.4.1)

We'd like to keep these commits distinct and clear.

JordanMartinez:
Meaning, can you squash this PR into only two commits: the first being the toolbar change and the second being the gradle upgrade and hash change?

JordanMartinez:
This was reconfigured and merged in #664. Closing.

PR #665, opened 2018-01-02
 New sample file compatible with richtext 0.81

scientificware:
Just a new version of the sample.rtfx file.


JordanMartinez:
Could you squash the commit history down into 2 commits: one for the demo screenshot part itself on the readme and adding the file itself and the second modifying the sample rich text file?

Also, could the demo screenshot's file name be renamed to something without the meaningless numbers? For example, `richtext-demo-screenshot.png`?

JordanMartinez:
Thank you for your work @scientificware. However, this is the second PR you've submitted where, rather than using git to change the commit history, you have opened a new PR that has the final changes/updates, which disconnects your original work and the discussion we have on it from the actual work that gets merged.

Could you explain why that is happening? It's not a pattern I would like to keep following.

scientificware:
I'm really sorry for that, but I don't know yet how to use git to change the commit history.
I tried with "git rebase -i", but I lost myself in my local repository and my changes.
So, I deleted my fork and my local repository and I cloned them again.

I'm working with netbeans and push my commits with the built-in git support.

As I was working on my master branch all changes were pushed in the PR.
I changed my strategy and now I work with new branch for each change.

JordanMartinez:
> I changed my strategy and now I work with new branch for each change.

Although we don't have anything documented regarding this (maybe it's assumed?), that's the better approach to take.

No need to apologize! This is simply providing you a learning experience, and I want to make sure you are exploiting this opportunity as much as possible. It seems like you could make a blank git repo (no code, just text documents) and simulate the kind of work you did here (a lot of developmental commits before reaching the final goal) before you try various ways to squash the commits down into fewer ones (simplifying the commit history).

scientificware:
Don't worry, I will continue "to play" in my forks until I'm ready to Pull a decent Request with a simplified commit history. :-)
Thanks and Happy New Year !

PR #656, opened 2017-11-24
RFE implementation for #653 / #655

Jugen:
RFE implementation for #655
The methods in `ViewActions` related to the following have been @**Deprecated**:

> onOutsideSelectionMousePress
> onInsideSelectionMousePress
> onNewSelectionDragEnd
> onSelectionDrop

They have been replaced with FXML friendly equivalent (non default) interface methods:

> onOutsideSelectionMousePress**ed**
> onInsideSelectionMousePress**ed**
> onNewSelectionDrag**Finished**
> onSelectionDropp**ed**

Which have been implemented in `GenericStyledArea`. The references to the deprecated methods in `GenericStyledAreaBehavior` have been changed to the new ones.

During this process it was found that the `onNewSelectionDragEnd/Finished` functionality hadn't been implemented in `GenericStyledAreaBehavior`  and this was rectified as well.

PR #678, opened 2018-02-06
Use correct replace method when SEG is String

### LINKED ISSUE #676: Constructor `InlineCssTextArea(String text)` initializes area with 1 paragraph with newline characters rather than multiple paragraphs
## Expected Behavior

The `BoundsPopup`s should disappear when they go **outside of the window**.

The `BoundsPopup`s should *also* disappear when the **window isn't focused**.

## Actual Behavior

The BoundsPopups *don't disappear* when they move **outside the window**.

The BoundsPopups *don't disappear* when the **window isn't focused**.

## Reproducible Demo

See [`PopupDemo.java`] in `richtextfx-demos`.

[![Demo]]Demo

## Environment info:

- RichTextFX Version: v0.8.2
- Operating System: Windows 7 Professional
- Java version: 1.8.0_161

[`PopupDemo.java`]: https://github.com/FXMisc/RichTextFX/blob/master/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/PopupDemo.java
[Demo]: https://user-images.githubusercontent.com/25108287/35783527-71450866-09ce-11e8-8c18-87b95597436b.gif

JordanMartinez:
Resolves #676.

In addition to fixing that issue, I also bumped the type of the area's segmentOps field from `SegmentOps` to `TextOps`. The area's constructor already requires the latter type, so this just makes it easier to call `StyledDocument`-related methods.

It is unlikely for this bug to be reproduced in the other areas because their SEG generic isn't purely text. Still, I wonder whether the method names should be changed to prevent problems like this from arising in the future. I think one test that could be written for this would be to check the number of visible paragraphs after constructing each kind of area.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
@@ -547,7 +547,7 @@ public class GenericStyledArea<PS, SEG, S> extends Region
      * ********************************************************************** */
 
     private final TextOps<SEG, S> segmentOps;
-    @Override public final SegmentOps<SEG, S> getSegOps() { return segmentOps; }
+    @Override public final TextOps<SEG, S> getSegOps() { return segmentOps; }
 
     private final EventStream<Boolean> autoCaretBlinksSteam;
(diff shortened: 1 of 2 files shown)


JordanMartinez:
I think I will wait to merge this until I hear back from Jugen in #656.

Issue #449, opened 2017-03-11
IndexOutOfBoundsException in RichText demo


clartaq:
When changing the font of the entire contents of the demo editor pane, an IndexOutOfBounds exception is thrown in the "JavaFX Application Thread".

Steps to reproduce:

1. Start the RichText demo (from the pre-built jar or building in Gradle).
2. Paste the README.md file into the demo.
3. Select All (Ctrl-A)
4. Select a different [font size or font family from their respective `ComboBox`] (edited by Jordan)

Environment:

- Windows 10 Pro, 64-bit
- RichTextFX 0.6.10
```
java -version
java version "1.8.0_121"
Java(TM) SE Runtime Environment (build 1.8.0_121-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.121-b13, mixed mode)

```

JordanMartinez:
@clartaq Does this also occur using the `0.7-M3` release?

clartaq:
The same error occurs when using 0.7-M3, whether I paste the contents of the 0.6.10 README.md or the 0.7-M3 version.

Also, using the new command in the demo to read a file produces an OutOfMemoryError when trying to load either of these files.

JordanMartinez:
I can't reproduce the bug on my end in either of those versions.
````
Linux Mint 18.1
java version "1.8.0_121"
Java(TM) SE Runtime Environment (build 1.8.0_121-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.121-b13, mixed mode)
````

> Also, using the new command in the demo to read a file produces an OutOfMemoryError when trying to load either of these files.

Could you create a new issue specifically for that and with more details as to the issue? I'm not sure which command to which you're referring.

**Comment by JordanMartinez** [comment]
Can someone else test this on a Windows environment?

**Comment by kuc** [comment]
I can confirm that bug, Windows 10 Enterprise.

Stacktrace:

```
Exception in thread "JavaFX Application Thread" java.lang.IndexOutOfBoundsException
        at com.sun.javafx.scene.control.ReadOnlyUnbackedObservableList.subList(ReadOnlyUnbackedObservableList.java:136)
        at javafx.collections.ListChangeListener$Change.getAddedSubList(ListChangeListener.java:242)
        at com.sun.javafx.scene.control.behavior.ListViewBehavior.lambda$new$177(ListViewBehavior.java:269)
[...]

PR #685, opened 2018-02-14
Updated Paragraph to apply styles to empty paragraphs.

Jugen:
This is a fix for an issue discovered in #449 where styles were not being applied to empty paragraphs. In that issue it was causing an IndexOutOfBoundsException within JavaFX.


JordanMartinez:
Thanks for the PR!

Issue #689, opened 2018-02-19
ArrayIndexOutOfBoundsException in ParagraphText::getRangeShapeSafely


samhutchins:
## Expected Behaviour

No crash

## Actual Behaviour

When changing the text that's in a CodeArea, with some text selected, it's possible to put the component in a state that will throw an `ArrayIndexOutOfBoundsException`

## Environment info:

- RichTextFX Version: 0.8.2
- Operating System: Windows 10, macOS High Sierra
- Java version: 1.8.0_161

## Current Workarounds

Can work around by calling `clear` on the CodeArea prior to calling `replaceText`

## Reproducible Demo

1. Run the following program
1. Click "Set text a"
1. Scroll to the bottom
1. Select the last 2 lines
1. Click "Set text b"
1. Click "Set text a"
1. You'll get the following backtrace:
```
Exception in thread "JavaFX Application Thread" java.lang.ArrayIndexOutOfBoundsException: -4
	at org.fxmisc.richtext.ParagraphText.getRangeShapeSafely(ParagraphText.java:320)
	at org.fxmisc.richtext.ParagraphText.updateSelectionShape(ParagraphText.java:281)
	at org.fxmisc.richtext.ParagraphText.layoutChildren(ParagraphText.java:388)
	at javafx.scene.Parent.layout(Parent.java:1087)
	at javafx.scene.Parent.layout(Parent.java:1093)
	at javafx.scene.Parent.layout(Parent.java:1093)
	at org.fxmisc.flowless.VirtualFlow.layoutChildren(VirtualFlow.java:257)
	at javafx.scene.Parent.layout(Parent.java:1087)
	at javafx.scene.Parent.layout(Parent.java:1093)
	at javafx.scene.Parent.layout(Parent.java:1093)
	at javafx.scene.Parent.layout(Parent.java:1093)
	at javafx.scene.Scene.doLayoutPass(Scene.java:552)
	at javafx.scene.Scene$ScenePulseListener.pulse(Scene.java:2397)
	at com.sun.javafx.tk.Toolkit.lambda$runPulse$29(Toolkit.java:398)
	at java.security.AccessController.doPrivileged(Native Method)
	at com.sun.javafx.tk.Toolkit.runPulse(Toolkit.java:397)
	at com.sun.javafx.tk.Toolkit.firePulse(Toolkit.java:424)
	at com.sun.javafx.tk.quantum.QuantumToolkit.pulse(QuantumToolkit.java:518)
[...]

PR #690, opened 2018-02-21
Fix replaceText IOOBE

Jugen:
This is a fix for #689  ArrayIndexOutOfBoundsException in `ParagraphText.getRangeShapeSafely` after calling `replaceText` under certain conditions.

### DIFF
--- a/richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/ReplaceTextTests.java
+++ b/richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/ReplaceTextTests.java
@@ -1,0 +1,629 @@
+package org.fxmisc.richtext.api;
+
+import org.fxmisc.richtext.InlineCssTextAreaAppTest;
+import org.junit.Test;
+
+public class ReplaceTextTests extends InlineCssTextAreaAppTest {
+
+    @Test
+    public void deselect_before_replaceText_does_not_cause_index_out_of_bounds_exception()
+    {
+        interact( () ->
+        {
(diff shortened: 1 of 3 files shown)


JordanMartinez:
Thanks for the PR!

JordanMartinez:
I looked through the Travis CI logs and came across [these lines](https://travis-ci.org/FXMisc/RichTextFX/jobs/344826898#L2384). Is it possible to get rid of them somehow? Or is that necessary in order to run the test due to the diaereses/umlauts and just an unfortunate issue on the Mac JDK?

Also, could the test work without such a long and verbose text that otherwise obscures the issue with a lot of noise? For example, you shared that you were able to reproduce the issue (though perhaps for different reasons) using the texts:
````
Text A: "1234567890\n\nabcdefghij\n\n1234567890\n\nabcdefghij"
Text B: "1234567890\n\nabcdefghijklmno\n\n1234567890\n\nabcde"
````

Jugen:
Yeh, I noticed that in the Travis logs as well and unfortunately I don't know if it's possible to get rid of them. As you surmised, I believe it is necessary for the reasons you mention.

I created two tests, one using the long & verbose text and the other my shorter version - since there is a subtle difference to the way the IOOBE gets triggered in each. I had another look, doing a text comparison between the two long texts and have managed to reduce it down to a much shorter version with the same properties and results as before. :-)

JordanMartinez:
Ok. It is what it is. I'm going to merge #687 first and then I'd like to merge this one since it'll take me a lot more work to include this one in mine than to include mine in yours.

Issue #691, opened 2018-02-21
Question: How to add style/padding to individual ParagraphBox? 


kszilagyi:
Hi,

So I figured out how to set padding for all lines:
```
.styled-text-area .paragraph-box{
    -fx-padding: 10 0 0 0;
}
```
Works very well. Now the problem is that I want to set it for particular lines (determined in runtime). But I couldn't figure out how to set a class for a particular paragraph box. Is that possible? 

Thanks

JordanMartinez:
No, it's not possible to set it for specific lines, except for the first and last paragraph in the area and the paragraph that has the caret.

kszilagyi:
Thanks for your answer. Is this by design? If I wanted this would I be able to add it with relatively little effort?

JordanMartinez:
It is a limitation in the current design. Such a feature would need to modify the internal code; you wouldn't be able to add it by building on top of this project (at least not in a non-hacky way). 

It could probably be added by defining a style class naming scheme like "paragraph-box-`n`" where `n` corresponds to the paragraph index. Then, whenever ParagraphBox's index is updated, the old style class is removed (e.g. "paragraph-box-3") from the paragraph and the new style class is added (e.g. "paragraph-box-4")

kszilagyi:
I see, ok, thanks for explaining

JordanMartinez:
Coming from #692:

> Basically what I want to do is this:
> Load a file. Based on the content set some paragraph spacing big and some small. Would that work?
> I would also need to recalculate all the spacing whenever the content changes.

Let me put this into my own words to insure I'm understanding you correctly:
- You want to load a file and display its contents in an area
- Depending on the content on each paragraph, you'll want the paragraph to have some or no amount of space separating it from the previous and/or next paragraph. You want to set this spacing amount dynamically and not statically through a CSS style class.
- As the contents are edited (whether programmatically or by the user), you want this spacing computation to be recalculated

If I'm understanding you correctly, then it seems that the feature you want is to configure each paragraph each time it is either rendered or modified. That implies a `Consumer<ParagraphBox>` field being set in an area and being called each time a new paragraph is rendered to the screen and each time it's index is updated.

kszilagyi:
[...]

PR #692, opened 2018-02-22
Added GenericStyledArea.setParagraphBoxStyle

kszilagyi:
Partially solves: #691

Now you can write this: 
``` 
codeArea.setParagraphBoxStyle(1, "-fx-padding: 20 0 0 0;")
```
And then you get something like this:
image

Problems:
- I believe this is not how it supposed to be solved, if you let me know how to make it better I am happy to fix it.
- If the line is edited the styling goes away. If you let me know how to fix it I will try.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
@@ -1114,6 +1114,10 @@ public class GenericStyledArea<PS, SEG, S> extends Region
         content.setParagraphStyle(paragraph, paragraphStyle);
     }
 
+    public void setParagraphBoxStyle(int paragraph, String paragraphStyle) {
+        virtualFlow.getCell(paragraph).getNode().setStyle(paragraphStyle);
+    }
+
     @Override
     public final S getTextStyleForInsertionAt(int pos) {
         if(useInitialStyleForInsertion.get()) {


JordanMartinez:
Thanks for the PR @kszilagyi. Unfortunately, this approach is severely flawed. If you follow the approach I suggested in the issue, it should properly implement this without any of the problems you mention in yours.

Here's a clearer description of what I mean. If you remove/add the style class to the box [in this method](https://github.com/FXMisc/RichTextFX/blob/3fae2c8e086034162aba687fa1a1f2b464cb6942/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java#L1305-L1379) via a ReactFX EventStream that is subscribed to the box's index values and is unsubscribed when the box is disposed of, it should do what you want without any problem.

kszilagyi:
Thanks for your response @JordanMartinez. So I am new to JavaFX. Would I be able to dynamically add styles with your approach? I couldn't figure out if I can have an in memory CSS file or not. But basically what I want to do is this:
Load a file. Based on the content set some paragraph spacing big and some small. Would that work?
I would also need to recalculate all the spacing whenever the content changes. 

Thanks

JordanMartinez:
Mm... ok. My approach works for adding styles statically. Adding it dynamically makes things a bit harder. Let's keep discussing this in the issue since you might be requesting an entirely different feature.

JordanMartinez:
Closing due to need to use a different approach than what is used here.

PR #710, opened 2018-03-28
Use correct class in `instanceOf` check; provide faster equals method

### LINKED ISSUE #709: Border text styling is segmented when overlaying styles
## Expected Behavior

Overlaying some style spans should not break down the border and box individual spans. The border should be around the original span, not individual spans resulting from the overlay.  

## Actual Behavior

capture du 2018-03-27 21-48-01

There are two StyleSpans here that are overlayed :
* One with the border outline and background variation. Each styled span in this layer is applied from opening brace to closing brace (in one continuous span for each block)
* One with the syntax highlighting.

Overlaying both layers breaks down the big spans with borders into smaller spans around the other styled spans. But the borders are shown around each particular span instead of around the original span, which is undesired. It's not entirely unexpected given the contract of overlay though, which is why I'm not sure I should submit a bug report. What do you think?

I think, if it's entirely expected, then one feature that could allow to work around that would be allowing to select the styles of the border for each side, somewhat like `-fx-border-style: none solid solid solid;`. With that, I could probably adjust the borders of the overlaid spans during the overlaying. 

Thanks for any help/ feedback. 

## Minimal reproducible demo

https://gist.github.com/oowekyala/8319f98f7ed77042e2a4fa0e79116aad#file-app-java

## Environment info:

- RichTextFX Version: 0.8.1
- Operating System: Ubuntu 16.04
- Java version: 8

JordanMartinez:
Fixes #709.

### DIFF
--- a/richtextfx/src/integrationTest/java/org/fxmisc/richtext/SceneGraphTests.java
+++ b/richtextfx/src/integrationTest/java/org/fxmisc/richtext/SceneGraphTests.java
@@ -4,6 +4,8 @@ import static org.junit.Assert.assertNotNull;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.function.Function;
+import java.util.function.Predicate;
 
 import org.fxmisc.flowless.Cell;
 import org.fxmisc.flowless.VirtualFlow;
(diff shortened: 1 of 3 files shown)


JordanMartinez:
Sigh... that Mac build taking so long to even start...

PR #713, opened 2018-03-29
Add PsychSQL to list of projects that use RichTextFX

tmptmpuser:
Update README.md to include PsychSQL (a one-file SQL client on top of JDBC) in the "Who uses RichTextFX?" section.


JordanMartinez:
Thanks for the PR!

PR #695, opened 2018-02-28
Allow multiple portions of an area's document to be updated in one call

JordanMartinez:
Resolves #574.

One feature still lacking in this PR: the UndoManager cannot undo/redo a list of changes because I still need to think through how two lists of changes should be merged and what their identity values would be.

PR #721, opened 2018-04-05
Prevent `multiPlainChanges()` from emitting an empty list

### LINKED ISSUE #720: MultiPlainChanges emits an empty list
First discovered (somewhat) [here](https://github.com/FXMisc/RichTextFX/issues/715#issuecomment-378030361)

Using the following code:
````java
codeArea
   .multiPlainChanges()
   .subscribe(ignore -> codeArea.setStyleSpans(0, computeHightlighting(codeArea.getText()));
````
the code runs recursively until throwing a `StackOverflowError` because `multiPlainChanges()` emits an empty list when only the area's styles are updated.

JordanMartinez:
Fixes #720 

Also cleans up the demos to use the new `multiPlainChanges()` API introduced in #695


JordanMartinez:
I inverted the identity check. Should build find now.

Issue #733, opened 2018-04-19
Links to the demo source code and demo page are broken


creativeArtie:
There are two demo links in the main README file that don't go to anywhere.

Also the source code links in the demo page goes to a 404 page.

JordanMartinez:
Closed by #734

PR #734, opened 2018-04-19
Fixed the links pointing to the demo source pages

creativeArtie:
Here's a quick fix to #733. I also removed the word "below" and add the missing "Demo" related to the links


JordanMartinez:
Thanks for the fix!

PR #744, opened 2018-05-10
Updates README.md to add Greenfoot as a project that uses RichTextFX

amjdhsn:
Adds Greenfoot as a project that uses RichTextFX


JordanMartinez:
Thanks for the PR!

To be honest, I find it somewhat ironic that you're using a library that utilizes a lot of functional programming to teach the object-oriented paradigm via Java.

PR #769, opened 2018-07-16
Update UndoFX (2.1.0); stop maintaining project

JordanMartinez:


JordanMartinez:
Builds will fail at this point because the UndoFX release hasn't hit Maven yet.

PR #779, opened 2018-10-10
Issue #627 Extreme memory usage with large blocks of text

### LINKED ISSUE #627: Extreme memory usage with large blocks of text
Is it normal/expected to see this kind of memory usage? Memory is shown at the top of the screencaps. 

The first one is a modified version of one of the demos which adds 4,000 lines. Memory usage on that one spikes to about twice what's shown here, and then drops down. It also functions fine. 

rtfx1

This second screenshot is with 400,000 extra rows of text added. Demo becomes unusable with this much text. When I add a single letter to a row, it freezes for a few seconds before showing the character. Scrolling still works okay, but the rendering speed isn't great. Memory usage... 2GB? :exclamation: :exclamation: :exclamation: It also fluctuates wildly (up to 1GB or so) when scrolling, which is only going to make things worse.

rtfx2

I made every line different on purpose since it looks like some optimizations are going on under the hood, but even without that I still see this behavior. 

I'm running on Linux. Oracle JRE 8u144

JonathanMarchand:
Resolves #627 

Fix memory leak of ParagraphText through listeners:
- Make listeners static class
- Use WeakReference to refer to the ParagraphText from the listeners
- Remove the listeners if the ParagraphText is garbage collected.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/ParagraphText.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/ParagraphText.java
@@ -1,5 +1,6 @@
 package org.fxmisc.richtext;
 
+import java.lang.ref.WeakReference;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


Jugen:
Thanks for this @JonathanMarchand

Issue #777, opened 2018-10-01
Random crashes in GenericStyleArea.visibleParToAllParIndex


kszilagyi:
## Expected Behavior
Shouldn't throw exception but return the index of the paragraph.
## Actual Behavior
Sometimes when I call this method it reaches: `throw new AssertionError("Unreachable code");`
As much as I can tell the problem is that the element in the visibleParagraphs is the same as the one in the paragraphs but they are not the same reference therefore the search can't find it. I have no idea how is this possible. 
## Reproducible Demo
Unfortunately I only have this rarely...
## Environment info:

- RichTextFX Version: 0.9.1
- Operating System: Ubuntu 18.04 but happened in 16.04 too.
- Java version: "1.8.0_181"

lhttjdr:
I write some blank paragraphs (only an ENTER each row).
- `getVisibleParagraphs()` will return an array with correct size (number of paragraphs).
- `visibleParToAllParIndex()`, `firstVisibleParToAllParIndex()` and `lastVisibleParToAllParIndex()` will throw out "Unreachable code" errors

I checked source code of `visibleParToAllParIndex()` in `GenericStyledArea.java`
```java
        Paragraph<PS, SEG, S> visibleP = getVisibleParagraphs().get(visibleParIndex);
        for (int index = 0; index < getParagraphs().size(); index++) {
            if (getParagraph(index) == visibleP) {
                return index;
            }
        }
```
It is a simple comparison between results of `getVisibleParagraphs()` and `getParagraphs()`.

Unfortunately, those two methods will give different results corresponding to the same blank paragraphs.

I use such a test case with 4 rows:
```
{Space}{Enter}
{Enter}
{Enter}
{Space}
```
, only the first and fourth row can be matched.

Jugen:
@lhttjdr Thank you for the update to this issue.

Could you please provide code for a simple test case that throws out "Unreachable code" error.
I would really appreciate it, thanks.

lhttjdr:
@Jugen Thank you for your great job at first.
I write some test cases for this problem and make a pull request.

Jugen:
[...]

PR #790, opened 2018-12-01
API for visible paragraphs works when contents have blank lines

lhttjdr:
These test cases are for issue #777.

It shows that `firstVisibleParToAllParIndex`, `lastVisibleParToAllParIndex` and `visibleParToAllParIndex` will throw out "Unreachable code" errors when the contents have blank lines/paragraphs.

However, it not always behaves like that.
- When the paragraphs list start and end with non-blank lines,  `visibleParToAllParIndex` will works correctly even there are some blank lines.
- When the edit area has scroll bar, its behavior has much more complicated pattern. For example, when visible region only contains blank lines, it will throw out errors. After scroll to some non-blank lines, it will work again.

What I know now is that `getVisibleParagraphs` returns wrong list corresponding to blank lines, so that all those 3 functions above won't work because it's impossible to find a same item in list returned from  `getParagraphs`.


Jugen:
Thanks, this was covered by the PR fix #795

PR #791, opened 2018-12-15
ParagraphText: immediately remove listeners from selections and carets on disposal

JFormDesigner:
This PR immediately removes listeners (in class `ParagraphText`) from `SelectionPath` and `Caret` on disposal instead of waiting for GC (this improves PR #779, issue #627). Without this change, disposed ParagraphText objects may still listen to selection and caret and do some useless stuff.

Since the weak listeners added in PR #779 are no longer necessary, I've removed the listener classes again and moved selection and caret listeners code back to constructor (saves 100 lines of code).


JFormDesigner:
Forgot to mention that I've debugged the change and also used VisualVM to check whether the change still fixes the old memory leak (that was fixed in PR #779).

@JonathanMarchand would you be so kind and also test this PR? Thx

Issue #638, opened 2017-11-01
Background color is applied in the wrong place when rendered segment type is not a TextExt


FHannes:
In the RichText demo applying styles works correctly, with the exception of the background color. When the editor just has text, there is no issue. When the text has an image insert before the selection, the application of the background color is negatively offset by the number of images before the selection.

JordanMartinez:
Meh... I guess my title doesn't quite state the issue correctly either...

Still, the problem occurs in `ParagraphText`. I think there's a line somewhere in there that only considers `TextExt` objects when constructing the `Path` background shape.

JordanMartinez:
Aye. The issue arises [here](https://github.com/FXMisc/RichTextFX/blob/067714bf6d5b8d989fa98c5bc48ad8de5ada2251/richtextfx/src/main/java/org/fxmisc/richtext/ParagraphText.java#L351-L381). We'd need a way to determine whether a given rendered node shares the same background color as another consecutive one while accounting for different subclasses of `Node` that are used to render the segment on the screen. While this issue only mentions the background color, it could equally apply to the rtfx-specific border and underline styles, too.

Due to accounting for non-text objects, we'd also need to calculate the length of a segment to determine how long the range of that segment's background color or border is. Calling `SegmentOps#length` on the SEG object would easily calculate its length, but it would mean adding the `ops` object as one of the constructor parameters of `ParagraphBox` and `ParagraphText`.

One way would be to add a lambda (Function) that can be passed an argument and return an `Optional<Paint>` to determine whether two nodes share the same background color. If the the passed-in object is be a `StyledSegment<SEG, S>`, this would help us determine the background color for some use cases (e.g. when inline css is used) but not for others (when a style class is used). However, passing in the rendered node with type `Node` will require `instanceOf` checks and casting before the background color could be determined. (To get around this, we'd need to use the Visitor pattern and make all rendered nodes implement it. Definitely not something I like).

PR #793, opened 2018-12-20
fix position of background, border and underline shapes in case that the line contains custom objects 

JFormDesigner:
If a line contains a custom object (e.g. image in RichText demo) and you apply background color to text after that image, then the background is painted one character to the left where it should be painted.

The problem is simply that the image is not considered when calculating the start position of the background color.

This PR fixes issue #638


JFormDesigner:
Steps to reproduce the issue:
1. start `RichTextDemo`
2. enter some text; e.g. "abc"
3. click "insert image" in toolbar and insert some image (into same line as previous text)
4. enter more text in the same line after the image; e.g. "123"
5. select the "3"
6. click "Text background" in toolbar and choose some color

Now the background color is shown at the "2", but it should at the "3"

Jugen:
@JFormDesigner thanks for the PR

Issue #724, opened 2018-04-09
Bug: Overlaying StyleSpans causes the edited line to move up


liz3:
So this is a bit more specific and not easy to explain, ill try my best though.

I had the question #715 which was answered successfully.
Now i had to discover a (probably) bug related to this.

Let me try to Explain:
When overlaying a paragraph with other StyleSpans, there are cases, where after it applied the changes to the style,  you than press a key which inserts a char, moves the Caret or anything else, moves the line/paragraph you just edited to the top of the area viewport.
This only happens once, so you have to edit the spans again to re create this Effect.
Another requirement, at least in my case is, that the spans of the paragraph are actually changed, when the old styleSpans from the paragraph get applied again to the same paragraph without changes, the bug will not appear.
So the creator of this bug is: `area.setStyleSpans` with edited style spans.

Could this be a bug, or am i doing a mistake?

Kind Regards,
Liz3

JordanMartinez:
Please provide a reproducible demo. It's often easier to understand what you're talking about when I can reproduce it myself. Sometimes, when pursuing that end, you'll realize what's causing the mistake yourself.

liz3:
Hey, i was not able to reproduce the error.
And i found ways to fix this, so the bug is depending on my Project.
Its just super Hard to debug this, because the Project/its gui layer has become quite big and there could be so many sources for this.

I opened this issue though, because i was able to clearly identify that the bug at the end comes from area.setStyleSpans.

This Issue is not closed but "put on hold".
Once i found the bug i want to provide the answer here.

Regards
Liz3

liz3:
Update, i have a reproducible demo.

Just as information, turned out its related to a lot of things, but most important it seams to be related to the `VirtualizedScrollPane`.

I added the Complete file.

To reproduce:
Just start the file(the css file is also included). Go somewhere in the Editor and edit one char to trigger the subscription.
Now we want to focus on line `26`.
Now for the Bug, the size of the window plays a important role, but the basic is:
when line 26 is in the upper half of the area viewport and the **width** of the window is **below** a certain width, as soon as a key gets pressed, which moves the caret, line 26 will jump to the top of the editor.
If either line 26 is in the lower half of the area viewport or the width of the window is higher than that magic amount, the bug will take effect.
I also recorded a little video, if i did not explained enough here.

Any ideas from what this is coming?

[DemoCode.txt](https://github.com/FXMisc/RichTextFX/files/1896775/DemoCode.txt)
[...]

PR #794, opened 2018-12-21
avoid scrolling current line to top of view when start editing

JFormDesigner:
This PR fixes a painful behavior that I noticed in [Markdown Writer FX](https://github.com/JFormDesigner/markdown-writer-fx):
When I start typing into the current line, RichTextFX immediately
scrolls the current line to the top of the view.

Turns out that this happens only if embedded images feature is enabled
and if there are of course some embedded images before the current line.

The cause for the behaviour is in `GenericStyledArea.followCaret()`.
For some unknown reason, the cell of the current line that is returned by
`virtualFlow.getCell(parIdx)` is not visible (`cell.getNode().isVisible() == false`)
although it is visible on screen.

Later on in `Navigator.visit(MinDistanceTo targetPosition)`, the VirtualFlow still
thinks that is is not visible and invokes `Navigator.placeStartAtMayCrop()` to
scroll the current line to the top of the view.

The fix is easy: `virtualFlow.layout()`

Then we have a visible cell in `followCaret()` and also later in
`Navigator.visit(MinDistanceTo targetPosition)`, which then invokes
`placeToViewport()` and scrolls only when really necessary.

This PR probably fixes also issue #724, which describes the same strange behavior:
> ... press a key which inserts a char, moves the Caret or anything else, moves the line/paragraph you just edited to the top of the area viewport.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
@@ -1479,6 +1479,12 @@ public class GenericStyledArea<PS, SEG, S> extends Region
     /** Assumes this method is called within a {@link #suspendVisibleParsWhile(Runnable)} block */
     private void followCaret() {
         int parIdx = getCurrentParagraph();
+
+        // Make sure that cells are up-to-date to avoid scrolling current line to top when
+        // start editing, which may occur under certain circumstances. E.g. if having embedded
+        // images, which results in some lines that are much higher then the average line height.
+        virtualFlow.layout();
+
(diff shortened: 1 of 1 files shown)


Jugen:
@JFormDesigner I'm not sure but I think that this is related to #777 (?), so I'm not going to merge this just yet. I'm having real heartache trying to resolve it properly though, as the solution I have causes chaos with undo at the moment.

JFormDesigner:
@Jugen why do you think that it is related to issue #777?
#777 is about `GenericStyleArea.visibleParToAllParIndex()` which is actually never invoked from RichTextFX (except in tests).

PR #810, opened 2019-03-25
Fix height paragraph box height calculation

### LINKED ISSUE #809: Lines sometimes overlap when one wraps by a small amount
## Expected Behavior

When adding lines to a vertical styled text area, none should overlap.

## Actual Behavior

If two lines,  A and B, are added sequentially, and A is very slightly too long to fit in one line, the wrapped part of A overlaps the start of B.

e.g.:

image

The first line *almost* fits in, but is a few pixels short. It wraps, but the height calculations don't seem to take notice of that and the second line renders on top of it.

csmith:
The insets were removed from the "overhead" when they should have been
added. This was asking the text for its preferred height if it had
(width + insets) space, rather than (width - insets).

Because of this, some paragraphs give a preferred height of one line
less than they should (because they wouldn't need to wrap given the
extra space that's actually occupied by the insets).

Fixes #809


Jugen:
@csmith thanks alot for the PR and fix.

PR #834, opened 2019-07-29
Fixed caret position off by one on wrapped multiline

### LINKED ISSUE #833: Caret Position Error on Wrapped Multiline
Often on wrapped multiline texts (such as the one shown below), the caret position is unable to be moved with the mouse along to encapsulate the entirety of the line. The caret position appears, at its maximum to be 1 character off from the actual requested end of line position

Capture

This was produced using the RichTextDemo

Jugen:
Fixes #833 where the caret isn't positioned correctly at the end of the last line in a multiline paragraph.


javaware777:
Java version: 10.0.2
RichTextFX: 0.10.1

Issue #841, opened 2019-08-15
In a multi-line paragraph how to highlight the line the caret is on?


garybentley:
I'm wanting to have the line the caret is currently on be highlighted.  I've asked this before #611 but a lot has changed since I first asked and now I'm thinking of doing:

1. Add a listener to the caret position.
2. In that listener get the line the caret is current on.
3. Adding a selection with a custom style for the start/end positions of the line.

Would this approach work?

The main issue I can see is that there doesn't seem to be any way to get the text offsets of the start and end of the line.  Is there any way to get those offsets?

garybentley:
I've now implemented this but using a different method.  I'm listening to the caret bounds and then drawing and positioning a Pane behind the area for the bounds position.

Thus:

```
Pane background = new Pane();
Pane highlight = new Pane();
background.getChildren().add(highlight);
GenericStyledArea<?, ?, ?> area = new GenericStyledArea(...);
StackPane sp = new StackPane();
sp.getChildren().addAll(background, area);
area.caretBoundsProperty().addListener((pr, oldv, newv) ->
{
   Bounds b = background.screenToLocal(newv.get());
   highlight.relocate(0, b.getMinY());
   highlight.setPrefHeight(b.getHeight());
   highlight.setPrefWidth(area.getWidth());
});
```

This produces something like the following:

![text-line-highlight](https://user-images.githubusercontent.com/921991/63134322-4ddd4c80-c00c-11e9-93ea-dcb894d84c7a.jpg)

Note: sometimes the caret bounds aren't present.

While this works, it would be nice to have an "in area" method of doing this.  This type of feature is common in writing applications and code editors.

Jugen:
That's a good hack :-)

I've had a look at this and I'm planning on adding the following API to `GenericStyledArea`:
```java
setLineHighlighterColor( Color highlight )
showLineHighlighter( boolean show )
getCurrentLineStartInParargraph()
getCurrentLineEndInParargraph()
```
Any thoughts or comments welcome.

garybentley:
That would be some great additions.  It might be better to have a Paint rather than a Color though, for those who want to do something a little fancier.

The only problem with my workaround is that I'm now seeing this problem #842.

Jugen:
RichTextFX 0.10.2 has been released with this feature.

PR #845, opened 2019-08-22
Added line highlighter enhancement

Jugen:
This PR is for RFE  #841 adding the following API to GenericStyledArea:

```java
setLineHighlighterFill( Paint highlight )
setLineHighlighterOn( boolean show )
isLineHighlighterOn()
```

The highlighter can be styled via CSS using selectors from Path, Shape, and Node. For example:

```css
.styled-text-area .line-highlighter { -fx-fill: lime; }
/* or */
.line-highlighter { -fx-highlight-fill: lime; }
```

Any added custom Selections can now also be styled via CSS using the selection's name.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
@@ -25,6 +25,7 @@ import javafx.beans.property.SimpleDoubleProperty;
 import javafx.beans.property.SimpleObjectProperty;
 import javafx.beans.value.ObservableValue;
 import javafx.collections.FXCollections;
+import javafx.collections.ListChangeListener.Change;
 import javafx.collections.ObservableSet;
 import javafx.css.CssMetaData;
 import javafx.css.PseudoClass;
@@ -47,6 +48,8 @@ import javafx.scene.layout.CornerRadii;
 import javafx.scene.layout.Region;
 import javafx.scene.paint.Color;
 import javafx.scene.paint.Paint;
+import javafx.scene.shape.LineTo;
+import javafx.scene.shape.PathElement;
 import javafx.scene.text.TextFlow;
 
 import org.fxmisc.flowless.Cell;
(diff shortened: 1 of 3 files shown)


garybentley:
This works great, thanks for adding it.  There is just one issue though, if the line is empty, if the paragraph has zero length, then the highlight isn't shown.

garybentley:
A couple of other minor things:

1. The default text selection highlight color used to be a shade of blue, it now seems to be white.  How do I change the default?

2. When you do set a highlight line paint the same color is used for the text selection (i.e. double clicking on a word).

For my use case I'm not setting the highlight line color via CSS since in my application it is user configurable.

garybentley:
Just one other thing, if the line highlight paint is set it is overriding the default text selection paint.

In my stylesheet I have:

```
.selection
{
    -fx-fill: blue;
}
```

And am setting the line highlight paint via code, thus:

`area.setLineHighlighterFill(Color.red);`

Any selected text on that line has a red background rather than blue thus masking what text is selected.

Jugen:
Ok, I've made some changes that hopefully resolve the issues you reported.
Can you please test and let me know, thanks.

garybentley:
Yes that fixed both issues.  Thanks.

garybentley:
Found another issue with this I'm afraid.  The highlight can sometimes be "half" applied.

In the screenshot below you can see that the line highlight is only partially rendered.  This is on a wrapping line and occurs after I resize (by a splitpane dragger) the left hand component to be larger.  The text area contracts but the highlight will often not be fully rendered.  Moving the caret to another line renders the highlight properly.

line-highlight-half

garybentley:
There is another issue with this.  I've added it here to keep things in the same place, hope you don't mind.  I don't want to spam up the issues list.

If you have the line highlighter enabled and type at the start of the line then the highlight isn't drawn on the newly added text.

See below.

line-highlight-start

Jugen:
Thanks, I've fixed this.

PR #859, opened 2019-10-28
Added methods to recreate or get a paragraph graphic

### LINKED ISSUE #854: How to force a relayout?
I've asked a similar question to this before: #839 but I now have a different use case that I can't get around by setting a style.

In this case I am modifying what is drawn in a paragraph graphic when the user makes a change in the text.  Specifically I am drawing a line in the paragraph graphic to match where the user selected a position in the text.  See the image below:

edit-marker

The grey area is the paragraph graphic, the blue line is the "marker" to indicate how far into the text the user has edited.

My issue is that when the user changes the "edit position" I need to also update the marker but the marker is only changed on the next layout pass.  I tried calling "requestLayout" on the area and the paragraph graphic but neither caused another layout.

So any suggestions on how to force a layout?

(Note: I do have a workaround but it involves a nasty removing/adding of a component to the graphic which is FAR from ideal, I'd ideally like to have a proper way of handling this situation, especially since I have a number of use cases that have the same issue.)

Jugen:
Fixes #854 RFE


garybentley:
I'm just testing these now but please note that there is typo in both method names, they are currently:

```
recreateParagraghGraphic
getParagraghGraphic
```

But I believe they should be (change the last g in paragragh to p):

```
recreateParagraphGraphic
getParagraghGraphic
```

Issue #864, opened 2019-11-11
How to apply style so that subsequent text is using it?


garybentley:
I am trying to set the text style (using GenericStyledArea) such that text that follows will inherit/use that style.

To set some text as bold I'm doing:

 ```java
   private void toggle(Consumer<TextStyle> mixin)
    {
        IndexRange r = this.getSelection();
        this.setStyleSpans(r.getStart(), this.getStyleSpans(r).mapStyles(ss ->
        {
            TextStyle _s = new TextStyle(ss);
            mixin.accept(_s);
            return _s;
        }));
    }

    public void toggleBold()
    {
        IndexRange r = this.getSelection();
        TextStyle s = this.getStyleAtPosition(r.getStart());
        boolean t = !s.isBold();
        this.toggle (_s -> _s.setBold(t));
    }
```

This code will let me bold a piece of selected text but not set the style to be used for the text that follows the style change, except when I apply the bold at the start of an empty paragraph.

The question is, how do I set the style so that text that follows it will use the style?

The image below hopefully highlights what I mean.

![style-change](https://user-images.githubusercontent.com/921991/68553897-99b75a80-0470-11ea-9348-e3b9337bf7e1.jpg)

Jugen:
So I've spent some time looking into this and come up with the following:

There's a method `setUseInitialStyleForInsertion( boolean )` which when switched on currently uses whatever the initial text style was set to via the GenericStyledArea constructor.

There's also a `getInitialTextStyle()` method but **no** `setInitialTextStyle( style )` as for some reason the initial style is marked as **final** ?

So either the initial style object needs to be mutable or a `setInitialTextStyle( style )` method needs to be added and the final modifier removed. The latter being preferred I think.

Then to set the style so that any text that follows will use that style we do the following:
````java
   TextStyle theStyle = .....
   area.setInitialTextStyle( theStyle );
   area.setUseInitialStyleForInsertion( true );
````
And to switch it off a listener is added to the cursor with something like:
````java
   area.caretPositionProperty().addListener( (ob,ov,nv) ->
   {
       area.setUseInitialStyleForInsertion( false );
       area.setInitialTextStyle( DEFAULT_STYLE );
   }); 
````
Do you think this would be a logical and acceptable way of doing it ?
Any other thoughts ?

garybentley:
I think adding separate "insertion style" methods would probably be best.  If the "insertion style" isn't set then use the current behavior.  Then as the caret moves I can update the insertion style.

[...]

PR #867, opened 2019-11-14
Added methods to preset the style of inserted text

Jugen:
RFE #864 How to apply style so that subsequent text uses it.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
@@ -1279,10 +1279,19 @@ public class GenericStyledArea<PS, SEG, S> extends Region
         content.setParagraphStyle(paragraph, paragraphStyle);
     }
 
+    /**
+     * If you want to preset the style to be used for inserted text. Note that useInitialStyleForInsertion overrides this if true.
+     */
+    public final void setTextInsertionStyle( S txtStyle ) { insertionTextStyle = txtStyle; }
+    public final S getTextInsertionStyle() { return insertionTextStyle; }
+    private S insertionTextStyle;
+
     @Override
     public final S getTextStyleForInsertionAt(int pos) {
         if(useInitialStyleForInsertion.get()) {
             return initialTextStyle;
+        } else if ( insertionTextStyle != null ) {
+        	return insertionTextStyle;
         } else {
             return content.getStyleAtPosition(pos);
         }
--- a/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
@@ -1297,10 +1297,19 @@ public class GenericStyledArea<PS, SEG, S> extends Region
         }
     }
 
+    private PS insertionParagraphStyle;
+    /**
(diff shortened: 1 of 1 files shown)


garybentley:
This looks fine, however I think similar methods should be added for the paragraph style.  Consider the case where a user cut-n-pastes a chunk of text, consisting of multiple paragraphs into the area, being able to set the paragraph style(s) would be good.

garybentley:
Just tested this change with the following code:

```
IndexRange r = this.getSelection();
TextStyle s = this.getStyleAtPosition(r.getStart() + 1);

boolean t = !s.isBold();
s.setBold (t);
this.setTextInsertionStyle(s);
```
However when I then type a single character all none explicitly styled text in the paragraph has the style applied.  In this case all the none styled text is becoming bold and then un-bold when I turn it off again.

Am I doing something wrong?

Jugen:
I think it's because you are modifying the TextStyle object that is already associated with the text in the paragraph instead of setting a new TextStyle. If you are using a modified TextStyle from RichTextFX then try something like this:

```java
 TextStyle s = getStyleAtPosition( r.getStart() + 1 );
 TextStyle newStyleObject = s.updateWith( TextStyle.bold( ! s.isBold() ) );
 setTextInsertionStyle( newStyleObject );
```

Jugen:
Otherwise what you effectively want is something like:

```java
 TextStyle s = getStyleAtPosition( r.getStart() + 1 );
 TextStyle newStyleObject = new TextStyle( s );
 newStyleObject.setBold( ! s.isBold() );
 setTextInsertionStyle( newStyleObject );
```

garybentley:
Thanks, I should have noticed that was missing.  I'm using that technique for bolding a chunk of text.

I'm now using the new object for the insertion style but it's not using that style.  That is even if I switch on bold the next character I type isn't bolded.  I have checked that the style is correct.  Any ideas?  Or better still do you have some test code that you know works and then I base my code on that.

Jugen:
Then you probably have to switch off "useInitialStyle" with `area.setUseInitialStyleForInsertion( false );`

At the moment the code prefers to "useInitialStyle" before using the "insertionStyle":
```java
        if ( useInitialStyleForInsertion.get() ) {
            return initialTextStyle;
        } else if ( insertionTextStyle != null ) {
            return insertionTextStyle;
        } else {
            return content.getStyleAtPosition(pos);
        }
```
but I wonder if it shouldn't rather be the other way round like this:
```java
       if ( insertionTextStyle != null ) {
           return insertionTextStyle;
        } else if( useInitialStyleForInsertion.get() ) {
            return initialTextStyle;
        } else {
            return content.getStyleAtPosition(pos);
        }
```

garybentley:
Personally I think it should be:

insertionTextStyle else styleAtPosition else initialStyle

I'm not entirely sure what the point of the initial style is.  In my view methods like insertText, insert, replaceText and appendText should have a "style" argument that is used when inserting/replacing and if not present then the style should default to whatever the current style is at the insertion/replacement position, then, ultimately, have a default style that is used when nothing else is available.

Jugen:
>  In my view methods like insertText, insert, replaceText and appendText should have a "style" argument that is used when inserting/replacing ....

See newly added API from [#868](https://github.com/FXMisc/RichTextFX/issues/868#issuecomment-555014379) "Append text with different style" with PR #869 "Added convenience API to InlineCss and StyleClassed TextAreas".

PR #875, opened 2019-11-24
Added test and anchor properties fix

### LINKED ISSUE #874: caret and anchor stale after selectRange with listeners
If you attempt to use `selectRange` to update caret and anchor positions _that have attached listeners_, the caret and anchor potitions will be stale.

```kt
@Test fun `sscce`(){
    val underTest = StyleClassedTextArea()

    //comment out this line of code and the test passes
    val dontCare = Bindings.createStringBinding(Callable { "binded" }, underTest.caretPositionProperty(), underTest.anchorProperty())
    underTest.appendText("asdf")
    
    //act
    underTest.selectRange(1, 2)

    //assert
    assertThat(underTest.selectedText).isEqualTo("s") //OK
    assertThat(listOf(underTest.anchor, underTest.caretPosition)).isEqualTo(listOf(1, 2)) //fail
    //actual: 0, 2
}

```

Jugen:
Fixes #874 anchor properties not updating correctly if they have listeners attached.

The problem seems to come down to the `internalStartedByAnchor` Boolean property not firing on each invocation of `setValue`, but only when the actual value held changes.

The most non intrusive and safest fix, I feel, is to just toggle the Boolean property when needed in order to fire any relevant change listeners.

### DIFF
--- a/richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/selection/PositionTests.java
+++ b/richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/selection/PositionTests.java
@@ -159,6 +159,7 @@ public class PositionTests extends InlineCssTextAreaAppTest {
         });
     }
 
+    @Test
     public void deletion_which_includes_selection_and_which_occurs_at_end_of_area_moves_selection_to_new_area_end() {
         interact(() -> {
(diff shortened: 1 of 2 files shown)


Jugen:
This PR was reverted in favor of a better solution in PR #877

PR #890, opened 2020-01-03
Groundwork for virtual flow cell reuse in the future.

Jugen:
This came about while investigating a memory leak.

Currently it FAILS a number of tests with Java 8 but only `org.fxmisc.richtext.style.StylingTests` with Java 9. (Strangely though simulating the StylingTests manually does work.)

As a result cell REUSE is **disabled** and can be revisited maybe once JRE 8 support is dropped.


Jugen:
Messed up the PR, have resubmitted the code on a new branch.

PR #922, opened 2020-04-12
Allow ALT and ALT + CONTROL (or ALTGR on Windows) accelerators.

gaeqs:
This pull request improves the Predicate 'controlKeysFilter' (previously named 'noControlKeys') allowing ALT and ALT + CONTROL (or ALTGR on Windows) accelerators.

When on Windows, if an ALT + CONTROL key combination prints a special character the event pass and is consumed. Else, the event is filtered and the accelerator can be invoked.
This is how the default TextArea and several IDEs such as JetBrains IDEs or Eclipse behave.

A test called AcceleratorsTests can be found inside the integrationTest module. This class tests
all possible situation, both on Windows and Unix / iOS. This test was executed on Windows (Latest and LTSC) and Ubuntu 19.04. A small project was also made to test the changes.

Note that some Windows versions invokes two KeyEvents. The first one contains a null character and the second one the special character. This is a Windows bug that makes an accelerator to be called even if the combination prints a special code. I've tested on two different Windows 10 versions (Latest and LTSC) and on the second one this happens. Luckily, this is a small bug that is only appreciable when there's an accelerator binded to a key combination that prints a special character.


Jugen:
Thank you very much for the PR.

Issue #1131, opened 2022-07-18
Support wavy underlines


shoaniki:
## Desired Feature

When a user is editing text in a StyledTextArea, I wish to indicate typos and other kinds of error using the standard visual representation of a wavy underline.

## Current Workarounds

So far as I can determine, underlines currently can only be solid or dotted straight lines.  Dotted underlines are not unheard-of for marking typos, and that's what the RichTextFX spellchecker demo currently uses, but it is not the most familiar representation and may confuse users, particularly if colors other than red are to be used for different types of annotation.

## Implementation Costs

I have a basic implementation which I am submitting for discussion as a PR.  It involves creating a couple of extra CSS properties and then adding a branch to the code that creates underline shapes.

Jugen:
Thanks a lot for the PR :-)

PR #1132, opened 2022-07-18
Support wavy underlines (issue #1131)

shoaniki:
Basic implementation of wavy underlines as described in #1131.

Adds two properties -- "wave" to control the radius of arc used to draw wavy lines, and "offset" to adjust the distance between text and underline, since it can be necessary to increase this for legibility when a wavy underline is in use.  Probably better names are possible.

These are used in `TextFlowExt#getUnderlineShape` to modify the shape produced.  If no radius is specified then a line is drawn as previously; otherwise a series of arcs is drawn, which replicates the appearance of the wavy underlines in common programs such as Microsoft Word.

I have only implemented the wavy-line shape in the Java 9+ version of the class because I can see there's an open question whether it's even worth supporting Java 8 any more (and my opinion FWIW is "no").

I haven't added any tests – if they're required then I would appreciate advice on how to test a purely visual change like this.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/TextExt.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/TextExt.java
@@ -43,7 +43,7 @@ public class TextExt extends Text {
         styleables.add(StyleableProperties.UNDERLINE_COLOR);
         styleables.add(StyleableProperties.UNDERLINE_WIDTH);
         styleables.add(StyleableProperties.UNDERLINE_OFFSET);
-        styleables.add(StyleableProperties.UNDERLINE_WAVE);
+        styleables.add(StyleableProperties.UNDERLINE_WAVE_RADIUS);
         styleables.add(StyleableProperties.UNDERLINE_DASH_ARRAY);
         styleables.add(StyleableProperties.UNDERLINE_CAP);
 
@@ -82,8 +82,8 @@ public class TextExt extends Text {
             null, "underlineOffset", this, StyleableProperties.UNDERLINE_OFFSET
     );
 
-    private final StyleableObjectProperty<Number> underlineWave = new CustomStyleableProperty<>(
-            null, "underlineWave", this, StyleableProperties.UNDERLINE_WAVE
+    private final StyleableObjectProperty<Number> underlineWaveRadius = new CustomStyleableProperty<>(
+            null, "underlineWaveRadius", this, StyleableProperties.UNDERLINE_WAVE_RADIUS
     );
 
     private final StyleableObjectProperty<Number[]> underlineDashArray = new CustomStyleableProperty<>(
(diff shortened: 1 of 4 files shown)


richtextfx/src/main/java/org/fxmisc/richtext/TextExt.java
```diff
@@ -291,6 +333,16 @@ private static class StyleableProperties {
                 0, n -> n.underlineWidth
         );
 
+        private static final CssMetaData<TextExt, Number> UNDERLINE_OFFSET = new CustomCssMetaData<>(
+            "-rtfx-underline-offset", StyleConverter.getSizeConverter(),
+            0, n -> n.underlineOffset
+        );
+
+        private static final CssMetaData<TextExt, Number> UNDERLINE_WAVE = new CustomCssMetaData<>(
+            "-rtfx-underline-wave", StyleConverter.getSizeConverter(),
```
Jugen, line 342: Let's make this UNDERLINE_WAVE_RADIUS and "-rtfx-underline-wave-radius" please. I know it's more verbose but I think it adds clarity, since we've already got "-rtfx-underline-dash-array".

Jugen:
This is great, thanks a stack for this PR !

Jugen (commented):
Could you please for clarity, change all `wave` variables to `waveRadius` please.

Jugen:
Okay, so I've played around with the code you've submitted and propose the following changes:

1.  I noticed that there's a relationship between the wave radius and the offset, in that the offset needs to be increased as the radius gets larger. I think providing a reasonable default will help developers to not have to figure out another underline option if they don't need to. Suggested change to `ParagraphText.UnderlineAttributes`:
```java
Number waveNumber = text.getUnderlineWave();
waveRadius = waveNumber == null ? 0 : waveNumber.doubleValue();
Number offsetNumber = text.getUnderlineOffset();
offset = offsetNumber == null ? waveRadius*0.5 : offsetNumber.doubleValue();
// The larger the radius the bigger the offset needs to be, so
// a reasonable default is provided if no offset is specified.
```
2.  The range of useful wave radii is quite limited and with larger radii the underline is rather bumpy and not wavy. If however the X radius of the arcs is stretched out slightly then the range of useful radii increases and the underline is smoothed/stretched out as well. Suggested change to `TextFlowExt.getUnderlineShape`:
```java
for ( int ele = 2; ele < shape.length; ele += 5 )
{
    LineTo bl = (LineTo) shape[ele+1];
    LineTo br = (LineTo) shape[ele];

    double y = snapSizeY( br.getY() + offset - 2.5 );
    double leftx = snapSizeX( bl.getX() );

    if (waveRadius <= 0) {
        result.add(new MoveTo( leftx, y ));
        result.add(new LineTo( snapSizeX( br.getX() ), y ));
    }
    else {
        // For larger wave radii increase the X radius to stretch out the wave.
        double radiusX = waveRadius > 1 ? waveRadius * 1.25 : waveRadius;
        double rightx = snapSizeX( br.getX() - radiusX ); // Prevent overshooting.
        result.add(new MoveTo( leftx, y ));
        boolean sweep = true;
        while ( leftx < rightx ) {
            leftx += waveRadius * 2;
            // Note, had to remove code here otherwise underline ended in strange tail ?
            result.add(new ArcTo( radiusX, waveRadius, 0.0, leftx, y, false, sweep ));
            sweep = !sweep;
        }
    }
}
```

I would appreciate it if you could try out the above changes and see how it affects what you want.
If you have any concerns, differences, or recommendations please comment.

shoaniki:
Thank you for giving it a try! I agree on both counts -- the shallower waves look much nicer! -- and have pushed an update accordingly.

The only difference is that I have restored a version of the endpoint computation code, fixed to remove the assumption that the arc is defined by a circle.  I think it's necessary to clip the path this way, because otherwise larger radii will always either overflow or underflow.  I appreciate the code is a bit dense, but I wasn't able to find a more idiomatic way to do the clipping (which surprised me a bit -- I was expecting JavaFX to have that kind of feature, did I miss something?)

Jugen (commented):
Thanks for the effort you're putting into this !

PR #1145, opened 2022-11-08
Add Recaf to 'Who uses RichTextFX'

Col-E:
Been using RichTextFX for a little over 4 years now, [been quite happy with it](https://github.com/Col-E/Recaf).

image


Jugen:
That's great, thanks for the PR.

PR #1200, opened 2023-10-25
feat: line number highlight demo

impactCn:


Jugen:
Hi Sinsy, thank you so much for the pull request :-)
I'll take a look and maybe make some changes or improvements before merging ....

impactCn:
Sorry, I didn't notice the package name.
Thank you for your careful inspection.

impactCn:
All said and done, I don't think it's a good idea to remove the HBox outside. The outer HBox can control color, size, and position. The components inside come to the business. It might be better.

richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/LineNumberHighLightDemo.java
```diff
@@ -27,13 +25,7 @@ public void start(Stage primaryStage) {
...
         IntFunction<Node> arrowFactory = new LineNumberHighLightFactory( codeArea );
 
-        IntFunction<Node> graphicFactory = line -> {
-            HBox hbox = new HBox(arrowFactory.apply(line));
-            hbox.setAlignment(Pos.CENTER_LEFT);
-
-            return hbox;
-        };
-        codeArea.setParagraphGraphicFactory(graphicFactory);
+        codeArea.setParagraphGraphicFactory(arrowFactory);
```
impactCn, line 28: All said and done, I don't think it's a good idea to remove the HBox outside. The outer HBox can control color, size, and position. The components inside come to the business. It might be better.

Jugen: So it depends on what this demo is trying to show. There are two situations:
1. If you want to use an existing standard paragraph graphic factory without changing its code, but you want to customize the returned Node in some way then this code shows how to intercept/wrap the graphic factory so that you can do that.
2. If however you want to show how to create a paragraph graphic factory that uses styling to highlight which line is active then this code isn't needed for that and any additional customization can be done in the factory instead.

If you want, we can revert the change and add comments instead to explain this ?

impactCn: ok, I get it.

Issue #1224, opened 2024-05-14
Question: Is possible to detect when INSERT mode is used?


PavelTurk:
I have a `CodeArea` and I need to show user when insert mode is used (the mode after pressing `Insert` button). Could anyone say if `CodeArea` provides something for to detect current mode?

Jugen:
I don't think there is.

Jugen:
Ahh you are right, it's in [GenericStyledAreaBehavior (see line 306)](https://github.com/FXMisc/RichTextFX/blob/733a7f7f40778c55800ed86caddb2aa031baee06/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledAreaBehavior.java#L306) but currently a private boolean.

Unfortunately there's currently no reference to GenericStyledAreaBehavior in [GenericStyledArea (see line 819)](https://github.com/FXMisc/RichTextFX/blob/733a7f7f40778c55800ed86caddb2aa031baee06/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java#L819).

Anyway it looks like a straightforward type of change, would you like to submit a PR for this ?

Jugen:
Executing `gradle clean build` in the root folder of the project should work, and finish with no errors.

Jugen:
I'm using gradle 7.5.1, try that and see ....

Jugen:
Try setting the  `JAVA_HOME` environment variable.
I think it's something like:  `export JAVA_HOME="path to JDK 11 root"`

Jugen:
Okay, just submit your PR and I'll check it my side.

Jugen:
Well done. Interesting those tests pass on Windows but fail on Linux/MacOs ?

Jugen:
Thanks for the feedback, will have a look.

PR #1226, opened 2024-05-15
Add property to area to control overwrite mode (#1224)

PavelTurk:
With this property it is possible to control when user starts overwrite mode using Insert key


Jugen:
Thanks for the PR, have reviewed and will merge :-)

PR #1246, opened 2024-10-15
Align TextFlowExt with future getLayoutInfo

Jugen:
Refactored `TextFlowExt` to prepare for future API proposed by JavaFX PR openjdk/jfx#1596
by renaming and adding methods in `TextFlowLayout` to match those proposed for `LayoutInfo`.

When the proposed API is integrated and available in JavaFX GA the following steps can be taken:
1. Rename/refactor the `TextFlowExt.textLayout()` method to `getLayoutInfo()`
2. Then delete the `TextFlowExt.getLayoutInfo()` method and reference field.
3. Delete `TextFlowLayout` and `TextFlowSpan`.

Special thanks to @andy-goryachev-oracle

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/TextFlowExt.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/TextFlowExt.java
@@ -1,15 +1,12 @@
 package org.fxmisc.richtext;
 
-import static org.fxmisc.richtext.model.TwoDimensional.Bias.*;
-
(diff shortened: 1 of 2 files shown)


andy-goryachev-oracle:
Thank you @Jugen for verifying early versions of the API!  Please keep in mind that the API might also change during the review process.  Hope it gets into jfx24, but no guarantees.  Feel free to comment in https://github.com/openjdk/jfx/pull/1596

Issue #1273, opened 2025-04-24
What is the proper way to lazily add styles to paragraphs in RichTextFX?


PavelTurk:
Suppose I have a document with a huge number of paragraphs (e.g., 10,000,000+). It makes no sense to add styles to all paragraphs at once for performance reasons. What is the correct approach to:

- Apply styles only when paragraphs become visible (e.g., during scrolling).
- Avoid showing unstyled paragraphs (ensure styles are applied before rendering).

Is there a built-in mechanism in RichTextFX for this, or do I need a custom solution?

Jugen:
Yes, have a look at [JavaKeywordsDemo](https://github.com/FXMisc/RichTextFX/blob/master/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/JavaKeywordsDemo.java)

Jugen:
Yes :-)

Jugen:
I'll try and have a look next week .....

Jugen:
@PavelTurk Sorry for the delayed response. Last week didn't go as planned and then this issue took me some time to figure out.
Anyway, see PR #1277 to address the issue you show in the video clip you posted.

Jugen:
> [@Jugen](https://github.com/Jugen) Could you say, what is the proper way to update styles when this approach is used? For example when it is necessary to update only styles for paragraphs 1 - 100? I tried this way:
> 
> ```
> var green = Collections.singleton("green");
>         var button = new Button("Test");
>         button.setOnAction(e -> {
>             for (var i = 0; i < 100; i++) {
>                 var p = codeArea.getParagraphs().get(i);
>                 StyleSpansBuilder<Collection<String>> ssb = new StyleSpansBuilder<>();
>                 ssb.add(green, p.getText().length() + 1);
>                 p.restyle(0, ssb.create());
>             }
>             System.out.println("Updated");
>         });
> ```
> 
> but it didn't work. I also tried to use absolute position and `codeArea.setStyleSpans(startPos, styles);` but it didn't work either.

WRT your green button test above. It doesn't work because Paragraph is an immutable object, so `p.restyle( ... )` returns a new Paragraph object without changing the original. If you replace that line with `codeArea.setStyleSpans( i, 0, ssb.create() );` instead, it should work.

Jugen:
I share your sentiment, but the problem is that Paragraph is an immutable object, so p.restyle( ... ) returns a new Paragraph object without changing the original.

[...]

PR #1277, opened 2025-05-22
Added getVisibleParagraphIndexes method

Jugen:
Addresses #1273 where styles are not always being applied.

To apply styling to particular portion of text we need to have a position of where that text is. The trouble is that the changes/modifications supplied by _getVisibleParagraph_ don't provide a document paragraph index so it has to be queried with _firstVisibleParToAllParIndex_ which at that point in time returns stale values as [the scene graph hasn't been updated yet](https://github.com/FXMisc/Flowless/issues/123).

Here I've added a new method `getVisibleParagraphIndexes()` to _GenericStyledArea_ that returns a LiveList of **Tuple2<Integer,Paragraph>** items, where the Integer is the document index for the Paragraph.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
@@ -94,6 +94,7 @@ import org.reactfx.SuspendableNo;
 import org.reactfx.collection.LiveList;
 import org.reactfx.collection.SuspendableList;
 import org.reactfx.util.Tuple2;
+import org.reactfx.util.Tuples;
 import org.reactfx.value.Val;
 import org.reactfx.value.Var;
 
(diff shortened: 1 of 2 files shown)


richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
```diff
@@ -1548,6 +1548,20 @@ public final PS getParagraphStyleForInsertionAt(int pos) {
         }
     }
 
+    private UnaryOperator<Paragraph<PS, SEG, S>> visibleOnlyStyler;
+    /**
+     * This styler will only be applied to Paragraphs, just before being displayed.
```
Symeon94, line 1553: Cosmetic: I think the world "visible" is missing: "This style will only be applied to **visible** Paragraphs"

Symeon94 (approved):

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
```diff
@@ -1548,6 +1548,20 @@ public final PS getParagraphStyleForInsertionAt(int pos) {
         }
     }
 
+    private UnaryOperator<Paragraph<PS, SEG, S>> visibleOnlyStyler;
```
Symeon94, line 1551: Actually, I do have an issue with this one. For some cases that might be good enough (those where the style fully depend only on the content of the line).
In my scenario, I have a group comment, meaning that the style of a line depends on its position.
Would it be possible to provide the paragraph and its position?

Symeon94 (approved):

PR #1285, opened 2025-08-27
Fixes #1284 - JavaFX25 introduced getCaretShape to replace caretShape to fix

### LINKED ISSUE #1284: JFX25+: class org.fxmisc.richtext.TextFlowExt overrides final method javafx.scene.text.TextFlow.getUnderlineShape(II)[Ljavafx/scene/shape/PathElement;
Hi -- as per title, it seems TextFlow introduced a new method with a signature clash of a method in TextFlowExt.

https://github.com/openjdk/jfx/commit/04c5e40cc116cb42150572959b53d1e465700e0e

Both these may need a rename:

getUnderlineShape(int from, int to)
getCaretShape(int charIdx, boolean isLeading)

So as of now RichTextFX cannot be used with JFX25+ ....

Symeon94:
My only concern with this change is that by removing the method, if someone extended `TextFlowExt` and used the method, it will suddenly hit the one from JavaFX 25 which is slightly different.

That being said, there is no real fix for that, as JavaFX 25 will not go away.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/ParagraphText.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/ParagraphText.java
@@ -328,7 +328,7 @@ class ParagraphText<PS, SEG, S> extends TextFlowExt {
     }
 
     private void updateSingleCaret(CaretNode caretNode) {
-        PathElement[] shape = getLeadingCaretShape(getClampedCaretPosition(caretNode));
+        PathElement[] shape = caretShape(getClampedCaretPosition(caretNode), true);
         caretNode.getElements().setAll(shape);
     }
 
--- a/richtextfx/src/main/java/org/fxmisc/richtext/ParagraphText.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/ParagraphText.java
@@ -328,7 +328,7 @@ class ParagraphText<PS, SEG, S> extends TextFlowExt {
     }
 
     private void updateSingleCaret(CaretNode caretNode) {
-        PathElement[] shape = getCaretShape(getClampedCaretPosition(caretNode), true);
+        PathElement[] shape = getLeadingCaretShape(getClampedCaretPosition(caretNode));
         caretNode.getElements().setAll(shape);
     }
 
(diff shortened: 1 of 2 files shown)


richtextfx/src/main/java/org/fxmisc/richtext/ParagraphText.java
```diff
@@ -328,7 +328,7 @@ private void updateAllCaretShapes() {
     }
 
     private void updateSingleCaret(CaretNode caretNode) {
-        PathElement[] shape = getCaretShape(getClampedCaretPosition(caretNode), true);
+        PathElement[] shape = caretShape(getClampedCaretPosition(caretNode), true);
```
Jugen, line 331: I think this should stay as getCaretShape since caretShape is the one being deprecated ?

Symeon94: I think we should use `caretShape` for two reasons:
1. `getCaretShape` is slightly different in JavaFX25 than `caretShape` (I don't know why they changed the behaviour and declared the old one deprecated)
2. I want to keep retrocompatibility (if we do the change you suggest, the code will be broken for versions before JavaFX25+)

Symeon94:
@Jugen for your comment about `hitTest`, we are not fixing deprecated methods here (and there might be risk of breaking support for older versions of JavaFX by doing so).
The instructions when pushing PR mentionned to only fix one problem per PR, I think, if we want to change the deprecated items we should proceed on a different PR.

Symeon94:
You are right about that, I didn't test the compilation on Java 25 and I assume the error from the comment was correct, but I don't see a final here.
That being said, it's not a good practice to override concrete methods (which only happens because JavaFX introduced the same method name as we have here).
Now, I totally see the problem of removing getCaretShape as suddenly a user that would override `TextFlowExt` and use that method would suddenly point to the one in `TextFlow`.
I think both solutions (removing/not removing it) are wrong, each in their way. But I tend to favor removing it.

That being said, I don't think there is any urge to merge this PR, so I'll let you decide what you want to do. I'm ready to modify it if you have any preference.

Jugen:
I'll give it some thought, however I'm currently inclined to leave it as is.
Thanks for trying though, I appreciate the effort very much !

PR #1286, opened 2025-08-28
Clean java demo

Symeon94:
I needed to be able to test some cases of multi-line edit, so before doing that I needed a clean example of styling to test with. I used this opportunity to clean the Java style example to make it easier to understand.

I still have one TODO for some point which I'm not sure the use of.


Symeon94:
There was a few bugs that were fixed here:
* The last line was not styled
* Replacing text without changing a line size on a previously edited line would not update the style.

Symeon94:
@Jugen You can merge it

Issue #1301, opened 2025-12-11
JavaKeywordsDemo doesn't update syntax highlighting when multiple lines inserted in succession before each other


willplatt:
JavaKeywordsDemo doesn't update syntax highlighting after certain changes. This is because it tries to be efficient by only recomputing highlighting for lines that are changed, but it does this incorrectly.

Specifically, when multiple changes come through at once (due to `reduceSuccessions()`), the first change that was made to line x may now actually apply to some other line y due to changes that happened since. The program does not account for this and hence highlighting is recomputed for line x instead of line y. For example, a change on line 5 may in fact apply to line 8 by the time we compute syntax highlighting because 3 lines have since been inserted before the original line 5. But the current logic will recompute highlighting for line 5 instead of line 8.

The easiest way I have found to see this bug is by deleting a line (with its line ending included), then deleting the line after it, then the line after that. Then quickly undo several times to bring the lines back in under 500ms. Only the top line will receive syntax highlighting. For example, with the default text in the editor, place the cursor just after the `/` closing the mult-line comment and select down to just after the next `{` and delete this selection. Then repeat by selecting just after the multi-line comment down to the end of the single-line comment. Then repeat by selecting down to just after the `for ... {`. After rapidly hitting undo three times, the single-line comment and `for` line won't be highlighted (see image).

I'm about to submit a pull request to fix this bug.

<img width="600" height="299" alt="Image" src="https://github.com/user-attachments/assets/757d3fe7-2255-4e71-ad56-72690c3b774d" />

Symeon94:
That I saw, what I meant is if you could guide us through what you think happens when the code is executed with the steps you are describing.

willplatt:
To be clear, I'm not saying the three insertions are merged into one insertion. I'm saying that the three insertions are performed correctly one after another but that the subscribed listener is only called once with data from all three changes. The lambda in `reduceSuccessions()` is called twice, but it treats the changes as independent when they aren't.

Symeon94:
Thanks for the clarification, that is actually very helpful to understand.
> [...] it treats the changes as independent when they aren't.

It looks to me as 3 independent changes, for as long as they are executed in the right sequence.

But executing the code I finally get what you mean. This is what happens:
1. Change indexes are collected and merged until a pause of 500ms is seen (that's done in the `reduceSuccession`)
2. Once a pause in changes is seen, it uses the merged indexes to tell what line it must update the style of

[...]

PR #1302, opened 2025-12-11
Fixed JavaKeywordsDemo not updating highlighting on the correct lines when multiple changes happen in succession

willplatt:
This pull request fixes issue #1301 by accounting for the fact that later changes affect the line numbers of earlier changes.

This solution may be of help to those wanting a way to run updates only on lines that have been changed (e.g. for efficiency), whether that's for syntax highlighting or other operations. Issue #1273 seems related to this.

### DIFF
--- a/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/JavaKeywordsDemo.java
+++ b/richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/JavaKeywordsDemo.java
@@ -1,15 +1,17 @@
 package org.fxmisc.richtext.demo;
 
 import java.time.Duration;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
 
 import javafx.application.Application;
 import javafx.application.Platform;
 import javafx.scene.Scene;
 import javafx.scene.control.ContextMenu;
-import javafx.scene.control.IndexRange;
 import javafx.scene.control.MenuItem;
 import javafx.scene.input.KeyCode;
 import javafx.scene.input.KeyEvent;
@@ -19,8 +21,10 @@ import javafx.stage.Stage;
 import org.fxmisc.flowless.VirtualizedScrollPane;
 import org.fxmisc.richtext.CodeArea;
 import org.fxmisc.richtext.LineNumberFactory;
-import org.fxmisc.richtext.model.TwoDimensional.Bias;
+import org.fxmisc.richtext.model.PlainTextChange;
 import org.reactfx.Subscription;
+
+import static org.fxmisc.richtext.model.TwoDimensional.Bias.Backward;
 
 public class JavaKeywordsDemo extends Application {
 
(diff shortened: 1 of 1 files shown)


willplatt:
Actually it appears this still isn't correct because `offsetToPosition()` is relying on the latest version of the text, not the text at the time of the edit. I think what we need to do is find the offsets affected across all the changes first (accounting for their influence on each other), then convert these to line numbers.

willplatt:
@Jugen I understand that, but the `subscribe()` lambda shouldn't be called unless no changes have occurred during that wait period. Is the issue that that the timer runs on another thread, so it's possible for it to run out but just before it notifies the FX thread another change occurs to the text on the FX thread?

willplatt:
I don't see the problem in your example. In my code both events will be passed to the subscribed listener, which will see the order in which the changes were made and work out that the first change is no longer at position 19, it's at 21, while the second change is at 4. Then it computes the lines of these positions based on the final text, which is what we want.

richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/JavaKeywordsDemo.java
```diff
@@ -69,27 +75,32 @@ public void start(Stage primaryStage) {
...
-                // in the meantime combine additional changes into one set of paragraph ranges with no duplicates 
-                .reduceSuccessions( (a,b) -> { a.addAll( b ); return a; }, Duration.ofMillis(500) )
+                .reduceSuccessions(
+						// We can't just add b to a and return a because a gets changed somewhere else
+						(a,b) -> Stream.concat(a.stream(), b.stream()).toList(),
+		                Duration.ofMillis(5000) )
 
                 // run the following code block when previous stream emits an event
-                .subscribe( paragraphs ->
+                .subscribe(changes ->
```
Symeon94: That becomes a huge lambda which makes the code a bit more dirty. Can't we move it to its own method and potentially split it?

richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/JavaKeywordsDemo.java
```diff
@@ -105,6 +116,22 @@ public void start(Stage primaryStage) {
         primaryStage.setTitle("Java Keywords Demo");
         primaryStage.show();
     }
+	
+	private void shiftAllIndicesAtOrAboveThreshold(Set<IndexRange> indexRanges, int threshold, int shift) {
+		if (shift != 0) {
+			Set<IndexRange> updatedRanges = new HashSet<>();
+			indexRanges.removeIf(range -> {
+				if (range.getEnd() >= threshold) {
```
Symeon94: Naming of the different values here does not bring clarity on what it does.
Here are the points that should be clarified:
- `threshold` what is a threshold ?
- `shiftAllIndicesAtOrAboveThreshold` this can be clear if `threshold` is a clear concept
- `range.getEnd() >= threshold` should be a method telling us what it means conceptually that the end is above the threshold.

Symeon94 (changes requested):

willplatt:
@Symeon94 I agree that the code would be nicer if broken into small functions, but I'm aware others might have a different opinion. You're probably right that it would be better to `map()` the Set instead of using `removeIf()`. But I'm a little confused about your comment on "threshold". It's a word I've often seen used in programming or when talking about numbers in any field. Some dictionary definitions are "The quantitative point at which an action is triggered, especially a lower limit" and "The point that must be exceeded to begin producing a given effect or result or to elicit a response" ([source](https://www.wordnik.com/words/threshold)).

But I think first @Jugen needs to be convinced that my proposed changes work correctly.

Symeon94:
Ok, but you get the general idea of my comments:
1. Split methods to keep them short
2. Have significant naming for variables and methods

Breaking in small methods is a good way to convey meaning. In this case, you used comments describing what happens as replacement for breaking into smaller units. It is also easier to understand when methods do a single task.

As far as `threshold` is concerned, the word itself is not conveying meaning on the underlying reality. You kept the method naming generic, which is good. Re-reading your method, it seems to me that `threshold` is fine, but it is the method name that could be improved by referring to the range end as this is what is compared to the threshold. This small change would clarify a lot I think. I'm thinking of `shiftWhenIndexesEndAboveThreshold` or to keep it short `shiftWhenEndAboveThreshold`.

willplatt:
@Jugen @Symeon94 I tried to incorporate both your suggestions, so please review again and hopefully it is clearer now.

richtextfx-demos/src/main/java/org/fxmisc/richtext/demo/JavaKeywordsDemo.java
```diff
@@ -105,6 +97,91 @@ public void start(Stage primaryStage) {
         primaryStage.setTitle("Java Keywords Demo");
         primaryStage.show();
     }
+	
+	private List<PlainTextChange> concatChangeLists(List<PlainTextChange> firstChanges, List<PlainTextChange> secondChanges) {
+		List<PlainTextChange> combinedChangeList = new ArrayList<>(firstChanges);
```
Symeon94: I suspect this will allocate an `ArrayList` of the size of `firstChange` and then have to extend it to add secondChanges (while you know in advance the size of the list.
It's test code, so it's fine. I just hope nobody copy that code 😉.

Symeon94 (approved):
I had a look. If it was production code I would have added more comments (and I would probably have moved most of that logic in its own class to not fill the `JavaKeywordDemo`), but for this test class I guess this can go through.

I have to say, I'm having a hard time following the overall logic, I hope you have it right ;-)

So for me it can go through. Waiting for @Jugen input.

Symeon94 (approved):

willplatt:
By the way, I think processing several changes after a delay is a common thing one might want to do (as in #1273). Therefore, we ought to provide a simpler way to do this. One way could be to offer methods in `GenericStyledArea` that would be equivalent to `multiPlainChanges().reduceSuccessions(...)`. For example, `changedPositionsAfterDelay(Duration timeout)` and `changedParagraphsAfterDelay(Duration timeout)` that each return `AwaitingEventStream<Set<Integer>>`. The changed paragraphs would include paragraphs after an inserted newline character.

We would be including positions and paragraphs which haven't actually changed, but which have been edited in some way. For example if an `A` was changed to a `B` then back to an `A`, we report this position and line as having changed. Similarly, if `b\n` is inserted before the `c` in `a\nc`, the line with `c` is returned but hasn't actually changed. Alternative method names could be `editedPositionsAfterDelay`, `modifiedPositionsAfterDelay`, etc., but these may also suggest that the characters are different now from before.

I think this would be best as a separate pull request. Any thoughts?

Symeon94:
Ok, I didn't look in details. My comment's purpose was to highlight the fact that the `GenericStyledArea` doesn't limit itself to text, so one must be careful when implementing for that one (it has the concept of paragraph though). Also EOL is OS or file specific, so that must also be considered.

But problaby/maybe there is a way to implement these in a generic way.

Symeon94:
> I suppose I could have used an iterator and a counter ?

That's what I had in mind, indeed.

Symeon94:
I understand that currently this seem to be used with an array but if you don't change the signature there is no guarantee nobody won't change that by mistake in the future and introduce inefficiencies.
Here is the change in your method:
```Java
private List<PlainTextChange> concatChangeLists(List<PlainTextChange> firstChanges, List<PlainTextChange> secondChanges) {
    int firstSize = firstChanges.size();
    PlainTextChange lastChange = firstChanges.get(firstSize - 1);
    int secondSize = secondChanges.size();
    int secondIndex = 0;

    // try and merge the second changes into the first
    for (PlainTextChange secondChange : secondChanges) {
        Optional<PlainTextChange> merged = lastChange.mergeWith(secondChange);
        if (merged.isEmpty()) {
            break;
        }
        lastChange = merged.get();
        secondIndex++;
    }

    firstChanges.set(firstSize - 1, lastChange);

    if (secondIndex == secondSize) {
        // everything merged so return
        return firstChanges;
    }

    // there are unmerged changes so combine into single list
    List<PlainTextChange> combinedChangeList = new ArrayList<>(firstSize + secondSize - secondIndex);
    combinedChangeList.addAll(firstChanges);
    combinedChangeList.addAll(secondChanges.subList(secondIndex, secondSize));
    return combinedChangeList;
}
```

willplatt:
@Jugen Thanks for accepting my changes.

I've been busy so only just found time to come back to this. I will make a utility class for these two static methods and make things a little more efficient. Would `PlainTextChangeAggregator` be a good name? Or `DelayedPlainTextChangeUtil`?

PR #1306, opened 2025-12-23
[Refactoring] Adding test for `Paragraph`

Symeon94:
This PR only adds test for `Paragraph` in the goal for future refactoring.

The tests are covering the current behaviour, without any modification.

A few likely bugs were discovered and are mark as such in the test.


richtextfx/src/test/java/org/fxmisc/richtext/model/ParagraphTest.java
Symeon94: @Jugen quick overall question, do you think we could upgrade to JUnit 5 (it has some interesting organization features which could help make the tests clearer. I'm specifically thinking of *Parametrized tests* and *DisplayName* and *Nested test classes*

PR #1308, opened 2025-12-23
Update package to JUnit 5

Symeon94:
Updating test package to JUnit 5


Jugen:
lgtm, thanks so much for all your work and input so far.

PR #1304, opened 2025-12-15
[Refactoring] Extract and test SelectionChange

Symeon94:
In this PR I extracted the content of the selection change update and covered it with UT.

PR #1309, opened 2025-12-23
Refactor caret and selection code

Symeon94:
After the merged PR #1303 and #1304 the code is merged.

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/CaretPositionChange.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/CaretPositionChange.java
@@ -39,27 +39,25 @@ class CaretPositionChange {
     private void applyFor(PlainTextChange plainTextChange) {
         int netLength = plainTextChange.getNetLength();
         if (netLength != 0) {
-            int changePosition = plainTextChange.getPosition();
-            if (position == changePosition) {
+            int changeStart = plainTextChange.getPosition();
+            if (position == changeStart) {
                 // If it's net removal, we don't change anything
                 position += Math.max(0, netLength);
             }
             else {
-                position = applyChange(position, changePosition, netLength);
+                int changeEnd = changeStart + Math.abs(netLength);
+                position = applyChange(position, changeStart, changeEnd, netLength);
             }
         }
     }
 
-    private static int applyChange(int position, int changeStart, int netLength) {
-        // Position is after the end of the change
-        if(position >= changeStart + Math.abs(netLength)) {
+    private static int applyChange(int position, int changeStart, int changeEnd, int netLength) {
+        if(position >= changeEnd) {
             position += netLength;
         }
-        // Position is before the end but after the start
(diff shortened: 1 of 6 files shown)


Symeon94:
Note that this code is very similar and I think it could be unified. Essentially the selection change of the caret should be the same as changing an interval where `start == end`. The problem is that unifying these two classes will change behaviour. I think though, the impossibility to merge is due to issue in on the two part of the logic.

richtextfx/src/main/java/org/fxmisc/richtext/CaretPositionChange.java
```diff
@@ -5,40 +5,48 @@
...
-            }
+    /**
+     * <ul>
+     * <li>If the position is after the interval of the change, the caret moves left/right if it's a net removal/addition.</li>
+     * <li>If it's within the change interval, the caret move back at the start of the change.</li>
+     * <li>If it's equal to the start position, the caret is moved at the end of the changed (it cannot be moved backwards
+     * if it's a removal of text, hence capping to 0).</li>
+     * </ul>
+     */
+    private static int applyChange(int position, int changeStart, int changeEnd, int netLength) {
```
Jugen, line 41: The following are just my preferences to making the code easier to understand. If you have strong reasons for not using them, that's fine:

1. remove static: it seems unnecessary, as it's not for instance being used by a test. Also see next ....
2. change return to void: position is a field so update it directly. It's confusing, for me a least, when the value is being changed here, then returned, only to update the field by the caller. The only reason it's like this at the moment is because the method is static.
3. remove the position parameter: it's a field variable so doesn't need to be passed all the time. Again, the only reason it's like this is because the method is static.
4. if you decide to keep the position parameter then please give it a different name to distinguish it from the field variable.

Symeon94: I think both a method or a static function are good solutions here, I don't have a strong preference. I'm totally ok changing to a method (and your comments 2/3/4 makes sense with that change).

I'll do ti.

To add some context, I initially extracted this bit of code in `CaretPositionChange` and `SelectionChange` to show that it was duplicated code. Then I adapted with the last bit which cannot be done in  `SelectionChange`:
```Java
// ...
else if(position == changeStart) {
    position += Math.max(0, netLength);
}
```

Symeon94: Actually, sorry to contradict myself, I just made the change and then looked at `SelectionChange` where I cannot make it a method (because it applies to two different values). So I propose that we keep the static to keep evident the close correspondance between the code in these two classes, if that's ok for you.

Jugen:
I think you need to rebase or merge from master ?

Symeon94:
I already did the pull from master normally (and I see the PR states there are no conflicts)

Jugen:
Ahh, seems my side was out of sync for some reason - strange.

Issue #1310, opened 2025-12-29
Restyle a paragraph up to the end throws indexes off


Symeon94:
I was working on refactoring the `Paragraph` class after having added tests and I looked at the bugs I highlighted in the PR adding the tests. I think it deserves its own issue, that's why I'm creating it here.

I'm refactoring the method, but I'm not modifying its behaviour for now, I will do that on a separate PR.

I have yet to figure out what is happening.

If you restyle a Paragraph up to `len - 1` you don't have any issue, but you do if you restyle up to `len`. As you can see in below test, the indexes in that cases are way off.

```Java
checkStyle(p3, 18, new String[] {"text", "unknown", "keyword", "text"},
	0, 3, // text
	3, 10, // unknown
	10, 12, // keyword
	12, 18); // text

// Restyle up to the end
checkStyle(p3.restyle(11, 17, "out"), 18, new String[] {"text", "unknown", "keyword", "out", "text"},
	0, 3,
	3, 10,
	10, 11,
	11, 17,
	17, 18);

// Restyle up to the end - Bug: indexes are totally of
checkStyle(p3.restyle(11, 18, "out"), 18, new String[] {"text", "unknown", "keyword", "out"},
	0, 3,
	3, 10,
	0, 1,
	0, 7);
```

Symeon94:
Ok, the issue is that the append is actually creating a new object of type `AppendedSpans` which is keeping the current `StyleSpans` and putting a new one next to it.
When you are getting the last `StyleSpan` it returns that appended value and it has the indexes it had.

It could be that the code does not has an issue with that, but it creates a flaw in the logic. I would be in favor of modifying the appended value to properly match the indexes.

Symeon94:
The problem with fixing that one is that it is likely that the styling use only the length, so it is likely that I will not be able to find an actual user facing bug. But it is a bomb waiting to explode the day anybody writes code expecting the start position to work.
So either;
* We fix it and hope it doesn't blow somewhere because some code used the bug as an assumption
* We remove start position fully (or only keep the first style position)
* We keep the bug and hope for the best

I prefer solution 1.

PR #1311, opened 2025-12-29
Cleanup `Paragraph` class after having covered it with UT

Symeon94:
As per title. Tests were previously added and merged and this branch is used to clean it up.

Note that I did change a little behaviour here, the trim method is now capping the end value to not allow negative number.
```Java
public Paragraph<PS, SEG, S> trim(int end) {
    end = Math.max(0, end);
    // ...
}
```

### DIFF
--- a/richtextfx/src/main/java/org/fxmisc/richtext/model/Paragraph.java
+++ b/richtextfx/src/main/java/org/fxmisc/richtext/model/Paragraph.java
@@ -261,19 +261,19 @@ public final class Paragraph<PS, SEG, S> {
     }
 
     public Paragraph<PS, SEG, S> trim(int length) {
-        if(length >= length()) {
-            return this;
-        } else {
-            Position pos = navigator.offsetToPosition(length, Backward);
-            int segIdx = pos.getMajor();
-            List<SEG> segs = new ArrayList<>(segIdx + 1);
-            segs.addAll(segments.subList(0, segIdx));
-            segs.add(segmentOps.subSequence(segments.get(segIdx), 0, pos.getMinor()));
-            if (segs.isEmpty()) {
-                segs.add(segmentOps.createEmptySeg());
-            }
-            return new Paragraph<>(paragraphStyle, segmentOps, segs, styles.subView(0, length));
+        length = Math.max(0, length);
+        if(length < length()) {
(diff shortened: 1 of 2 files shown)


richtextfx/src/main/java/org/fxmisc/richtext/model/Paragraph.java
```diff
@@ -248,20 +260,13 @@ public Paragraph<PS, SEG, S> subSequence(int start, int end) {
...
-            int segIdx = pos.getMajor();
-            List<SEG> segs = new ArrayList<>(segIdx + 1);
-            segs.addAll(segments.subList(0, segIdx));
-            segs.add(segmentOps.subSequence(segments.get(segIdx), 0, pos.getMinor()));
-            if (segs.isEmpty()) {
-                segs.add(segmentOps.createEmptySeg());
-            }
-            return new Paragraph<>(paragraphStyle, segmentOps, segs, styles.subView(0, length));
+    public Paragraph<PS, SEG, S> trim(int end) {
+        end = Math.max(0, end);
```
Symeon94, line 264: This is a behaviour change (you can see the adaptation in the test for it), any trim with negative number should equal to trimming with 0 (it was already the case for some negative values, but not all).

richtextfx/src/test/java/org/fxmisc/richtext/model/ParagraphTest.java
```diff
@@ -241,8 +257,18 @@ public void multipleStyle() {
...
         checkStyle(p3, 18, new String[] {"text", "unknown", "keyword", "text"}, 0, 3, 3, 10, 10, 12, 12, 18);
 
+        // Restyle up to the end
+        checkStyle(p3.restyle(11, 17, "out"), 18,
+                new String[] {"text", "unknown", "keyword", "out", "text"},
+                0, 3, 3, 10, 10, 11, 11, 17, 17, 18);
+
+        // Restyle up to the end
+        // Bug
+        checkStyle(p3.restyle(11, 18, "out"), 18,
```
Symeon94, line 267: I thought the bug was with out-of-bound indexes, but it is actually a bug when you go up to the length. I created #1310 to handle that one.

PR #1318, opened 2026-07-17
Improve layout performance of folded paragraphs

### LINKED ISSUE #1317: Scrolling past large collapsed paragraphs can lead to slight visual stutter
# Overview

Scrolling in a `GenericStyledArea` _(Or subtypes like `CodeArea`)_ with many paragraphs folded can lead to the scrolling _"stuttering"_. The styled are passes the full document to `VirtualFlow.createVertical(...)`.

The virtual flow ends up walking this list whenever it fills the viewport during layout/scrolling. Each paragraph requires computing the cell's height & materializing the layout. Folded paragraphs do collapse to zero visual height (`ParagraphBox.computePrefHeight` returns `0.0` when `isFolded` is true), but the virtual flow still performs the iteration and materialization work for every paragraph in the passed list.

VisualVM screenshot of CPU profiling after scrolling for ~30 seconds over a large folded area:

Image

## Possible fix

Pass in a `FilteredList<Paragraph<PS, SEG, S>>` which only includes visible paragraphs. The filtering predicate would be `paragraph -> !getFoldStyleCheck().test(paragraph.getParagraphStyle())` which requires pulling the contract for `getFoldStyleCheck()` down to `GenericStyledArea` _(but it can be no-op by default since children still would override it)_. This will require a little extra work also doing some index translation now that some paragraphs are effectively gone in the model. We wouldn't want to break things like line number counts. But it should be worth it to reduce the work done by the virtual flow.

Col-E:
Addressing #1317 - Description + video in the linked issue.

### DIFF
--- a/richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
+++ b/richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
@@ -2,6 +2,8 @@ package org.fxmisc.richtext.api;
 
 import javafx.geometry.BoundingBox;
 import javafx.geometry.Bounds;
+import javafx.scene.control.Label;
+import javafx.scene.layout.Region;
 import javafx.stage.Stage;
 import org.fxmisc.richtext.CaretNode;
 import org.fxmisc.richtext.InlineCssTextAreaAppTest;
@@ -111,13 +113,14 @@ public class FoldedParagraphTests extends InlineCssTextAreaAppTest {
     public void paragraph_graphic_access_rejects_folded_paragraph() {
         // fold 'one' paragraph, so that 'zero' and 'two' are visible
         interact(() -> {
+            area.setParagraphGraphicFactory(i -> new Label("Line: " + i));
             area.replaceText("zero\none\ntwo");
             area.foldParagraphs(0, 1);
         });
 
-        // getting/creating graphics for the folded paragraph throws an exception, since it is not visible.
-        assertThrows(IllegalArgumentException.class, () -> area.getParagraphGraphic(1));
-        assertThrows(IllegalArgumentException.class, () -> area.recreateParagraphGraphic(1));
+        // getting/creating graphics for the folded paragraph is no-op (returns null and does not throw an exception).
+        assertNull(area.getParagraphGraphic(1));
+        assertDoesNotThrow(() -> area.recreateParagraphGraphic(1));
 
         // getting/creating graphics for the visible paragraphs works as expected.
         interact(() -> {
--- a/richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
+++ b/richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
@@ -22,22 +22,30 @@ public class FoldedParagraphTests extends InlineCssTextAreaAppTest {
     }
 
     @Test
-    public void line_index_works_for_visible_paragraph_and_rejects_folded_paragraph() {
+    public void line_index_works_for_visible_and_folded_paragraphs() {
         interact(() -> {
             // fold 'second' paragraph, so that 'first' and 'third' are visible
             area.replaceText("first\nsecond\nthird");
             area.foldParagraphs(0, 1);
 
-            // verify pos[0, 0] is in the first line of the first paragraph
+            // verify lineIndex[par, 0] is zero for all paragraphs
+            // - visible content is short enough to fit on one line
+            // - folded content is not realized and thus there is no line wrapping to be done
             assertEquals(0, area.lineIndex(0, 0));
+            assertEquals(0, area.lineIndex(1, 0));
+            assertEquals(0, area.lineIndex(2, 0));
 
             // verify current line start/end range is the length of 'first' (5 chars)
             area.moveTo(0, 2);
+            assertEquals(0, area.getCurrentParagraph());
             assertEquals(0, area.getCurrentLineStartInParargraph());
             assertEquals(5, area.getCurrentLineEndInParargraph());
 
-            // verify that lineIndex rejects folded paragraph
-            assertThrows(IllegalArgumentException.class, () -> area.lineIndex(1, 0));
+            // moving to the folded paragraph will skip forward to the next visible paragraph, which is 'third' (5 chars)
+            area.moveTo(1, 2);
+            assertEquals(2, area.getCurrentParagraph());
+            assertEquals(0, area.getCurrentLineStartInParargraph());
+            assertEquals(5, area.getCurrentLineEndInParargraph());
         });
     }
 
--- a/richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
+++ b/richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
@@ -12,6 +12,7 @@ import org.fxmisc.richtext.CaretNode;
 import org.fxmisc.richtext.InlineCssTextAreaAppTest;
 import org.junit.Test;
 
+import java.util.Optional;
 import java.util.stream.IntStream;
 
 import static javafx.scene.input.MouseButton.PRIMARY;
@@ -81,18 +82,24 @@ public class FoldedParagraphTests extends InlineCssTextAreaAppTest {
 
     @Test
     public void character_bounds_are_empty_for_folded_paragraph() {
-        // fold 'one' paragraph, so that 'zero' and 'two' are visible
+        // fold 'bbb' paragraph, so that 'aaa' and 'ccc' are visible
         interact(() -> {
-            area.replaceText("zero\none\ntwo");
+            area.replaceText("aaa\nbbb\nccc");
             area.foldParagraphs(0, 1);
 
-            // the folded paragraph should not have character bounds on screen, but the visible paragraphs should.
-            int start = area.getAbsolutePosition(0, 0);
-            assertTrue(area.getCharacterBoundsOnScreen(start, start + 1).isPresent());
-            start = area.getAbsolutePosition(1, 0);
-            assertFalse(area.getCharacterBoundsOnScreen(start, start + 1).isPresent());
-            start = area.getAbsolutePosition(2, 0);
-            assertTrue(area.getCharacterBoundsOnScreen(start, start + 1).isPresent());
+            // the folded paragraph (lineIndex 1) should not have character bounds on screen, but the visible paragraphs should.
+            int foldedLine = 1;
+            for (int lineIndex = 0; lineIndex <= 2; lineIndex++) {
+                for (int charIndex = 0; charIndex <= 2; charIndex++) {
+                    int pos = area.getAbsolutePosition(lineIndex, charIndex);
+                    Optional<Bounds> charBounds = area.getCharacterBoundsOnScreen(pos, pos + 1);
+                    if (lineIndex == foldedLine) {
+                        assertFalse(charBounds.isPresent());
+                    } else {
+                        assertTrue(charBounds.isPresent());
+                    }
+                }
+            }
         });
     }
 
--- a/richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
+++ b/richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
@@ -118,6 +118,14 @@ public class FoldedParagraphTests extends InlineCssTextAreaAppTest {
         // getting/creating graphics for the folded paragraph throws an exception, since it is not visible.
         assertThrows(IllegalArgumentException.class, () -> area.getParagraphGraphic(1));
         assertThrows(IllegalArgumentException.class, () -> area.recreateParagraphGraphic(1));
+
+        // getting/creating graphics for the visible paragraphs works as expected.
+        interact(() -> {
+            assertNotNull(area.getParagraphGraphic(0));
+            assertDoesNotThrow(() -> area.recreateParagraphGraphic(0));
+            assertNotNull(area.getParagraphGraphic(2));
+            assertDoesNotThrow(() -> area.recreateParagraphGraphic(2));
+        });
     }
 
     @Test
--- a/richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
+++ b/richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
@@ -1,0 +1,219 @@
+package org.fxmisc.richtext.api;
+
+import javafx.geometry.BoundingBox;
+import javafx.geometry.Bounds;
+import javafx.stage.Stage;
+import org.fxmisc.richtext.CaretNode;
+import org.fxmisc.richtext.InlineCssTextAreaAppTest;
+import org.junit.Test;
+
+import java.util.stream.IntStream;
+
+import static javafx.scene.input.MouseButton.PRIMARY;
+import static org.testfx.util.WaitForAsyncUtils.asyncFx;
+import static org.junit.jupiter.api.Assertions.*;
+
+public class FoldedParagraphTests extends InlineCssTextAreaAppTest {
+
+    @Override
+    public void start(Stage stage) throws Exception {
+        super.start(stage);
+        stage.setHeight(100);
+    }
+
+    @Test
+    public void line_index_works_for_visible_paragraph_and_rejects_folded_paragraph() {
+        interact(() -> {
+            // fold 'second' paragraph, so that 'first' and 'third' are visible
+            area.replaceText("first\nsecond\nthird");
+            area.foldParagraphs(0, 1);
+
+            // verify pos[0, 0] is in the first line of the first paragraph
+            assertEquals(0, area.lineIndex(0, 0));
+
+            // verify current line start/end range is the length of 'first' (5 chars)
+            area.moveTo(0, 2);
+            assertEquals(0, area.getCurrentLineStartInParargraph());
+            assertEquals(5, area.getCurrentLineEndInParargraph());
+
+            // verify that lineIndex rejects folded paragraph
+            assertThrows(IllegalArgumentException.class, () -> area.lineIndex(1, 0));
+        });
+    }
+
+    @Test
+    public void paragraph_bounds_are_empty_for_folded_paragraph_and_present_for_visible_paragraph() {
+        // fold 'two' paragraph, so that 'zero', 'one', and 'three' are visible
+        interact(() -> {
+            area.replaceText("zero\none\ntwo\nthree");
+            area.foldParagraphs(1, 2);
+            area.showParagraphAtTop(3);
+        });
+
+        // the folded paragraph should not have bounds on screen, but the visible paragraphs should.
+        assertTrue(area.getParagraphBoundsOnScreen(0).isPresent());
+        assertTrue(area.getParagraphBoundsOnScreen(1).isPresent());
+        assertFalse(area.getParagraphBoundsOnScreen(2).isPresent());
+        assertTrue(area.getParagraphBoundsOnScreen(3).isPresent());
+
+        // visible paragraph filtering will only go up to three paragraphs since one of them is folded.
+        assertNotNull(area.getVisibleParagraphBoundsOnScreen(0));
+        assertNotNull(area.getVisibleParagraphBoundsOnScreen(1));
+        assertNotNull(area.getVisibleParagraphBoundsOnScreen(2));
+        assertThrows(IndexOutOfBoundsException.class, () -> area.getVisibleParagraphBoundsOnScreen(3));
+
+        // since 'two' is folded then 'three' is the third visible paragraph (but fourth in the unfiltered list).
+        assertEquals(area.getVisibleParagraphBoundsOnScreen(2), area.getParagraphBoundsOnScreen(3).get());
+    }
+
+    @Test
+    public void character_bounds_are_empty_for_folded_paragraph() {
+        // fold 'one' paragraph, so that 'zero' and 'two' are visible
+        interact(() -> {
+            area.replaceText("zero\none\ntwo");
+            area.foldParagraphs(0, 1);
+
+            // the folded paragraph should not have character bounds on screen, but the visible paragraphs should.
+            int start = area.getAbsolutePosition(0, 0);
+            assertTrue(area.getCharacterBoundsOnScreen(start, start + 1).isPresent());
+            start = area.getAbsolutePosition(1, 0);
+            assertFalse(area.getCharacterBoundsOnScreen(start, start + 1).isPresent());
+            start = area.getAbsolutePosition(2, 0);
+            assertTrue(area.getCharacterBoundsOnScreen(start, start + 1).isPresent());
+        });
+    }
+
+    @Test
+    public void caret_bounds_are_empty_for_caret_in_folded_paragraph() {
+        // fold 'one' paragraph, so that 'zero' and 'two' are visible
+        interact(() -> {
+            area.replaceText("zero\none\ntwo");
+            area.foldParagraphs(0, 1);
+        });
+
+        // create a caret in the folded paragraph, which should not have bounds on screen.
+        CaretNode caret = new CaretNode("folded caret", area, area.getAbsolutePosition(1, 0));
+        interact(() -> {
+            assertFalse(area.getCaretBoundsOnScreen(caret).isPresent());
+            caret.dispose();
+        });
+    }
+
+    @Test
+    public void paragraph_graphic_access_rejects_folded_paragraph() {
+        // fold 'one' paragraph, so that 'zero' and 'two' are visible
+        interact(() -> {
+            area.replaceText("zero\none\ntwo");
+            area.foldParagraphs(0, 1);
+        });
+
+        // getting/creating graphics for the folded paragraph throws an exception, since it is not visible.
+        assertThrows(IllegalArgumentException.class, () -> area.getParagraphGraphic(1));
+        assertThrows(IllegalArgumentException.class, () -> area.recreateParagraphGraphic(1));
+    }
+
+    @Test
+    public void hit_on_visible_paragraph_after_folded_paragraph_uses_source_index() throws Exception {
+        // fold 'two' paragraph, make 'two' the top visible paragraph.
+        interact(() -> {
+            area.replaceText("zero\none\ntwo\nthree");
+            area.foldParagraphs(1, 2);
+        });
+
+        // click on the first character of the 'three' paragraph,
+        int start = area.getAbsolutePosition(3, 0);
+        Bounds bounds = asyncFx(() -> area.getCharacterBoundsOnScreen(start, start + 1).get()).get();
+        moveTo(bounds).clickOn(PRIMARY);
+
+        // verify that the caret is now in the 'three' paragraph, which is the fourth paragraph in the source text.
+        assertEquals(3, area.getCurrentParagraph());
+        assertEquals(0, area.getCaretColumn());
+    }
+
+    @Test
+    public void show_paragraph_in_viewport_resolves_a_folded_target_to_a_visible_paragraph() {
+        interact(() -> {
+            // create a large number of paragraphs, fold the first 50 paragraphs
+            area.replaceText(IntStream.range(0, 100).mapToObj(i -> "paragraph " + i).reduce((a, b) -> a + "\n" + b).get());
+            area.foldParagraphs(1, 50);
+
+            // scroll down
+            area.showParagraphAtTop(99);
+
+            // scroll back up to a folded paragraph, which should resolve to paragraph index 1 since that is
+            // the first visible paragraph in the folded range.
+            area.showParagraphInViewport(2);
+        });
+
+        // after scrolling to the folded paragraph index 2 the first visible paragraph index 1 should be at the top of the viewport.
+        assertVisibleParagraph(1);
+    }
+
+    @Test
+    public void show_paragraph_region_resolves_a_folded_target_to_a_visible_paragraph() {
+        interact(() -> {
+            // same idea as the test before, but with showParagraphRegion
+            area.replaceText(IntStream.range(0, 100).mapToObj(i -> "paragraph " + i).reduce((a, b) -> a + "\n" + b).get());
+            area.foldParagraphs(1, 50);
+            area.showParagraphAtTop(99);
+            area.showParagraphRegion(2, new BoundingBox(0, 0, 1, 1));
+        });
+
+        // after scrolling to the folded paragraph index 2 the first visible paragraph index 1 should be at the top of the viewport.
+        assertVisibleParagraph(1);
+    }
+
+    @Test
+    public void show_paragraph_at_center_resolves_a_folded_target_to_a_visible_paragraph() {
+        interact(() -> {
+            // same idea as the test before, but with showParagraphAtCenter
+            area.replaceText(IntStream.range(0, 100).mapToObj(i -> "paragraph " + i).reduce((a, b) -> a + "\n" + b).get());
+            area.foldParagraphs(1, 50);
+            area.showParagraphAtTop(99);
+            area.showParagraphAtCenter(2);
+        });
+
+        // after scrolling to the folded paragraph index 2 the first visible paragraph index 1 should be at the top of the viewport.
+        // this is mainly because you can't really center an index that close to the start of the document,
+        // so of course the first visible paragraph is at the top of the viewport.
+        assertVisibleParagraph(1);
+    }
+
+    @Test
+    public void navigation_skips_folded_paragraphs_in_both_directions() {
+        interact(() -> {
+            // fold 'two' and 'three' paragraphs, so that 'zero', 'one', and 'four' are visible
+            area.replaceText("zero\none\ntwo\nthree\nfour");
+            area.foldParagraphs(1, 3);
+
+            // verify that moving forward to the folded paragraph skips to the next visible paragraph
+            area.moveTo(2, 0);
+            assertEquals(4, area.getCurrentParagraph());
+
+            // verify that moving backward to the folded paragraph skips to the previous visible paragraph
+            area.moveTo(3, 0);
+            assertEquals(1, area.getCurrentParagraph());
+        });
+    }
+
+    @Test
+    public void auto_height_excludes_folded_paragraphs() {
+        // if we set auto-height we effectively verify GenericStyledArea#computePrefHeight counts only visible paragraphs.
+        interact(() -> {
+            area.setAutoHeight(true);
+            area.replaceText("zero\none\ntwo\nthree");
+        });
+
+        double unfoldedHeight = area.prefHeight(-1);
+        interact(() -> area.foldParagraphs(0, 2));
+        double foldedHeight = area.prefHeight(-1);
+
+        // the folded height should be less than the unfolded height since we folded some paragraphs.
+        assertTrue(foldedHeight < unfoldedHeight);
+    }
+
+    private void assertVisibleParagraph(int allParagraphIndex) {
+        assertTrue(area.allParToVisibleParIndex(allParagraphIndex).isPresent(),
+                "Paragraph " + allParagraphIndex + " should be visible");
+    }
+}
--- a/richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
+++ b/richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
@@ -1,8 +1,11 @@
 package org.fxmisc.richtext.api;
 
+import javafx.event.Event;
 import javafx.geometry.BoundingBox;
 import javafx.geometry.Bounds;
 import javafx.scene.control.Label;
+import javafx.scene.input.KeyCode;
+import javafx.scene.input.KeyEvent;
 import javafx.scene.layout.Region;
 import javafx.stage.Stage;
 import org.fxmisc.richtext.CaretNode;
@@ -216,6 +219,35 @@ public class FoldedParagraphTests extends InlineCssTextAreaAppTest {
     }
 
     @Test
+    public void navigation_skips_folded_paragraphs_in_both_directions_key_events() {
+        // variant of the prior test, but with key events instead of moveTo
+        interact(() -> {
+            // fold 'two' and 'three' paragraphs, so that 'zero', 'one', and 'four' are visible
+            area.replaceText("zero\none\ntwo\nthree\nfour");
+            area.foldParagraphs(1, 3);
+
+            // verify that moving forward to the folded paragraph skips to the next visible paragraph
+            assertEquals(1, area.getCurrentParagraph());
+            assertEquals(5, area.getCaretPosition());
+            area.fireEvent(createKeyEvent(KeyCode.RIGHT));
+            assertEquals(6, area.getCaretPosition());
+            area.fireEvent(createKeyEvent(KeyCode.RIGHT));
+            assertEquals(7, area.getCaretPosition());
+            area.fireEvent(createKeyEvent(KeyCode.RIGHT));
+            assertEquals(8, area.getCaretPosition());
+            assertEquals(1, area.getCurrentParagraph());
+            area.fireEvent(createKeyEvent(KeyCode.RIGHT));
+            assertEquals(19, area.getCaretPosition());
+            assertEquals(4, area.getCurrentParagraph());
+
+            // verify that moving backward to the folded paragraph skips to the previous visible paragraph
+            area.fireEvent(createKeyEvent(KeyCode.LEFT));
+            assertEquals(8, area.getCaretPosition());
+            assertEquals(1, area.getCurrentParagraph());
+        });
+    }
+
+    @Test
     public void auto_height_excludes_folded_paragraphs() {
         // if we set auto-height we effectively verify GenericStyledArea#computePrefHeight counts only visible paragraphs.
         interact(() -> {
@@ -234,5 +266,9 @@ public class FoldedParagraphTests extends InlineCssTextAreaAppTest {
     private void assertVisibleParagraph(int allParagraphIndex) {
         assertTrue(area.allParToVisibleParIndex(allParagraphIndex).isPresent(),
                 "Paragraph " + allParagraphIndex + " should be visible");
+    }
+
+    private static Event createKeyEvent(KeyCode keyCode) {
+        return new KeyEvent(KeyEvent.KEY_PRESSED, keyCode.getChar(), "unset", keyCode, false, false, false ,false);
     }
 }
--- a/richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/ParagraphIndexMappingTests.java
+++ b/richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/ParagraphIndexMappingTests.java
@@ -169,7 +169,7 @@ public class ParagraphIndexMappingTests extends InlineCssTextAreaAppTest {
         for (int i = 1; i < paragraphs.size(); i++) {
             final int j = i;
             assertTrue(area.allParToVisibleParIndex(i).isEmpty());
-            assertEquals(-1, area.visibleParToAllParIndex(j));
+            assertThrows(IllegalArgumentException.class, () -> area.visibleParToAllParIndex(j));
         }
     }
 
@@ -179,8 +179,8 @@ public class ParagraphIndexMappingTests extends InlineCssTextAreaAppTest {
              assertEquals(Optional.empty(), area.allParToVisibleParIndex(-1));
              assertEquals(Optional.empty(), area.allParToVisibleParIndex(area.getParagraphs().size()));
 
-             assertEquals(-1, area.visibleParToAllParIndex(-1));
-             assertEquals(-1, area.visibleParToAllParIndex(Integer.MAX_VALUE));
+            assertThrows(IllegalArgumentException.class, () -> area.visibleParToAllParIndex(-1));
+            assertThrows(IllegalArgumentException.class, () -> area.visibleParToAllParIndex(Integer.MAX_VALUE));
         });
     }
 
--- a/richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/ParagraphIndexMappingTests.java
+++ b/richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/ParagraphIndexMappingTests.java
@@ -3,12 +3,13 @@ package org.fxmisc.richtext.api;
 import javafx.stage.Stage;
 import org.fxmisc.richtext.InlineCssTextAreaAppTest;
 import org.fxmisc.richtext.TextBuildingUtils;
+import org.fxmisc.richtext.model.Paragraph;
 import org.junit.Test;
+import org.reactfx.collection.LiveList;
 
 import java.util.Optional;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.*;
 
 public class ParagraphIndexMappingTests extends InlineCssTextAreaAppTest {
 
@@ -53,7 +54,6 @@ public class ParagraphIndexMappingTests extends InlineCssTextAreaAppTest {
 
         interact(() -> area.showParagraphAtBottom(LAST_PAR_INDEX));
         assertEquals(LAST_PAR_INDEX, area.visibleParToAllParIndex(area.getVisibleParagraphs().size() - 1));
-
     }
 
     @Test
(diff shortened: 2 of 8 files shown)


richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
```diff
@@ -776,8 +783,14 @@ public GenericStyledArea(
...
         // Initialize content
-        virtualFlow = VirtualFlow.createVertical(
+        foldStyleCheck = getFoldStyleCheck();
+        virtualFlowParagraphs = new FilteredList<>(
                 getParagraphs(),
+                paragraph -> !foldStyleCheck.test(paragraph.getParagraphStyle())
+        );
+
+        virtualFlow = VirtualFlow.createVertical(
+                virtualFlowParagraphs,
```
Symeon94, line 807: Why is there a need to keep virtualFlowParagraphs as an attribute of `GenericStyledArea`? The `virtualFlow` will anyway keep it as the observable list `items` isn't it?

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
```diff
@@ -977,14 +992,13 @@ public final Optional<Integer> allParToVisibleParIndex(int allParIndex) {
...
-        int firstVisibleParIndex = visibleList.get( 0 ).getNode().getIndex();
-        int targetIndex = allParIndex - firstVisibleParIndex;
-
-        if ( allParIndex >= firstVisibleParIndex && targetIndex < visibleList.size() )
-        {
-            if ( visibleList.get( targetIndex ).getNode().getIndex() == allParIndex )
-            {
-                return Optional.of( targetIndex );
+        int virtualIndex = virtualFlowParagraphs.getViewIndex(allParIndex);
+        if ( virtualIndex < 0 ) {
```
Symeon94: Small cosmetic: if you write the if in the other direction you avoid repeating the return of the empty
```Java
if(virtualIndex >= 0) {
// ...
}
return Optional.empty();
```

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
```diff
@@ -2069,11 +2095,50 @@ private void followCaret() {
...
+    }
+
+    private boolean isParagraphFolded(int index) {
+        return foldStyleCheck.test(getParagraphs().get(index).getParagraphStyle());
+    }
+
+    private int getVirtualParagraphIndex(int allParagraphIndex) {
+        int virtualIndex = virtualFlowParagraphs.getViewIndex(allParagraphIndex);
+        if ( virtualIndex < 0 ) {
+            throw new IllegalArgumentException("Paragraph " + allParagraphIndex + " is folded");
```
Symeon94, line 2111: It's a bit radical to throw an exception in that case, ins't there a better way?

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
```diff
@@ -1039,13 +1053,12 @@ public CharacterHit hit(double x, double y) {
...
     public final int lineIndex(int paragraphIndex, int columnPosition) {
-        Cell<Paragraph<PS, SEG, S>, ParagraphBox<PS, SEG, S>> cell = virtualFlow.getCell(paragraphIndex);
-        return cell.getNode().getCurrentLineIndex(columnPosition);
+        return getCell(paragraphIndex).getCurrentLineIndex(columnPosition);
     }
 
     @Override
     public int getParagraphLinesCount(int paragraphIndex) {
-        return virtualFlow.getCell(paragraphIndex).getNode().getLineCount();
+        return isParagraphFolded(paragraphIndex) ? 0 : getCell(paragraphIndex).getLineCount();
```
Symeon94: I have made a comment regarding negative indexes on `getCell()` but if you hadn't raised an IllegalArgument exception in there would the `isParagraphFolded` be avoidable? (in the sense that if `getCell` returned either null or Optional we would know it's folded). It's just a suggestion.

Symeon94: Though, I would understand if the rationale is to keep the code as close as possible to what it was before.

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
```diff
@@ -1765,7 +1789,9 @@ protected double computePrefHeight( double width )
             if ( calcWidth <= 0.0 ) calcWidth = getPrefWidth();
 
             for ( int p = 0; p < getParagraphs().size(); p++ ) {
-                height += getCell( p ).computePrefHeight( calcWidth );
+                if ( ! isParagraphFolded(p) ) {
```
Symeon94, line 1795: I understand why there is that if, but that is a slight change in behaviour compared to previous.

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
```diff
@@ -2004,7 +2030,7 @@ public ParagraphBox<PS, SEG, S> getNode() {
 
             @Override
             public void updateIndex(int index) {
-                box.setIndex(index);
+                box.setIndex(virtualFlowParagraphs.getSourceIndex(index));
```
Symeon94, line 2036: This one I'm wondering about it, why are you transforming the index to a visible index in this Cell?

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
```diff
@@ -2069,11 +2095,50 @@ private void followCaret() {
...
+            if ( virtualIndex >= 0 ) return virtualIndex;
+        }
+
+        for ( int i = allParagraphIndex + 1; i < getParagraphs().size(); i++ ) {
+            virtualIndex = virtualFlowParagraphs.getViewIndex(i);
+            if ( virtualIndex >= 0 ) return virtualIndex;
+        }
+
+        return 0;
+    }
```
Symeon94, line 2149: You can clean the code by merging the first two lines with the loop and separating in before and after
```suggestion
    private int getVisibleParagraphIndex(int allParagraphIndex) {
        int virtualIndex = getVisibleParagraphIndexBefore(allParagraphIndex);
        return virtualIndex >= 0 ? virtualIndex : Math.max(0, getVisibleParagraphIndexAfter(allParagraphIndex));
    }

    private int getVisibleParagraphIndexBefore(int allParagraphIndex) {
        for (int i = allParagraphIndex ; i >= 0 ; i--) {
            int virtualIndex = virtualFlowParagraphs.getViewIndex(i);
            if ( virtualIndex >= 0 ) return virtualIndex;
        }
        return -1;
    }

    private int getVisibleParagraphIndexAfter(int allParagraphIndex) {
        for (int i = allParagraphIndex + 1; i < getParagraphs().size(); i++ ) {
            int virtualIndex = virtualFlowParagraphs.getViewIndex(i);
            if ( virtualIndex >= 0 ) return virtualIndex;
        }
        return -1;
    }
```

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
```diff
@@ -2069,11 +2095,50 @@ private void followCaret() {
...
+            virtualIndex = virtualFlowParagraphs.getViewIndex(i);
+            if ( virtualIndex >= 0 ) return virtualIndex;
+        }
+
+        for ( int i = allParagraphIndex + 1; i < getParagraphs().size(); i++ ) {
+            virtualIndex = virtualFlowParagraphs.getViewIndex(i);
+            if ( virtualIndex >= 0 ) return virtualIndex;
+        }
+
+        return 0;
```
Symeon94: Does it make sense to return 0 if nothing was found? -1 wouldn't make more sense ? (as 0 is a valid value, you show it by testing `if ( virtualIndex >= 0 )`).
(my correction proposal kept your code and returned 0 too, but it might be wrong).

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/ParagraphIndexMappingTests.java
```diff
@@ -54,4 +55,17 @@ public void visible_par_to_all_par_index_is_correct() {
...
+    public void folding_keeps_all_paragraph_indices_and_removes_folded_paragraphs_from_view() {
+        interact(() -> {
+            area.foldParagraphs(1, LAST_PAR_INDEX - 1);
+            area.showParagraphAtTop(LAST_PAR_INDEX);
+        });
+
+        assertTrue(area.isFolded(2));
+        assertEquals(Optional.empty(), area.allParToVisibleParIndex(2));
+        assertEquals(Optional.of(2), area.allParToVisibleParIndex(LAST_PAR_INDEX));
+        assertEquals(LAST_PAR_INDEX, area.visibleParToAllParIndex(2));
```
Symeon94, line 74: I don't see a test where there would be lines after the paragraph, this one checks a paragraph folded at the end of the test.
Cases that could be covered:
- Paragraph folded at first index
- Two paragraphs folded inside the text

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/ParagraphIndexMappingTests.java
Symeon94, line 1: Lots of public methods were modified but not tested (for instance `lineIndex`, but that is not the only one).

Symeon94 (changes requested):

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
Col-E: Couldn't find any way to do index mapping with just the `virtualFlow` wrapping the list. So keeping it as a field `virtualFlowParagraphs` allows use further below when mapping between filtered/non-filtered indices.

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
Col-E: Its what I saw a number of other places doing for negative values _(like in `Paragraph` for negative positions)_.

I don't know if there's a great alternative. Clamping to 0 seems like it could lead to silent unintended behavior. Using Optional seems like it would just kick the issue to the call site plus is pointless object allocation / wrapping. In some places defaulting to 0 may work, but in others it may be wrong.

Also this method is used by `getCell(int index)` which is effectively `@NonNull`. We have to provide _some_ valid value here that maps to a realized cell.

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
Col-E: The ternary here is intentional since if the paragraph is folded we skip through `getCell(index)`. This is where we avoid a bit of cell node realization.

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
Col-E: Similar to the last comment thread, this is for avoiding cell realization while getting the same results. Folded paragraphs have 0 height.

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
Col-E: The `Cell` in this context is operating on the virtual flow's filtered list index. This needs to be mapped to keep in-alignment with the other areas of index mapping / filtering, otherwise you get some _weird_ artifacts like this:

java_9O7reQATwg

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
Col-E: Also I see that for the suggested change the spacing of the for loop variable is different. I tried to keep the original format in the changes, but if there's a different formatting direction you want for new code let me know.

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
Col-E: > Does it make sense to return 0 if nothing was found?

~~Using 0 here since I believe the only case for getting here is if there are no paragraphs at all, in which 0 seems like a sensible default. I'll see if I can sanity check this in a test case.~~

Yeah, nevermind will change.

> my correction proposal kept your code and returned 0 too, but it might be wrong

The proposed changes kept the final value clamped to >= 0 so it still matches the intent of whats here.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/ParagraphIndexMappingTests.java
Col-E: Pushed up a number of additional tests for folding, lmk if there's anything I missed

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
Symeon94: (I thought I had closed this comment, sorry about that. I have seen that it's extensively used, closing this comment).

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
Symeon94: The spacing was lost when I rewrote it on my side. (I don't care too much about spacing, but it's always a good idea to stay aligned with what was there, you are right). Resolved.

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
```diff
@@ -1819,25 +1845,25 @@ protected void layoutChildren() {
...
     TwoDimensional.Position currentLine() {
         int parIdx = getCurrentParagraph();
-        Cell<Paragraph<PS, SEG, S>, ParagraphBox<PS, SEG, S>> cell = virtualFlow.getCell(parIdx);
-        int lineIdx = cell.getNode().getCurrentLineIndex(caretSelectionBind.getUnderlyingCaret());
+        ParagraphBox<PS, SEG, S> cell = getCell(parIdx);
+        int lineIdx = cell.getCurrentLineIndex(caretSelectionBind.getUnderlyingCaret());
         return paragraphLineNavigator.position(parIdx, lineIdx);
     }
 
     void showCaretAtBottom() {
```
Symeon94, line 1856: Maybe something for another PR (that's more a comment for me), I see `showCaretAtBottom` and `showCaretAtTop` are package-private and are not used anywhere.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/ParagraphIndexMappingTests.java
```diff
@@ -54,4 +57,119 @@ public void visible_par_to_all_par_index_is_correct() {
...
+        // fold paragraphs 1-3, then show paragraph 4 at the top of the viewport
+        interact(() -> {
+            area.foldParagraphs(1, 3);
+            area.showParagraphAtTop(4);
+        });
+
+        // sanity check expected folding state
+        assertTrue(area.isFolded(2));
+        assertTrue(area.isFolded(3));
+        assertFalse(area.isFolded(4));
```
Symeon94: Why not check before the fold and the first index of the fold?
```Java
assertFalse(area.isFolded(0));
assertTrue(area.isFolded(1));
```

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/ParagraphIndexMappingTests.java
```diff
@@ -54,4 +57,119 @@ public void visible_par_to_all_par_index_is_correct() {
...
+        // fold paragraphs 1-2 and 4-5, then show paragraph 3 at the top of the viewport
+        interact(() -> {
+            area.foldParagraphs(1, 2);
+            area.foldParagraphs(4, 5);
+            area.showParagraphAtTop(3);
+        });
+
+        // sanity check expected folding state
+        assertTrue(area.isFolded(2));
+        assertTrue(area.isFolded(5));
```
Symeon94: Why limit to these two cases, why not test all entries from 0 to 6? (meaning from before the first fold to after the last one).
It is true for the checks below too.
You could define:
```Java
void checkFolded(int... lines) {
	for(int line : lines) {
		checkFolded(line);
	}
}

void checkFolded(int index) {
	assertTrue(area.isFolded(index), "Error for is folded for index " + i);
	assertTrue(area.allParToVisibleParIndex(index).isEmpty(), "Error for is visible for index " + i);
}
```
Then simplify in your test:
```Java
checkFolded(1, 2, 4, 5);
```

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/ParagraphIndexMappingTests.java
Col-E: Both would be false since for any range `A-B` the _"preview"_ of the folded space is line `A` - but I can add the two checks.

Symeon94: Ok, I understand now that it must be false. In that case, adding a test to verify this behavior makes sense.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/ParagraphIndexMappingTests.java
Col-E: Per last thread `checkFolded(2, 5)` and maybe a `checkUnfolded(1, 3, 4)` - Can do though.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/ParagraphIndexMappingTests.java
```diff
@@ -54,4 +57,119 @@ public void visible_par_to_all_par_index_is_correct() {
...
+            area.showParagraphInViewport(0);
+            area.showParagraphAtTop(0);
+            area.showParagraphAtBottom(0);
+            area.showParagraphRegion(0, new javafx.geometry.BoundingBox(0, 0, 1, 1));
+            area.showParagraphAtCenter(0);
+        });
+
+        // verify that all paragraphs are folded and not mapped to any visible index
+        assertEquals(Optional.empty(), area.allParToVisibleParIndex(0));
+        assertEquals(-1, area.visibleParToAllParIndex(0));
```
Symeon94: It seems the verification only covers index 0 instead of verifying as advertised that "all paragraphs are folded".

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/ParagraphIndexMappingTests.java
Symeon94: Your change touches a lot of some central class, so I hope you'll forgive me for insisting on this test part. I have listed many methods which are impacted and which it would be proper to test the behaviour in the circumstances of a fold.
Here is the list I have (I removed those methods which I have seen covered already):
* removeSelection(Selection)
* hit(double, double)
* lineIndex(int, int)
* getCharacterBoundsOnScreen(int, int)
* getParagraphBoundsOnScreen(int)
* getCaretBoundsOnScreen(T)
* showParagraphInViewport(int)
* showParagraphAtBottom(int)
* showParagraphRegion(int, Bounds)
* showParagraphAtCenter(int)
* getCurrentLineStartInParargraph()
* getCurrentLineEndInParargraph()
* caretSelectionBind.paragraphIndexProperty().addListener( this::skipOverFoldedParagraphs ); (this one is impacted, but it's through private method).
* double computePrefHeight(double) (that's a protected one)
* updateIndex(int)
* next/previous lines move is impacted by private method changes
* getSelectionBoundsOnScreen is used in defining the bounds, not sure it can be tested.

Not sure all of them can be easily tested, but some are simpler to cover.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/ParagraphIndexMappingTests.java
Col-E: If all paragraphs are folded, then the only visible index you can have is `0` since that is the preview for the folded range. Any value `n >= 1` for `area.visibleParToAllParIndex(n)` will throw since its out of bounds.

`area.allParToVisibleParIndex(n)` would be safe since it maps to `Optional.empty()` for any folded paragraph.

Symeon94: Make sense that it would throw, indeed. You can then check in a loop that "all paragraphs are folded":
```Java
for(int i = 1 ; i < /* length */ ; i++) {
assertTrue(area.allPartToVisibleParIndex().isEmpty());
assertThrows(IllegalArgumentException.class, area.visibleParToAllParIndex(i);
}
```

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
Col-E: I can make `visibleParToAllParIndex` and `allParToVisibleParIndex` not throw, still don't think I can here.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/ParagraphIndexMappingTests.java
Col-E: Pushed up some additional tests, I believe everything is now covered directly or indirectly in cases like `computePrefHeight` and `skipOverFoldedParagraphs`

| Method | Covering test |
|-----------------|---------------|
| `removeSelection(Selection)`           | `MultipleCaretSelectionTests.removing_selection_spanning_folded_paragraph_works` |
| `hit(double, double)`                  | `FoldedParagraphTests.hit_on_visible_paragraph_after_folded_paragraph_uses_source_index` |
| `lineIndex(int, int)`                  | `FoldedParagraphTests.line_index_works_for_visible_paragraph_and_rejects_folded_paragraph` |
| `getCharacterBoundsOnScreen(int, int)` | `FoldedParagraphTests.character_bounds_are_empty_for_folded_paragraph` |
| `getParagraphBoundsOnScreen(int)`      | `FoldedParagraphTests.paragraph_bounds_are_empty_for_folded_paragraph_and_present_for_visible_paragraph` |
| `getCaretBoundsOnScreen(T)`            | `FoldedParagraphTests.caret_bounds_are_empty_for_caret_in_folded_paragraph` |
| `showParagraphInViewport(int)`         | `FoldedParagraphTests.show_paragraph_in_viewport_resolves_a_folded_target_to_a_visible_paragraph` |
| `showParagraphAtBottom(int)`           | `ParagraphIndexMappingTests.showing_a_folded_paragraph_at_the_end_uses_the_last_visible_paragraph` |
| `showParagraphRegion(int, Bounds)`     | `FoldedParagraphTests.show_paragraph_region_resolves_a_folded_target_to_a_visible_paragraph` |
| `showParagraphAtCenter(int)`           | `FoldedParagraphTests.show_paragraph_at_center_resolves_a_folded_target_to_a_visible_paragraph` |
| `getCurrentLineStartInParargraph()`    | `FoldedParagraphTests.line_index_works_for_visible_paragraph_and_rejects_folded_paragraph` |
| `getCurrentLineEndInParargraph()`      | `FoldedParagraphTests.line_index_works_for_visible_paragraph_and_rejects_folded_paragraph` |
| `skipOverFoldedParagraphs` listener    | `FoldedParagraphTests.navigation_skips_folded_paragraphs_in_both_directions` |
| `computePrefHeight(double)`            | `FoldedParagraphTests.auto_height_excludes_folded_paragraphs` |
| `updateIndex(int)`                     | Indirectly covered by `FoldedParagraphTests.hit_on_visible_paragraph_after_folded_paragraph_uses_source_index` and the paragraph-index mapping tests |
| Next/previous over folded paragraphs   | `FoldedParagraphTests.navigation_skips_folded_paragraphs_in_both_directions` |
| `getSelectionBoundsOnScreen`           | Indirectly covered by `MultipleCaretSelectionTests.removing_selection_spanning_folded_paragraph_works`, via `selection.getSelectionBounds()` |

Symeon94: Thanks for taking the time to implement additional tests.

I have a comment before reviewing the changes fully, it came to me while looking at the first test from the new test class. It caught my attention, for the new code is throwing an exception if `area.linedIndex(1,0)` is called on folded paragraph. It appeared to me that it could depart from the existing behaviour. I therefor tried to run the new test class on the existing code and 7 out of 11 new tests failed on existing code (I didn’t try the ones from the other classes). This likely means the modifications changed the library’s behaviour. Someone could, for instance, be calling a `lineIndex` on a folded paragraph and discover its code starts throwing exceptions around.

The purpose of the overall change was to fix the jitter issue due to performance, which is an internal implementation problem. The change shouldn’t be visible to the end users.

Col-E: `lineIndex` relies on the real `ParagraphBox` since it calls into the `ParagraphText` _(A real `TextFlow` with FX layout handling)_ to query for line wrapping. With the changes its not possible to do this since those scene nodes no longer get created while their containing paragraph is folded. 

I don't think there will be a perfect _"invisible"_ change here. However, I think that for this case rather than an exception throw, falling back to `0` for folded lines can be a reasonable expectation. The text isn't visible since its folded, so what line wrapping is there to be done, right?

Col-E: Pushed up a change modeling this fallback handling and updated the first test in `FoldedParagraphTests` to show the non-throwing behavior.

Symeon94: I understand there might be areas where changes compared to previous behaviour could be acceptable and even unavoidable. For instance `getParagraphBoundsOnScreen` is present only if the cell is visible, and it seems that your change on purpose made the cell invisible. However, we should be transparent about what these changes are and why they are there. I’ll review the test and see if I see any additional point to raise.

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
```diff
@@ -1006,14 +1000,8 @@ public final Optional<Integer> allParToVisibleParIndex(int allParIndex) {
...
     public final int visibleParToAllParIndex(int visibleParIndex) {
-        if (visibleParIndex < 0) {
-            throw new IllegalArgumentException("Visible paragraph index cannot be negative but was " + visibleParIndex);
-        }
-        if (visibleParIndex > 0 && visibleParIndex >= getVisibleParagraphs().size()) {
-            throw new IllegalArgumentException(String.format(
-                    "Visible paragraphs' last index is [%s] but visibleParIndex was [%s]",
-                    getVisibleParagraphs().size() - 1, visibleParIndex)
-            );
+        if (visibleParIndex < 0 || (visibleParIndex > 0 && visibleParIndex >= getVisibleParagraphs().size())) {
```
Symeon94, line 1016: This condition look strange (you kept previous code though, so was there an issue before?):
* 0 for some reason isIsgnored (if there is no visible paragraph it will go through)
* If 0 being ignored is an error, it can be simplified to `visibleParIndex <0 || visibleParIndex >= getVisibleParagraphs().size()`

It might be another issue to open (I need to look into it), **no fix on this PR**

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
```diff
@@ -1006,14 +1000,8 @@ public final Optional<Integer> allParToVisibleParIndex(int allParIndex) {
...
+            return -1;
```
Symeon94: You are changing an existing behaviour: It used to throw an exception and now it isn't.

I think your change make sense and should be an improvement, but we cannot assume all implementation will agree.

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
```diff
@@ -1053,7 +1041,9 @@ public CharacterHit hit(double x, double y) {
 
     @Override
     public final int lineIndex(int paragraphIndex, int columnPosition) {
-        return getCell(paragraphIndex).getCurrentLineIndex(columnPosition);
+        return getCellIfVisible(paragraphIndex)
+                .map(c -> c.getNode().getCurrentLineIndex(columnPosition))
+                .orElse(0);
```
Symeon94, line 1059: I suspect a change of behaviour again, if getCell used to throw an exception (which I suspect), your new implementation will return 0. If you really are out of bound on this one, I think you should indeed have IndexOutOfBounds (as there used to be).

Where there shouldn't be an exception is for cell that are collapsed (where you proposed to return the first line of the collapsed entry).

Also, I see something else: your UT if I recall test collapse on line 0, which is not good because it's also the default value here, you should probably cover a case != 0.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
```diff
@@ -0,0 +1,227 @@
...
+            area.replaceText("zero\none\ntwo");
+            area.foldParagraphs(0, 1);
+
+            // the folded paragraph should not have character bounds on screen, but the visible paragraphs should.
+            int start = area.getAbsolutePosition(0, 0);
+            assertTrue(area.getCharacterBoundsOnScreen(start, start + 1).isPresent());
+            start = area.getAbsolutePosition(1, 0);
+            assertFalse(area.getCharacterBoundsOnScreen(start, start + 1).isPresent());
+            start = area.getAbsolutePosition(2, 0);
+            assertTrue(area.getCharacterBoundsOnScreen(start, start + 1).isPresent());
```
Symeon94: Why do you limit your check on the first char? Why not loop on values that you know are there or not
Here it's [5;8) that are not visible, hence just do this to cover all cases:
```suggestion
for(int i = 0 ; i < 12 ; i++) {
    assertEquals(i < 5 || i >= 8, area.getCharacterBoundsOnScreen(i, i+ 1).isPresent());
}
```

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
```diff
@@ -0,0 +1,227 @@
...
+        // fold 'one' paragraph, so that 'zero' and 'two' are visible
+        interact(() -> {
+            area.replaceText("zero\none\ntwo");
+            area.foldParagraphs(0, 1);
+        });
+
+        // create a caret in the folded paragraph, which should not have bounds on screen.
+        CaretNode caret = new CaretNode("folded caret", area, area.getAbsolutePosition(1, 0));
+        interact(() -> {
+            assertFalse(area.getCaretBoundsOnScreen(caret).isPresent());
```
Symeon94, line 117: Are you sure the bounds must not be present? (previous behaviour returned `true`) The caret must be somewhere, I would expect that it would be at the end of the folded line or somewhere else.
Isn't there are risk that if moved the caret might dissapear when moving on the folded line?

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
Symeon94, line 1: I have noticed a common pattern in all the tests you are making: they only cover fold on index = 0. I think you should reconsider the cases, as folding on index = 0 is an edge case. What happens when there is a line before the fold is never covered in these different cases present here.
Optimally we should test fold at the start, in the middle and at the end.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
```diff
@@ -0,0 +1,227 @@
...
+    @Test
+    public void paragraph_graphic_access_rejects_folded_paragraph() {
+        // fold 'one' paragraph, so that 'zero' and 'two' are visible
+        interact(() -> {
+            area.replaceText("zero\none\ntwo");
+            area.foldParagraphs(0, 1);
+        });
+
+        // getting/creating graphics for the folded paragraph throws an exception, since it is not visible.
+        assertThrows(IllegalArgumentException.class, () -> area.getParagraphGraphic(1));
```
Symeon94: Previous behaviour was sending null when there was no graphic set (and it was sending null for folded). Why not send null instead of throwing an exception?

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
```diff
@@ -0,0 +1,227 @@
...
+        assertThrows(IllegalArgumentException.class, () -> area.recreateParagraphGraphic(1));
```
Symeon94: you are not testing for 0 and 2?

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/ParagraphIndexMappingTests.java
```diff
@@ -52,6 +54,162 @@ public void visible_par_to_all_par_index_is_correct() {
...
+    }
+
+    @Test
+    public void invalid_paragraph_indices_are_rejected() {
+        interact(() -> {
+             assertEquals(Optional.empty(), area.allParToVisibleParIndex(-1));
+             assertEquals(Optional.empty(), area.allParToVisibleParIndex(area.getParagraphs().size()));
+
+             assertEquals(-1, area.visibleParToAllParIndex(-1));
+             assertEquals(-1, area.visibleParToAllParIndex(Integer.MAX_VALUE));
```
Symeon94: Both threw IllegalArgumentException before, which made sort of sense, as these are not valid values.
Now it's behaving differently

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/ParagraphLinesCountTests.java
```diff
@@ -32,4 +32,15 @@ public void single_line_returns_one() {
...
 
+    @Test
+    public void folded_paragraph_returns_zero_lines() {
+        interact(() -> {
+            area.replaceText("first\nsecond\nthird");
+            area.foldParagraphs(0, 1);
+        });
+
+        assertEquals(1, area.getParagraphLinesCount(0));
+        assertEquals(0, area.getParagraphLinesCount(1));
```
Symeon94, line 43: That's probably the most dangerous change, it used to return 1 and the method has no mention of visible index. If anyone does any logic calculation on this, it will change the outcome.

Symeon94:
I just finished reviewing, I wanted to point the method changes I have seen. 

These have changed but should be ok (mostly they refer to visibility, so it's normal it would change), maybe it should be documented somewhere still:

getParagraphBoundsOnScreen
getVisibleParagraphBoundsOnScreen
getCharacterBoundsOnScreen
showParagraphInViewport

This one the change could be risky if the caller supposed that empty() is the end of the visible lines (now empty() could be folded):
allParToVisibleParIndex

These three I'm not sure it is advisable to change:

getCaretBoundsOnScreen -> Caret needs still to get bounds, even on folded
getParagraphGraphic -> Maybe return null instead of empty
getParagraphLinesCount -> I wrote a comment where I have seen it, the issue is that any existing logic using this one might have not considered that it was referring to visible lines (and there is no reference in the name that let us think that).

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
Symeon94: I was wrong on this one, negative index is an invalid input, and it is right to throw it there and elsewhere. You cant ignore my comment (it's like indexing a non-existing index in an array, it's not allowed).

Col-E:
I've been trying to leave comments but github is not playing nice... Its putting comments in the "review mode" where you bulk comment then approve, despite me not ever starting a review. Then it just orphans the comments from the threads I was replying to so now they're floating without any proper context 😫 

image

Symeon94:
@Col-E I cannot answer on your last comment, so I put it here. I didn't create this library, so as far as exception is concerned I'm like you left to guess.
My original comment was on your new addition of exception. During review I saw how it was done elsewhere in this code and I thought a bit about it. It seems to me that it makes sense to throw exception for invalid indexing (like you would get from referencing a non-existing reference in an array). That would mean -1 and >= size/length. So your original exception made sense (and was aligned with the existing code). My apologies for commenting there.

Optional.empty() or -1 should be kept for valid input that returns non-existing.

What I'd like to avoid is to change existing behaviour if not absolutely necessary (that also means exceptions). As this might impact library users. For visible cells it is most of the time intrinsically related to your change. So, there changing the behaviour is mostly acceptable (I have commented on those where I thought it is either not as clear or wrong).

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
Col-E: The first char is all that is needed to validate the per-line contract of whether or not it is folded. Checking each character on a line would be a bit redundant. The suggested for loop also is more difficult to read / parse intent. The three checks line up with the three lines.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
Col-E: > Are you sure the bounds must not be present?

The caret itself is technically present. However, its screen-space bounds when it models an index in a folded range is correctly missing. The containing paragraph is folded and thus not realized. This follows the same pattern as the test above `character_bounds_are_empty_for_folded_paragraph`.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
Col-E: This followed the cases where its backed by a `getCell(int)` which we've since gone to non-throwing and back on. I can do the same `getCellIfVisible().getOr(null)` approach if desired. The same would also apply to `recreateParagraphGraphic(int)`.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
Col-E: >  Optimally we should test fold at the start, in the middle and at the end.

Well, folding `0, N` still keeps the first paragraph visible. Its only possible to fold middle and end paragraphs.

image

The only change that could be made would be to expand the fold ranges in tests to not include just middle lines, but the end line as well. Less of an edge case considering how the folding really only treats line 0 as an edge with it being the minimum anchor line.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
Symeon94: I understand, but when you say "The first char is all that is needed" that is an assumption that I have not seen validation for.
You test covers `start` with values 0,5,9 (from what I remember). If not a loop, we can check one case in one of the test.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
Symeon94: Before closing this comment, do you mean that when we move the charet (for instance with the keys) the caret jump above the folded paragraph to the next visible position and that no request for bound in the folded area will ever take place?
If yes, you can close this comment.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
Symeon94: In JavaFX, often you can use `null` to disable the display of something (for instance `setGraphics(null)`). So, it could be that the null was a valid expected case and that this will break.
Yes, I think it's safer to keep that behaviour and revert.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
Symeon94: By edge case, I meant that 0 is the minimum acceptable value. Optimally that kind of test should cover for min, max values and something in between. And invalid cases should also get some coverage like min-1, max+1 and why not cases where end < start and end == start.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
Symeon94: (Small addition, I tested this on the existing code, you are missing a graphic factory, else it will all be null):
```Java
area.setParagraphGraphicFactory(i -> new Region());
```

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
Col-E: For extra carets like this new one, the test `//` comment description is accurate. There aren't going to be any bounds since its located in the folded paragraph and this is an additional `CaretNode`, not the primary one used in navigation.

For the primary caret it does what you describe with jumping across folded paragraphs, see `navigation_skips_folded_paragraphs_in_both_directions`

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
Col-E: Pushed

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
Symeon94: What the test you are mentioning does is calling `area.moveTo`. Stating that a call to this method or having the user press a key on his keyboard is the same is an assumption, which can be true or false. To cover the case, the test should have fired a key event after having selected before the paragraph.
I actually tried that locally on my system to fire key events instead of `area.moveTo` and it indeed jumps above the paragraphs:
```Java
    @Test
    public void navigation_skips_folded_paragraphs_in_both_directions() {
        interact(() -> {
            // fold 'two' and 'three' paragraphs, so that 'zero', 'one', and 'four' are visible
            area.replaceText("zero\none\ntwo\nthree\nfour");
            area.foldParagraphs(1, 3);

            // verify that moving forward to the folded paragraph skips to the next visible paragraph
            assertEquals(1, area.getCurrentParagraph());
            assertEquals(5, area.getCaretPosition());
            area.fireEvent(createKeyEvent(KeyCode.RIGHT));
            assertEquals(6, area.getCaretPosition());
            area.fireEvent(createKeyEvent(KeyCode.RIGHT));
            assertEquals(7, area.getCaretPosition());
            area.fireEvent(createKeyEvent(KeyCode.RIGHT));
            assertEquals(8, area.getCaretPosition());
            assertEquals(1, area.getCurrentParagraph());
            area.fireEvent(createKeyEvent(KeyCode.RIGHT));
            assertEquals(19, area.getCaretPosition());
            assertEquals(4, area.getCurrentParagraph());

            // verify that moving backward to the folded paragraph skips to the previous visible paragraph
            area.fireEvent(createKeyEvent(KeyCode.LEFT));
            assertEquals(8, area.getCaretPosition());
            assertEquals(1, area.getCurrentParagraph());
        });
    }

    private static Event createKeyEvent(KeyCode keyCode) {
        return new KeyEvent(KeyEvent.KEY_PRESSED, keyCode.getChar(), "unset", keyCode, false, false, false ,false);
    }
```
This means you can close this conversation as indeed there will never be a call to `getCaretBoundsOnScreen` for folded bounds.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
```diff
@@ -215,6 +218,34 @@ public void navigation_skips_folded_paragraphs_in_both_directions() {
...
+            assertEquals(7, area.getCaretPosition());
+            area.fireEvent(createKeyEvent(KeyCode.RIGHT));
+            assertEquals(8, area.getCaretPosition());
+            assertEquals(1, area.getCurrentParagraph());
+            area.fireEvent(createKeyEvent(KeyCode.RIGHT));
+            assertEquals(19, area.getCaretPosition());
+            assertEquals(4, area.getCurrentParagraph());
+
+            // verify that moving backward to the folded paragraph skips to the previous visible paragraph
+            area.moveTo(3, 0);
```
Symeon94: You removed the triggering of the left key event?

Col-E: Whoops only copied the middle block you posted.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
Col-E: The three lines are testing if any char has bounds for each paragraph. If the first character in a paragraph has bounds _(because the paragraph is visible)_ then the others will to. The only case where the bounds will be missing is if the line is folded, in which all characters will also have their bounds missing for that paragraph.

Col-E: Tweaked the for loop for legibility purposes

Symeon94: The statement says that the lines are testing "_if any char has bounds_". The test is only testing for the first char of each paragraph, therefore not "_any char_".
> If the first character in a paragraph has bounds (because the paragraph is visible) then the others will to.

This is an assumption. I agree that it is likely true, but we can only be sure if there is a test validating enough cases to be confident in that statement with a high level of probability.

I proposed to validate that assumption to loop on every case. I agree with you a loop is often less clear, but I think we should still test at least one value inside a paragraph.

(Update: I posted before seeing your second comment)

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/ParagraphIndexMappingTests.java
Symeon94: I think we can close this comment.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
```diff
@@ -81,18 +82,24 @@ public void paragraph_bounds_are_empty_for_folded_paragraph_and_present_for_visi
...
+                for (int charIndex = 0; charIndex <= 2; charIndex++) {
+                    int pos = area.getAbsolutePosition(lineIndex, charIndex);
+                    Optional<Bounds> charBounds = area.getCharacterBoundsOnScreen(pos, pos + 1);
+                    if (lineIndex == foldedLine) {
+                        assertFalse(charBounds.isPresent());
+                    } else {
+                        assertTrue(charBounds.isPresent());
+                    }
+                }
+            }
```
Symeon94, line 102: Your double loop is much more complicated than a single loop like I have shown, I'm not sure what is the reason for that additional complexity. The absolute position is simply a value from 0 to the text length, and you can check the bounds is present with `assertEquals`. But ok, if you want to keep it the way you wrote it, I'm fine with it. You can close this comment.

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
Symeon94: Ok, this is not an issue because `getVisibleParagraphs()` is always greater than 0 (even for empty content). This actually means that the `visibleParIndex > 0` check is unnecessary (unless it is intended as some kind of optimization). Since we know the size is always at least 1, keeping the `0` case is fine and avoids calling `getVisibleParagraphs()` unnecessarily (although I think it makes the code a bit obscure).

Closing the comment.

richtextfx/src/main/java/org/fxmisc/richtext/GenericStyledArea.java
Symeon94: I managed to test multiline with:
```Java
        interact(() -> {
            area.setWrapText(true);
            area.replaceText("""
                    To be, or not to be, that is the question: whether 'tis nobler in the mind to suffer the slings and arrows of outrageous fortune, or to take arms against a sea of troubles and by opposing end them.
                    """);
            for(int i = 0 ; i < 190 ; i++) {
                int lineIndex = i > 153 ? 2 : i > 77 ? 1 : 0;
                assertEquals(lineIndex, area.lineIndex(0, i), "Failed for index " + i);
            }
        });
```
Now it returns something else than 0 only if it's wrapped. So for invisible cell you can indeed return 0 (although I'd say for invisible cell it's an invalid case strictly speaking).
So you can ignore my comment or add a test as your prefer, I'll resolve this comment.

richtextfx/src/integrationTest/java/org/fxmisc/richtext/api/FoldedParagraphTests.java
Col-E: > Your double loop is more complicated than the single loop
> 
> ```java
> for(int i = 0 ; i < 12 ; i++) {
>     assertEquals(i < 5 || i >= 8, area.getCharacterBoundsOnScreen(i, i+ 1).isPresent());
> }
> ```

The single loop was going over character indices with a range check for what was on the 2nd line.
This double loop iterates over paragraphs, then the characters per-paragraph.

If you had a test failure and set a breakpoint it would be easier with this to see that you're failing on line N+offset. IMO despite being more code, this follows the line/paragraph model and is easier to understand at a glance.

Symeon94: The following gives the absolute position (from 0 to length -1):
```Java
int pos = area.getAbsolutePosition(lineIndex, charIndex);
```
You can add a `System.out.println` to check it out. So the double loop is really unnecessary (unless the point is to test `getAbsolutePosition`, but I don't think that's the case).

If you want to improve my loop, you can do:
```Java
for(int i = 0 ; i < 12 ; i++) {
    boolean isVisibleParagraph = i < 5 || i >= 8; // That was the right interval for the previous text
    assertEquals(isVisibleParagraph, area.getCharacterBoundsOnScreen(i, i+ 1).isPresent());
}
```

That said, I'm fine leaving as is, I have made enough comments ;-)

